Issue #65, opened 2014-02-10
bwdist


yakir12:
I'm dying to move from Matlab to Julia, and doing my best, but there are a few power-functions that Matlab has and Julia doesn't. One of which is bwdist:

> D = bwdist(BW) computes the Euclidean distance transform of the binary image BW. For each pixel in BW, the distance transform assigns a number that is the distance between that pixel and the **nearest nonzero pixel** of BW.  

The Matlab algorithm is based on [A linear time algorithm for computing exact Euclidean distance transforms of binary images in arbitrary dimensions](http://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=1177156), which is beyond my scope as a novice Julia user :(
This is a suggestion for an enhancement... Seems natural it should belong to Images.jl (as opposed to https://github.com/JuliaStats/Distance.jl), but I might be wrong!

P.S.
Thanks for all the awesome work!!!

juliohm:
@yakir12, meanwhile you can use the nice PyCall package:

```julia
using PyCall
@pyimport scipy.ndimage.morphology as pymorph

A = eye(5)
D = pymorph.distance_transform_edt(1-A)
```

Check this thread on SO: http://stackoverflow.com/questions/5260232/matlab-octave-bwdist-in-python-or-c

timholy:
There's also #99. It would be great if someone who uses this function could test it.

yakir12:
This is great! I'll test #99 ASAP.

muhlba91:
Work of #99 has been carried on in #576

juliohm:
I think this issue can be closed now, can you confirm?

yakir12:
Yes!

Den 2 okt 2017 07:54 skrev "Júlio Hoffimann" <notifications@github.com>:

PR #99, opened 2014-05-29
Euclidean Distance Transform

wkearn:
I needed a Euclidean Distance Transform for something I'm working on, so I wrote up the algorithm from [Maurer et al. 2003](http://ieeexplore.ieee.org/xpl/login.jsp?tp=&arnumber=1177156&url=http%3A%2F%2Fieeexplore.ieee.org%2Fstamp%2Fstamp.jsp%3Ftp%3D%26arnumber%3D1177156) Matlab uses for its `bwdist` function (but I wrote this from scratch after puzzling out the Maurer et al. pseudocode). 

This solves #65

Calling `(F,D) = bwdist(I)` where `I` is some binary image returns two results. F is the feature transform (i.e. the closest feature pixel given as a vector of its indices in I). D is the Euclidean Distance Transform calculated from F by computing the Euclidean distance from the pixel of interest to its feature pixel.

I'm fairly confident that it works on 2D images, but it should also work on multi-dimensional ones as well. The rudimentary tests that I've done on images of various dimensions seem to hold up. Let me know if you encounter any cases where it doesn't behave as expected.

Current problems: This is written using Base.Cartesian, so it won't work in 0.2. I had a bit of trouble getting the Cartesian.jl version of `@ngenerate` to work. I'll revisit that later (or someone with a bit more Cartesian-fu can take a crack at it :wink:), but until then, I've left it in a separate file rather than merging it with algorithms.jl.

### DIFF
--- a/src/bwdist.jl
+++ b/src/bwdist.jl
@@ -1,0 +1,87 @@
+using Base.Cartesian
+using Distance
+
+macro nvect(N,ex)
+    _nvect(N,ex)
+end
+
+function _nvect(N::Int,ex)
+    vars = [Base.Cartesian.inlineanonymous(ex,i) for i = 1:N]
+    Expr(:escape, Expr(:vcat,vars...))
+end
+
+@ngenerate N (Array{Vector{Int}},Array{Int}) function bwdist{N}(I::AbstractArray{Bool,N})
+    F = Array(Vector{Int},size(I))
+    computeft!(F,I)
+    d = [1:N]
+    for i in d
+        voronoift!(F,I)
+        F = permutedims(F,circshift(d,1))
+        for j in 1:length(F)
+            F[j] = circshift(F[j],1)
+        end
+    end
+    D = zeros(Int,size(I))
+    @nloops N i F begin
+        (@nref N D i) = evaluate(SqEuclidean(),(@nvect N i),(@nref N F i))
+    end
+    return (F,D)
+end
+
+@ngenerate N typeof(F) function computeft!{N}(F::Array{Vector{Int},N},I::AbstractArray{Bool,N})
+    @nloops N i I begin
+        (@nref N F i) = (@nref N I i) ? (@nvect N i) : (@nvect N j->0)
+    end
+end
+
+function voronoift!(F::Array{Vector{Int}},I::AbstractArray{Bool})
+    nfv = count(x->x,I)
+    for d = 1:size(F,2)
+        Fstar = slicedim(F,2,d)
+        l = 0
+        g = fill([0 for j = 1:ndims(I)],nfv+1)
+        for i in 1:size(Fstar,1)
+            f = Fstar[i]
+            if f != [0,0]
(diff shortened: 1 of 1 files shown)


src/bwdist.jl
```diff
@@ -0,0 +1,156 @@
+using Base.Cartesian
+
+# A function to permute a tuple of subscripts given a
+# permutation vector
+function permutesubs(subs::(Int...),perm::Vector{Int})
```
juliohm: Do we need this function? Why not use `subs[perm]`?

src/bwdist.jl
```diff
@@ -0,0 +1,156 @@
...
+    s
+end
+
+sqeuclidean(a::AbstractVector, b::AbstractVector) = sumsqdiff(a,b)
+euclidean(a::AbstractVector,b::AbstractVector) = sqrt(sumsqdiff(a,b))
+
+# Be careful with this: you really need an Array of 0s and 1s
+bwdist(I::AbstractArray{Real}) = bwdist(convert(Bool,I))
+# bwdist: call this one!
+@ngenerate N (Array{Int},Array{Int}) function bwdist{N}(I::AbstractArray{Bool,N})
```
juliohm: `@ngenerate` is deprecated in `VERSION >= v"0.4.0-dev+3184"`

timholy: True, but it will be available in `Compat` for some time yet.

iamed2:
Is this a thing that will be merged at some point?

timholy:
It's clearly something of great interest. I've thought several times about merging it, but the fact that it's a lot slower than Matlab still concerns me. Someone will file a performance issue, and then what happens. Do I have to fix it myself? I don't use this functionality (yet), and have far too many packages I maintain.

But I hadn't remembered that the gap was "only" 10x, so maybe it's not so horrible. I was still remembering the 100x days.

Anyone else want to chime in? I could be persuaded either way. But it would need tests and documentation first (if @wkearn doesn't feel like adding them, someone else could fork this branch, add tests, and submit a new PR).

rsrock:
Having some code for this is better than none. That will encourage one of us to write tests, docs, and improve performance. The perfect is the enemy of the good, etc.  I'm in favor of merging.

wkearn:
I'm sorry I've neglected this for so long. I stopped needing a distance transform for my own work, and this fell by the wayside. Let me look at this again this week and see if I can clean it up and improve the performance. Unless someone else has good ideas about how to improve it right now.

timholy:
I've not looked into this algorithm at all, but one comment is that multidimensional stuff is a lot easier now thanks to http://docs.julialang.org/en/latest/manual/arrays/#iteration (you can add/subtract/min/max `CartesianIndex`es) and http://docs.julialang.org/en/latest/manual/metaprogramming/#generated-functions.

muhlba91:
note: work carried on in PR #576

timholy:
Closed by #580

PR #126, opened 2014-07-22
Further document the structure of spatialorder

rennis250:
For some reason, this one always catches me off guard and I sit for a while wondering why `permutedims` fails.

Best,
Rob


timholy:
Now that spatial order is determined by coordinate order, I think this is not necessary. Thanks for the offer, @rennis250!

PR #231, opened 2014-12-04
Discrete Fourier transform registration (subpixel translation correction)

romainFr:
Something I needed at some point that might be useful to more people. It's taken from [Guizar & al](http://www.opticsinfobase.org/ol/fulltext.cfm?uri=ol-33-2-156&id=148843) (Manuel Guizar-Sicairos, Samuel T. Thurman, and James R. Fienup, "Efficient subpixel image registration algorithms," Opt. Lett. 33, 156-158 (2008)) and (very) inspired by the [matlab code](http://www.mathworks.com/matlabcentral/fileexchange/18401-efficient-subpixel-image-registration-by-cross-correlation) provided by the authors.

Function `subpixelshift(img,rowShift,colShift)` translates an image by the given amount. `dftReg(imgser,[ref,ufac,align])` performs the registration of `imgser` to a reference `ref` (by default the first frame of `imgser`). Either returns an array of registration parameters (by default), or the registered image (if `align==true`). `usfac` is the upsampling factor (10 by default). Finally `alignFromDft` performs the registration given an array of registration parameters.

This doesn't work with color images for now.

As a side-note, this would also work for rotations by transforming the image to polar coordinates beforehand.

``` julia
> using Images, TestImages
> img = testimage("mandril_gray");
> img2 = subpixelshift(img,40.3,30.7);
> dftReg(img2,ref=img,ufac=10)
       1x3 Array{Float64,2}:
       0.332617  -40.3  -30.7
```

### DIFF
--- a/src/dftRegistration.jl
+++ b/src/dftRegistration.jl
@@ -45,8 +45,7 @@ function dftRegfft(reffft,imgfft,usfac)
         rfzero = sumabs2(reffft)
         rgzero = sumabs2(imgfft)
         error = sqrt(abs(1 - CCmax*conj(CCmax)/(rgzero*rfzero)))
-        diffphase = atan2(imag(CCmax),real(CCmax))
-        output = ["error" => error, "diffphase" => diffphase]
+        output = error
     elseif usfac==1
         ## Whole-pixel shift - Compute crosscorrelation by an IFFT and locate the peak
         L = length(reffft)


jianghaizhu:
Has the license issue been solved yet?

romainFr:
No, the matlab code is still without license... The alternative would be to recode from the description in the paper (not translating their code). As a first step we can add simple phase correlation (without the subpixel part). I'll try to do it when I get a chance.

rsrock:
I don't think you can do it.  By looking at the Matlab code, you're effectively "tainted".  Someone else will have to pick it up.

jianghaizhu:
I sent an email to the senior author of the paper and asked for the clarification of the license for the Matlab code. He told me that he doesn't know anything about Matlab licensing, but pointed me to [a Python version] (http://scikit-image.org/docs/dev/auto_examples/plot_register_translation.html), which has a modified BSD license. I guess this can be merged to master now?

timholy:
Is the author of the python code the same as the author of the matlab code? From a legal perspective, it doesn't really matter if there is _other_ code out there based on the same paper; what matters is the license of whatever code was used as the foundation for this work.

Someone could start from scratch and implement something based on the python code, if the matlab license situation does not resolve.

jianghaizhu:
Just looked at the python code. It stated that it was ported from [the Matlab code] (http://www.mathworks.com/matlabcentral/fileexchange/18401-efficient-subpixel-image-registration-by-cross-correlation). I don't know who the author of the python code is. But the senior author of the original paper believes the license for the python code is fine.

bjarthur:
upon my urging, the original matlab code is now BSD licensed by the author of the manuscript describing the algorithm:

http://ch.mathworks.com/matlabcentral/fileexchange/18401-efficient-subpixel-image-registration-by-cross-correlation

so i think we can go ahead and merge this.

romainFr:
As my understanding is that Images.jl is really more for primitives I reorganized the code and put it together as a small package here :

 https://github.com/romainFr/subpixelRegistration.jl

The changes -- compared to the original version -- allow to run the registration on arrays of arbitrary size (for example if one wants to register image volumes over time). As far as my testing goes, it works with Images or Arrays.

timholy:
Image registration is a big topic, and I confess I'm sort of inclined to make it a separate package. I see that's what @romainFr has done. Seem OK to others? Perhaps we could host it on JuliaImages, just so it's easier to find?

bjarthur:
+1 for making `subPixelRegistration` easier to find by hosting on `JuliaImages`

ashwani-rathee:
I doubt we are ever going to merge this, can we close this PR?

PR #249, opened 2015-01-19
Parse the ImageMagick binaries page to extract URL

timholy:
It looks like the proper URL for downloading ImageMagick on Windows has changed yet again. In an attempt to reduce the frequency of breakages, I've tried to parse the downloads page to extract the currently-valid URL. We'll see what AppVeyor does with this.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -27,12 +27,17 @@ end
 @windows_only begin
     const OS_ARCH = (WORD_SIZE == 64) ? "x64" : "x86"
 
-    # Will need to be updated for releases
     # TODO: checksums: we have gpg
-    magick_exe = "ImageMagick-6.9.0-3-Q16-$(OS_ARCH)-dll.exe"
+    # Extract the appropriate filename to download
+    magick_base = "http://www.imagemagick.org/download/binaries"
+    binariesfn = download(magick_base)
+    str = readall(binariesfn)
+    pattern = ">ImageMagick-6.9.*-Q16-$(OS_ARCH)-dll.exe"
+    m = match(Regex(pattern), str)
+    magick_exe = convert(ASCIIString, m.match)[2:end]
 
     magick_tmpdir = BinDeps.downloadsdir(libwand)
-    magick_url = "http://www.imagemagick.org/download/binaries/$(magick_exe)"
+    magick_url = "$(magick_base)/$(magick_exe)"
     magick_libdir = joinpath(BinDeps.libdir(libwand), OS_ARCH)
 
     innounp_url = "https://julialang.s3.amazonaws.com/bin/winnt/extras/innounp.exe"


timholy:
@tkelman, I notice that older versions of branches remain in the AppVeyor queue when I push a new file. Is there any way to get it to automatically cancel an older queued item if I update the branch? The most egregious example of this is the following: when I push to a new branch, it starts an AppVeyor run. Then when I click "Create pull request" (or whatever that message is), it creates a second queue item. I think Travis does the same thing, but it cancels the previous build when you do that.

tkelman:
Actually I don't think Travis cancels old builds, it just lets you build multiple at the same time.

So there are a few things you can do to mitigate this. First, add a `branches: only:` whitelist for master and any other important release branches (do you have any of those here?), so it won't build on your internal test branches unless you open a PR for them.

For cancelling old builds when there are newer commits in a PR, I do have some code to do this in the base Julia repo - see https://github.com/JuliaLang/julia/blob/239ca2f2c6ce574a2a37644db1d9ade920f6e330/contrib/windows/msys_build.sh#L12-L26

tkelman:
Note that I have that code set up to run at the start of the build, it won't automatically cancel an in-progress build. You'd have to do that from somewhere else - set up a new webhook that goes to some other always-on machine, which checks the AppVeyor API for running builds from the same PR when it receives a new hook, and send the cancel post. You'd need a secure way of storing an authentication token here. Maybe it would be possible to have Travis do this part when it starts running...

timholy:
I didn't initially understand what you mean by `branches: only:`, but I think they have a browser-based method for setting the same thing. We'll see if that works.

Given that change, basically you're saying that I could add this:
```sh
build_script:
  - sh --login /c/projects/Images/deps/windows/cancel_if_newer.sh
```
to `appveyor.yml`, and copy the code you linked above, and that would cover it?

timholy:
(I'm fine with the idea of checking when a build is about to start whether there are newer builds available.)

tkelman:
Right you can do a branch whitelist in the web UI too, dunno which takes precedence. See base Julia's `appveyor.yml` for reference: https://github.com/JuliaLang/julia/blob/f9792e76b696a325ade3996743e059cbb639086f/appveyor.yml#L9-L13

Also check a few lines lower where I'm doing
```
cache:
  - jq.exe
```
to avoid failure if it can't download the CLI json parser for some reason. Could also do the API / JSON queries in powershell, but if you've never used powershell before then now's probably not the time to start.

tkelman:
And you'll want to change `StefanKarpinski/julia` to `timholy/images-jl`

lucasb-eyer:
In other news:

> ImageMagick version 7 is available now as an Beta release. Look for an official release around 1st Q 2016

But do not despair!

> Once ImageMagick version 7 is released, we will continue to support and enhance version 6 for a minimum of 10 years.

I'm basically just saying :+1: 

([source](http://www.imagemagick.org/script/porting.php))

PR #251, opened 2015-01-19
AppVeyor: fail early if there is a newer one queued in this PR

timholy:


appveyor.yml
```diff
@@ -18,6 +21,8 @@ install:
 build_script:
 # Need to convert from shallow to complete for Pkg.clone to work
   - IF EXIST .git\shallow (git fetch --unshallow)
+# if there's a newer build queued, cancel this one
+  - sh --login /c/projects/Images/deps/windows/cancel_if_newer.sh
```
tkelman: err sorry, I think the checkout folder will probably be `/c/projects/images-jl`

tkelman:
and apparently the web-ui whitelist didn't work, https://ci.appveyor.com/project/timholy/images-jl/build/1.0.24 is a branch build, https://ci.appveyor.com/project/timholy/images-jl/build/1.0.25 is the PR build

appveyor.yml
timholy: What a perfect opportunity to test this functionality! I pushed a fixed, then did a squash and re-pushed.

timholy:
I noticed that too. I already have a fix queued, but I'll wait to see how this aspect works before I add it to this PR.

timholy:
I don't understand this: https://ci.appveyor.com/project/timholy/images-jl/build/1.0.26/job/a2n0qwitmjy22f9i

timholy:
Oh, was that just from my force-push?

tkelman:
> What a perfect opportunity to test this functionality!

Eternal optimist :)

The queue on the free plan is much less responsive than on the pro plan, kind of a bummer.

> Oh, was that just from my force-push?

Yeah, the commit doesn't exist any more by the time it tried to check it out...

tkelman:
That was a branch build, which always checks out a specific sha. A PR build starts a little differently, by doing:

```
git fetch -q origin +refs/pull/251/merge:
git checkout -qf FETCH_HEAD
```
So that should still work.

timholy:
Can the 32-bit build cancel the corresponding 64-bit build, too? What seems to happen is:
- the 32-bit queue is shorter, and so that starts first
- the 32-bit job fails quickly
- one waits quite some time for the 64-bit build to get to the top of its queue; this stalls any further 32-bit builds
- once the 64-bit build reaches the top of its queue, it fails quickly
- only then can the next one in line start, but presumably by that point one is at the end of the queue again

tkelman:
They're not actually running in different queues. It's just setting an environment variable differently, which downloads 32-bit Julia instead of 64-bit Julia in subsequent commands. The VM is always 64 bit.

I'm not really sure how the queue works exactly, other than for the pro plan that base Julia is on, builds start one right after the other with at most a few seconds delay. The free plan VM's seem to take a few minutes. And sometimes cancelling builds doesn't seem to free things up completely as far as the queue is concerned, but maybe I'm imagining that.

timholy:
I suspect each project has its own queue, and once one job finishes then the next one gets to go to the bottom of the "global build queue."

timholy:
The logic works, though: https://ci.appveyor.com/project/timholy/images-jl/build/1.0.27/job/xfr5cmd5x1j7incp

tkelman:
Saw that. You might want to move it to before downloading and installing the Julia binary to finish even faster.

Also ref line 85 of http://www.appveyor.com/docs/appveyor-yml, you can add a `fast_finish` to the matrix to skip the 64-bit build if the 32-bit build fails. That would make sense here given the longer queue times.

timholy:
> You might want to move it to before downloading and installing the Julia binary to finish even faster.

I have the impression it gets Images.jl via `Pkg.clone`, so I'm not sure I can move it before the julia binary. (That script is inside Images.jl.) But it's a good idea, so please do correct me if I'm wrong.

tkelman:
> I have the impression it gets Images.jl via `Pkg.clone`

Not exactly. It does `Pkg.clone` from a local filesystem path to register it with Julia's package manager in order to handle dependencies, make `Pkg.build` and `Pkg.test` work, etc. But AppVeyor (like Travis) does a `git clone` from the repo first thing. Try just moving the line that runs the script to first-thing under `install:`.

https://ci.appveyor.com/project/timholy/images-jl/build/1.0.28/job/91ch4c69dthbspfs is running the whole deal because it is a branch build, so `$APPVEYOR_PULL_REQUEST_NUMBER` is not set. We could implement similar logic for non-PR builds, I just haven't done so yet. This is also a planned feature by AppVeyor https://github.com/appveyor/ci/issues/38 but dunno when they'll get to it.

timholy:
This is looking promising; we may have reached a local optimum.

Is there any reason for me to contact AppVeyor with what we've discovered here and see if they want to do something to make all this easier?

tkelman:
Sure. We could just ping `@FeodorFitsner` and let him know he might find our discussion here (and in https://github.com/timholy/Images.jl/pull/249) worth skimming over, unless you were planning on summarizing it.

I also learned a bit more powershell over the holidays (was messing around with some xml stuff), the API query here should be pretty simple to translate and avoid depending on that little json parser and adding a bash script to the repo. `jq` is available cross-platform though and I figure the Julia community is a little more familiar with bash than powershell, so my code is easier to read this way (for us at least).

timholy:
Pinging @FeodorFitsner as recommended above by @tkelman. @FeodorFitsner, the aim here was to reduce the load on your servers, which I imagine is something you might appreciate. Very briefly, we added a shell script to test whether there was a more recent version of the same PR already queued, and if so to fail early (and trigger the failure across all builds). I wonder if "good citizens" who use these tricks might not lose their place in line for the commits they really care about :smile:.

FeodorFitsner:
Yeah, seems like a right logic. I still don't have an ETA for this feature.

mronian:
Awesome!

PR #252, opened 2015-01-27
Update to new bumped version of ImageMagick

axsk:
new bumped version of ImageMagick according to http://www.imagemagick.org/download/binaries/


axsk:
Woopsa, u changed more then I saw, I'm closing this...

timholy:
Thanks for going to the effort to fix this. Hopefully manual version bumps won't be necessary in the future.

PR #259, opened 2015-02-25
Use relative paths instead of Pkg.dir("Images")

tkelman:
this allows installing in a non-default location

ref https://github.com/JuliaLang/julia/issues/8679#issuecomment-75796502

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -22,7 +22,7 @@ filesrc = Array(String, 0)
 function _loadformat(index::Int)
     filename = joinpath("ioformats", filesrc[index])
     if !isfile(filename)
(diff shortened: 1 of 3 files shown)


timholy:
Hmm, the version of ImageMagick on Travis seems to have a transparency bug. And AppVeyor seems to be finicky about redefining writedir. At least the latter is something I can fix.

timholy:
Have you tested this on your machine?

swt30:
I just cherry-picked this commit onto v0.4.21 and tested on my machine with a standard install path - no problems there.

tkelman:
It appears to work fine in the standard install path on Linux and Windows. Have not tested installing outside of `Pkg.dir` but I believe this could only help in that scenario.

timholy:
Thanks!

PR #264, opened 2015-03-05
RFC: strip ColorType information in imgradients return type (fixes #263)

### LINKED ISSUE #263: magnitude inappropriately calls hypot
Does not appear this should have ever worked:
```julia
function hypot{T<:FloatingPoint}(x::T, y::T)
```
For example:
```julia
blur = imfilter_gaussian(twob, [50,50])
gx, gy = imgradients(blur)
magnitude(gx, gy)
```
>`hypot` has no method matching hypot(::Image{RGB{Float64},2,Array{RGB{Float64},2}}, ::Image{RGB{Float64},2,Array{RGB{Float64},2}})
while loading In[74], in expression starting on line 3 
>in magnitude at /home/juser/.julia/v0.3/Images/src/edge.jl:95

kmsquire:
* ~~Used by magnitude, orientation functions~~
* ~~Added Image{Gray} tests to test/edge.jl~~

~~I believe these will work, but I've only tested on Julia v0.4 so far, and only on gray-scale images.~~

**Edit:** The best fix ended up being modifying the behavior of `imgradients` to strip out ColorValue information.

### DIFF
--- a/src/edge.jl
+++ b/src/edge.jl
@@ -91,6 +91,11 @@ function imgradients(img::AbstractArray, method::String="ando3", border::String=
     return grad_x, grad_y
 end
 
+function imgradients{T<:ColorValue}(img::AbstractArray{T}, method::String="ando3", border::String="replicate")
+    # Remove ColorValue information
+    imgradients(reinterpret(eltype(eltype(img)), img), method, border)
+end
+
 # Magnitude of gradient, calculated from X and Y image gradients
 magnitude(grad_x::AbstractArray, grad_y::AbstractArray) = hypot(grad_x, grad_y)
 
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -49,8 +49,8 @@ end
 (./)(img::AbstractImageDirect, A::AbstractArray) = shareproperties(img, data(img)./A)
 (.^)(img::AbstractImageDirect, p::Number) = shareproperties(img, data(img).^p)
 sqrt(img::AbstractImageDirect) = shareproperties(img, sqrt(data(img)))
-atan2(img::AbstractImageDirect) = shareproperties(img, atan2(data(img)))
-hypot(img::AbstractImageDirect) = shareproperties(img, hypot(data(img)))
+atan2(img1::AbstractImageDirect, img2::AbstractImageDirect) = shareproperties(img1, atan2(data(img1),data(img2)))
+hypot(img1::AbstractImageDirect, img2::AbstractImageDirect) = shareproperties(img1, hypot(data(img1),data(img2)))
 
 
 function sum(img::AbstractImageDirect, region::Union(AbstractVector,Tuple,Integer))
--- a/test/edge.jl
+++ b/test/edge.jl
@@ -1,6 +1,6 @@
 ### Edge and Gradient related tests ###
 
-using Images, Base.Test
+using Images, Base.Test, Color
 
 EPS = 1e-14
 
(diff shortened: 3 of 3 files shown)


src/colortypes.jl
```diff
@@ -354,6 +354,22 @@ for CV in subtypes(AbstractRGB)
...
         zero{T}(::Type{AlphaColorValue{$CV{T},T}}) = AlphaColorValue{$CV{T},T}(zero(T),zero(T),zero(T),zero(T))
         zero{T}(::Type{AlphaColor{$CV{T},T}}) = AlphaColorValue{$CV{T},T}(zero(T),zero(T),zero(T),zero(T))
+
+        atan2{T}(a::$CV{T}, b::$CV{T}) = $CV{T}(atan2(a.r, b.r), atan2(a.g, b.g), atan2(a.b, b.b))
+        atan2{T}(a::AlphaColorValue{$CV{T},T}, b::AlphaColorValue{$CV{T},T}) = AlphaColorValue{$CV{T},T}(atan2(a.c.r, b.c.r), atan2(a.c.g, b.c.g), atan2(a.c.b, b.c.b), atan2(a.alpha, b.alpha))
+        atan2{T}(a::AlphaColor{$CV{T},T}, b::AlphaColor{$CV{T},T}) = AlphaColorValue{$CV{T},T}(atan2(a.c.r, b.c.r), atan2(a.c.g, b.c.g), atan2(a.c.b, b.c.b), atan2(a.alpha, b.alpha))
+        hypot{T}(a::$CV{T}, b::$CV{T}) = $CV{T}(hypot(a.r, b.r), hypot(a.g, b.g), hypot(a.b, b.b))
+        hypot{T}(a::AlphaColorValue{$CV{T},T}, b::AlphaColorValue{$CV{T},T}) = AlphaColorValue{$CV{T},T}(hypot(a.c.r, b.c.r), hypot(a.c.g, b.c.g), hypot(a.c.b, b.c.b), hypot(a.alpha, b.alpha))
+        hypot{T}(a::AlphaColor{$CV{T},T}, b::AlphaColor{$CV{T},T}) = AlphaColorValue{$CV{T},T}(hypot(a.c.r, b.c.r), hypot(a.c.g, b.c.g), hypot(a.c.b, b.c.b), hypot(a.alpha, b.alpha))
```
kmsquire: @timholy, I just copied the format from above, but I'm wondering if `AlphaColorValue` on the RHS should just be `AlphaColor` (here and above)?

kmsquire: Never mind.  I'm about to push a change which makes this totally unnecessary.

src/edge.jl
```diff
@@ -91,6 +91,11 @@ function imgradients(img::AbstractArray, method::String="ando3", border::String=
     return grad_x, grad_y
 end
 
+function imgradients{T<:ColorValue}(img::AbstractArray{T}, method::String="ando3", border::String="replicate")
+    # Remove ColorValue information
+    imgradients(reinterpret(eltype(eltype(img)), img), method, border)
```
kmsquire: @timholy, is there a better way to remove the ColorValue type?

timholy:
Yes, that's a highly-recommended approach. Basically the color types can be thought of as labeling the meaning of the data in a dispatch-friendly fashion (or, will be once we get triangular dispatch :smile:). But it's often easiest to do computations on the "raw" numbers. It's just that we have to support computation on the various types because it's guaranteed that people will do so.

src/edge.jl
timholy: I should just add a more convenient method, like `raw(img)`.

kmsquire:
Great, thanks!  Will merge after tests run (though I'm not sure they're passing right now in general).

kmsquire:
I find it interesting that the Windows tests pass, but the OSX and Linux tests are failing on IO. :-/

Anyway, the relevant tests (`test/edge.jl`) pass for me locally on v0.3 and v0.4, and the failure doesn't seem relevant (also fails for me locally), so I'm going to merge.

timholy:
It's some weird problem with Travis' imagemagick. I even have a machine with the same Ubuntu version, and it passes there. I should get back to trying to figure out what's up.

src/edge.jl
timholy: `raw` is now available

PR #265, opened 2015-03-13
When imwrite is passed a file name with a .nhdr extension its behavior i...

Cody-G:
...s now different than when passed a name with a .nrrd extension.  .nhdr now only writes an nrrd header, and throws an error when the datafile property is not specified. When passing the .nhdr extension it is no longer necessary for the user to specify the headeronly property, and the props keyword arg is now set to img.properties by default.  When passed a .nrrd extension, imwrite behaves the same as before.

Note that this breaks at least one nrrd test in the Images test suite.  If you think these changes are good, I am happy to fix the broken test as well.

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -590,7 +590,9 @@ add_image_file_format(".dummy", b"Dummy Image", Dummy, "dummy.jl")
 # NRRD image format
 type NRRDFile <: ImageFileType end
 add_image_file_format(".nrrd", b"NRRD", NRRDFile, "nrrd.jl")
-add_image_file_format(".nhdr", b"NRRD", NRRDFile, "nrrd.jl")
+# NRRD header only
+type NRRDHeader <: ImageFileType end
+add_image_file_format(".nhdr", b"NRRD", NRRDHeader, "nrrd.jl")
 
 # Andor Technologies SIF file format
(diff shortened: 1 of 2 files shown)


Cody-G:
Actually I the breakages are more serious than I thought.  Since the magic bytes of .nrrd and .nhdr files are the same, they are assigned the same extension by the add_image_file_format function.  I could make further changes to fix that, but I'm starting to think that the fuzzy distinction between the .nhdr and .nrrd extensions is due to the design of the nrrd format itself.  Maybe we should just keep things the same, though it's a little unsatisfying that the .nhdr and .nrrd extensions are handled exactly the same way.  What do you think?

timholy:
The Travis failure is a weird problem with Travis' libimagemagick; I can't even reproduce it on another machine that runs Ubuntu 12.04. And the log on appveyor seems incomplete, I don't see where the actual error takes place (unless I'm just missing something).

The magic number thing is an interesting challenge. Do you have a "demo" of how you'd like this to look from a user-perspective? (And how it looks now?) That might help us decide whether it's worth revamping this.

Cody-G:
Sorry I lost track of this.  I think it's best to stick with the current implementation for now.  Closing.

PR #266, opened 2015-03-14
fix for cross-platform bindeps

rened:
Fix to support the https://github.com/JuliaLang/BinDeps.jl/pull/130 PR

(tests will fail without that PR)

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -79,12 +79,13 @@ end
     """ )
 end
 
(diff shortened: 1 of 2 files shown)


timholy:
Oh, sorry this went stale. If it's still relevant, [ImageMagick.jl](https://github.com/JuliaIO/ImageMagick.jl) is the place for this.

PR #270, opened 2015-03-27
Fix LoadError with latest Julia

simonster:
Since JuliaLang/julia@6e06078e1ba8ef517a89518c632ca96c3a704464 I get:
LoadError: syntax: space not allowed after ":" used for quoting

This fixes it so the package loads, but there are still a lot of deprecation warnings.


timholy:
Nice. Just today I started tackling this from the other direction; I've got depwarns fixed in FixedPointNumbers.

PR #274, opened 2015-04-03
Fixes nnrd output issue #272

cbecker:

PR #276, opened 2015-04-03
Silence a failing test for image io

rsrock:
At the moment, it's unclear whether ImageMagick's at fault or the
OS X reader.  This test also fails on Travis.


rsrock:
Ah, looks like you beat me to it.  I set up a pull request just to see how the tests would run after this change.

timholy:
Sorry, I was just trying to get testing working for #274.

Issue #273, opened 2015-04-02
MemoryError() when writing pixel value


cbecker:
Hello, I am getting a very weird error I haven't seen before,
```julia
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "help()" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.3.8-pre+2 (2015-03-25 14:33 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit af187c5* (8 days old release-0.3)
|__/                   |  x86_64-unknown-linux-gnu

julia> using Images

julia> im=imread("posweights.nrrd");

julia> im[1,1,1] = 10
ERROR: MemoryError()
 in setindex! at /home/cjb/.julia/v0.3/Images/src/core.jl:295

julia> im
Gray Image with:
  data: 749x617x308 Array{Float32,3}
  properties:
    colorspace: Gray
    space: left-posterior-superior
    spatialorder:  x y z
    spacedirections:  [1.0,0.0,0.0] [0.0,1.0,0.0] [0.0,0.0,1.0]
    comments:  Complete NRRD file format specification at: http://teem.sourceforge.net/nrrd/format.html
    pixelspacing:  1.0 1.0 1.0
```

I am not sure how to proceed to debug it further, it is a common float32 nrrd file.
Let me know if I can run other diagnostics to help find out what is going on.

cbecker:
This seems to be because Images is memory mapping the contents because `prod(size(im)) > 10^8`, as from https://github.com/timholy/Images.jl/blob/master/src/ioformats/nrrd.jl#L126 , which is actually amazing since I didn't know it was already doing so.

There are a few improvements that could help handling mmaped arrays
  * Add an option like `force_mmap` to `imread()`, which would force memory mapping, if available
  * Add a field in the `Image` type to indicate that the data may be read-only
  * I don't know how to fix this, but julia showing `MemoryError` when trying to write to an array elemen tis not very descriptive, but this is beyond `Images.jl`

The first two seem easy to implement and I can take care of them, let me know if you agree or if you see potential issues.

timholy:
If one wants to change values, the file will need to be opened with read/write permissions. Note that with `mmap`, `A[i,j,k] = v` will *change the data on disk*.

cbecker:
yes, I agree.

[...]

PR #278, opened 2015-04-06
imread() now supports mmap argument

cbecker:
This is a working prototype that required minimal changes in the code, improves #273


timholy:
This is awesome, thanks very much!

PR #289, opened 2015-04-26
Minor change to remove deprecation warning.

kmsquire:
The warning was
```julia
WARNING: convert{T}(p::Type{Ptr{T}},a::Array) is deprecated, use convert(p,pointer(a)) instead.
```
However, it seems to me that using a pointer in the first place was probably unnecessary.  @rsrock, am I missing something?


kmsquire:
Answer: looking at the build log, yes, I was missing something!

timholy:
LGTM, OK to merge @rsrock?

rsrock:
LGTM too

Issue #284, opened 2015-04-19
Windows- or 32bit-specific `InexactError`


timholy:
Appveyor builds are failing:
https://ci.appveyor.com/project/timholy/images-jl/build/1.0.83/job/sgy1gmtq36rbsemh
https://ci.appveyor.com/project/timholy/images-jl/build/1.0.80/job/wv5yw4qxj0fvugqw
on the following command:

```
using FixedPointNumbers
aa = Ufixed12[0.6 0.2;
              1.4 0.8]
```

Can anyone on Windows test this? I don't understand what the problem could be.

elsuizo:
In my 32-bits Debian laptop with Julia 0.3.7 works fine Tim

timholy:
Thanks for the useful data point!

timholy:
Ah, but it will probably only fail on 0.4---`trunc` didn't throw errors on 0.3.

tkelman:
`Ufixed12[0.6 0.2; 1.4 0.8]` in isolation doesn't error for me on win32, 0.4, but it does give different answers every time I run it.

`Pkg.test("Images")` fails for me on win32, 0.4, with the same error you're seeing on appveyor.

timholy:
That's alarming. Sounds like some kind of memory error?

I'm going to file this at FixedPointNumbers and CC some folks who know more. Just to double-check, this works for you on Linux, right? (You have Linux available, right?)

timholy:
Sorry to pester you more, but can you also test things like:
```jl
Ufixed(0.2)
Ufixed(1.4)
Ufixed([0.2,1.4])
```
and let me know what happens?

tkelman:
> Just to double-check, this works for you on Linux, right? (You have Linux available, right?)

Yes and yes.

> Ufixed(0.2)
> Ufixed(1.4)
> Ufixed([0.2,1.4])

```MethodError: `rawtype` has no method matching rawtype(::Type{FixedPointNumbers.Ufixed})` ```

With `Ufixed12(0.2)` things appear to work. Worth noting that for the array `Ufixed12[0.6 0.2; 1.4 0.8]` it looks like the first row is always zeros, the second row has different values every time.

timholy:
Sorry for initially omitting the 12, and thanks for catching it.

timholy:
I think I'm going to file this in julia proper, since I don't think this can be limited to FixedPointNumbers.

tkelman:
[...]

PR #290, opened 2015-04-26
Fixed ERROR: imresize_julia! not defined

ktisha:
Call to imresize() resulted in ERROR: imresize_julia! not defined
Bug introduced in f203853c5b8dafc2115a1a7ba29a31ceeda18892
Initially function imresize was named imresize_julia (see https://groups.google.com/forum/#!msg/julia-users/wG1g9C7S6Mg/zx9Pqntv1e0J) to compare different imresize approaches

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1037,7 +1037,7 @@ function imresize!(resized, original)
     resized
 end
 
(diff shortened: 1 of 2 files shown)


timholy:
Thanks for the fix! The AppVeyor failure is #284 and not your problem.

PR #293, opened 2015-05-01
Fix nrrd write issue when eltype is Gray{UfixedBase{...}}

cbecker:
Hi Tim.
I noticed there was a problem with a previous PR. The problem is that line 295 in nrrd.jl was not working if UfixedBase is wrapped around another type, like Gray{}.


cbecker:
Hi Tim, I also have another PR to add the `raw()` functionality, let me kno

timholy:
Sorry, I didn't merge this because of the test failures...but I see it has nothing to do with this PR.

timholy:
See https://github.com/Keno/SIUnits.jl/pull/59, which should fix the failure.

Issue #190, opened 2014-09-24
Error opening supposedly-32-bit Tiff


cbecker:
I am getting the following error when trying to open a 32-bit Tiff:

``` julia
exception on 1: ERROR: key not found: 32
 in imread at .julia/v0.3/Images/src/io.jl:241
 in imread at .julia/v0.3/Images/src/io.jl:113
ERROR: `convert` has no method matching convert(::Type{TimeInstantData}, ::KeyError)
 in setindex! at array.jl:307
 in copy! at abstractarray.jl:149
 in convert at array.jl:220
 in include at ./boot.jl:245
 in include_from_node1 at ./loading.jl:128
```

Tim, I cannot post the file here, but I sent you an email with an example.
Thanks.

PR #297, opened 2015-05-05
Added raw(img) to wrap the data out of UfixedBase, #190

cbecker:
This adds the raw(img) functionality, as discussed in #190

### DIFF
--- a/doc/core.md
+++ b/doc/core.md
@@ -100,7 +100,9 @@ The `data` array here just encodes the index used to look up the color in the
 For any valid image type, `data(img)` returns the array that corresponds to the
 image.  This works when `img` is a plain `Array` (in which case no operation is
 performed) as well as for an `Image` (in which case it returns `img.data`).
-This is our first example of how to write generic algorithms.
+For some image formats, Images.jl may interpret raw data with the `FixedPointNumbers`
+package. The function `raw(img)` can be used to recover the buffer in its raw format
+(e.g. `UInt8`). This is our first example of how to write generic algorithms.
 
 If `img` is an `Image`, then `img[i,j]` looks up the value `img.data[i,j]`.
 Assignment, `sub`, and `slice` work similarly. In other words, for indexing an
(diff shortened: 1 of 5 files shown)


timholy:
Can I ask that you add a couple of tests, too? In particular, it's worth checking this works for both grayscale and `RGB` images. It _should_ work because of https://github.com/timholy/Images.jl/blob/79eb3551b9c8dd06b414ba21a65b021ebd76354a/src/core.jl#L129-L162, but it's worth double-checking.

Adding documentation to https://github.com/timholy/Images.jl/blob/master/doc/function_reference.md and/or the README would be a bonus!

If you wait until https://github.com/JuliaLang/METADATA.jl/pull/2549 is merged, then the tests should pass again.

timholy:
Actually, they may not pass due to https://github.com/JuliaLang/julia/issues/10513#issuecomment-99119291.

cbecker:
sounds good, I will add tests and docs.

Currently I am avoiding julia v0.4, i'm afraid of having to spend too much time porting my code, as there are many code breaking changes.

timholy:
Nothing you need to worry about here. While you don't have to test 0.4 locally, Travis tests it for you. I noticed it was broken, so dug into it and discovered SIUnits needed updating.

The parallel test (the last one) may still be broken on 0.4, however. Again, nothing you need to worry about.

cbecker:
Hi Tim, I added the docs and some tests.
So far `raw()` expects an image as its argument, so it will throw an error if called with an array, for example. This contrasts with `data()`, shall we then define a fallback for arrays, as in `data()`?
```julia
julia> data(zeros((2,3)))
2x3 Array{Float64,2}:
 0.0  0.0  0.0
 0.0  0.0  0.0

julia> raw(zeros((2,3)))
ERROR: `raw` has no method matching raw(::Array{Float64,2})
```

timholy:
Nice! Yes, I agree a fallback for arrays would be good. All you have to do is change the signature to `raw(img::AbstractArray)` and you should be good to go.

cbecker:
Done, I rebased the PR as well.
I had to add a particular case for `raw([])`.

timholy:
Thanks!

PR #308, opened 2015-06-16
Don't clamp by default before writing (fixes #256)

### LINKED ISSUE #256: Float32 to Ufixed16 conversion fails when creating an image
I have a Float32 matrix

    (typeof(result_scaled),size(result_scaled)) = (Array{Float32,2},(512,512))

and trying to convert it to a 16-bit greyscale image like this:

    umin, umax = quantile(vec(result_array), [0.025, 0.975])
    result_scaled = (result_array - umin) / (umax - umin)
    image = Image{RGB{Ufixed16}}(result_scaled)

fails with a very uninformative error message. If I make an image from result_array instead of result_scaled, everything works fine, same if I clamp result_scaled using `min(0.f0, max(1.f0, result_scaled))`, it works too. Note that Images doesn't seem to care if a Float32 input is out of [0,1], so I didn't expect it to fail with Ufixed16.

    ERROR: InexactError()
     in trunc at float.jl:345
     in copy! at abstractarray.jl:174
     in convert at array.jl:233
     in _convert at /***/.julia/v0.4/Images/src/core.jl:277
     in go2 at /***.jl:190

    julia> versioninfo()
    Julia Version 0.4.0-dev+2876
    Commit f164ac1 (2015-01-22 22:58 UTC)
    Platform Info:
      System: Linux (x86_64-linux-gnu)
      CPU: Intel(R) Xeon(R) CPU E5-2620 0 @ 2.00GHz
      WORD_SIZE: 64
      BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Sandybridge)
      LAPACK: libopenblas
      LIBM: libopenlibm
      LLVM: libLLVM-3.3

timholy:
Current, `imwrite` clamps values before writing, so that images that contain (for example) negative values get truncated at 0. This commit changes the behavior to throw an error, as argued in #256.
 
This also adds a new `mapinfo(Clamp, img)` to allow users to obtain the earlier behavior.

@rsrock, IIUC I don't need to do anything special to have the github.io documentation update, right?

### DIFF
--- a/doc/function_reference.md
+++ b/doc/function_reference.md
@@ -650,7 +650,15 @@ directly. Some writers take additional arguments, for example
 imwrite(img, "myimage.jpg", quality=80)
 ```
 
-to control the quality setting in JPEG compression.
+to control the quality setting in JPEG compression. Another useful keyword argument
+is `mapi`, which allows you apply a MapInfo transformation on the elements
+before writing. For example, if you have floating-point images whose values are
+sometimes out-of-bounds (smaller than 0 or bigger than 1), you can avoid an error
+upon writing with
+```{.julia execute="false"}
+imwrite(img, "myimage.png", mapi=mapinfo(Clamp, img))
+```
+and values will be clamped, as needed, before writing.
 
 ## loadformat
 ```{.julia execute="false"}
--- a/src/io.jl
+++ b/src/io.jl
@@ -370,25 +370,45 @@ end
 
 # ImageMagick mapinfo client. Converts to RGB and uses Ufixed.
 mapinfo{T<:Ufixed}(::Type{ImageMagick}, img::AbstractArray{T}) = MapNone{T}()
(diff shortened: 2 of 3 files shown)


timholy:
CI errors are a failure to find `CGImageSourceGetCount` on OSX (0.4-only) and one that I don't believe I've seen before (GrayAlpha I/O) on windows. Neither are likely to be related to this PR, but of course we should fix those. Sadly, both are platforms for which I lack direct access.

rsrock:
Sorry for the late update, duties got in the way. I just updated the docs with these changes.

To update the docs, here's the procedure.  You'll need Judo.jl, and the `judo` script under `Judo/bin` should be on your path.

1) cd to `Images`
2) `git checkout gh-pages`
3) `git merge master`
4) `judo --package Images && cp -R doc/html/* . && rm -rf doc/html`
5) `git push origin gh-pages`

Gee, I should probably write a script for that at some point...

PR #310, opened 2015-06-26
small fixes to restore nrrd read and write functionality.  Maybe thes…

Cody-G:
…e changes were necessary after a recent change to SIUnits?

I also changed the behavior slightly so that the image properties will be used by default if there is no "props" keyword argument.  (This could probably be done more elegantly by just setting props to img.properties if it is not specified by the user, but I was concerned that might break other code)
Also fixed one depwarn.


src/ioformats/nrrd.jl
```diff
@@ -214,6 +214,8 @@ function imread{S<:IO}(stream::S, ::Type{Images.NRRDFile}; mmap=:auto)
     elseif haskey(header, "space dimension")
         spacedim = int(header["space dimension"])
         spacedim == sum(isspatial) || error("spacedim $spacedim disagrees with isspatial=$isspatial")
+    else
```
timholy: It would seem to be better to handle this in the initialization of `spacedim` on line 209.

src/ioformats/nrrd.jl
```diff
@@ -270,7 +272,7 @@ function imread{S<:IO}(stream::S, ::Type{Images.NRRDFile}; mmap=:auto)
         length(ps) == nd || error("parsing of spacings: $(header["spacings"]) is inconsistent with $nd dimensions")
         pss = ps[isspatial]
         if !isempty(units)
-            unitss = units[isspatial]
+#             unitss = units[isspatial]
```
timholy: Good catch. Probably better to just delete.

src/ioformats/nrrd.jl
```diff
@@ -358,11 +362,15 @@ function imwrite(img, sheader::IO, ::Type{Images.NRRDFile}; props::Dict = Dict{A
         datafilename = get(props, "data file", "")
     end
     if isempty(datafilename)
+        datafilename = get(img.properties, "datafile", "")
+    end
+    if isempty(datafilename)
+        print("No data file name was provided; writing data to header file\n")
```
timholy: Won't this print a warning every time people try to use this to write a single `.nrrd` file (rather than splitting as two files)? That seems like an unnecessary amount of noise.

src/core.jl
```diff
@@ -723,6 +723,7 @@ colordim{C<:ColorType}(img::AbstractArray{C,3}) = 0
 colordim{C<:ColorType}(img::AbstractImage{C}) = 0
 colordim(img::AbstractVector) = 0
 colordim(img::AbstractMatrix) = 0
+colordim(img::AbstractArray) = 0
```
timholy: Could this perhaps be a separate commit? I'm a little worried that adding this fallback will change something else subtly, and having this part of the same commit means bisecting is harder.

timholy:
For me our current tests pass even without this. Can you also add a test that currently fails?

Also, do you understand the segfault on Travis? I can't replicate this locally.

Cody-G:
Okay I modified my original pull request and added a couple more commits, including a test that triggers the problem: imread was expecting an entry for "space units" even for non-spatial dimensions.  Also it seems that the failure on Travis is happening on windows.  So I can't easily test that on my machine.

timholy:
This looks better to me. The AppVeyor failure is indeed Windows (and is due to #316). The Travis failure is on both Linux and OSX, only on nightly (0.4). I can't reproduce it locally (the tests work for for me), and the error seems rather surprising: how the heck does the equivalent of
```jl
Y = similar(X)
@fact Y == X => false
```
ever fail?? I will push a `fill!` to make sure random member doesn't accidentally contain the correct answer.

I'm going to merge, but clearly we have something that needs to be figured out.

timholy:
Thanks, Cody!

Cody-G:
Wow the AppVeyor error is strange!  Thanks for the merge!

PR #314, opened 2015-07-22
Testing revamp

rsrock:
I've been trying to track down some testing issues, and ended up running down this rabbit hole.  I'm hoping it will be worth it in the end.  If not, I'm hoping someone will tell me what sort of problems I didn't anticipate.

In playing with the tests, I found cases where state from one test file was leaking into another.  This was making it difficult to pin down the source of errors, because success or failure depended on which tests I enabled.  Specifically, there was a coupling between `readnrrd.jl` and `readremote.jl` related to the workdir constants.  So, I decided to refactor the tests into groups with limited scope.  While I was at it, I decided to use the FactCheck package.  One nice advantage of FactCheck is that when you group tests using `facts()` /  `context()`, you get a new scope. Another nice thing is the satisfying parade of dots as the tests complete.  My limited testing suggests that the tests are even a smidgen faster with FactCheck.

One big reason that this is WIP is that I took one look at `edge.jl` and decided that I wanted some input on this plan before tackling that one.  I'm sure it can be switched over, but will require more care. @kmsquire.   

The Travis failure on `release` is due to the Base.Graphics -> Graphics migration.  I know I saw a magic incantation somewhere that will pick the right one.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -27,8 +27,9 @@ include("readstream.jl")
 @osx_only include("readOSX.jl")
 include("algorithms.jl")
 include("edge.jl")
-include("parallel.jl") # TODO: some errors, related to "require"?
 
 isinteractive() || FactCheck.exitstatus()
 
 end
+
+include("parallel.jl")
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,16 +1,34 @@
+module ImagesTests
+
+using Compat, FactCheck, Base.Test, Images, Color, FixedPointNumbers
+if VERSION < v"0.4.0-dev+3275"
+    using Base.Graphics
+else
+    using Graphics
+end
+
 testing_units = Int == Int64
+if testing_units
+    using SIUnits, SIUnits.ShortUnits
+end
+
+FactCheck.setstyle(:compact)
 
 include("colortypes.jl")
 include("core.jl")
 include("map.jl")
-include("overlays.jl")
+include("overlays.jl")  # This one can't handle precompiled Color yet
 include("io.jl")
 if testing_units
-    include("readnrrd.jl")
+     include("readnrrd.jl")
 end
 include("readremote.jl")
 include("readstream.jl")
 @osx_only include("readOSX.jl")
 include("algorithms.jl")
-include("edge.jl")
-include("parallel.jl")
+include("edge.jl") # TODO:
+##include("parallel.jl") # TODO: some errors, related to "require"?
+
+isinteractive() || FactCheck.exitstatus()
+
+end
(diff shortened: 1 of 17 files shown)


rsrock:
The mac nightly failure is JuliaLang/Homebrew.jl#90

rsrock:
Looks like this is ready to go.  Wasn't that difficult after all.

In using this the last few days, I like how it runs.  Another feature I forgot to mention is that it will run the entire test suite, rather than stopping at the first error.

I'll let this simmer for a bit-- these are major changes and it's probably best if someone else (@timholy, @kmsquire) takes a look.

kmsquire:
Thanks for doing this!  I'll try to take a look tomorrow.

timholy:
Very interesting. I agree with the problems of leaky scope. In the past I've resisted putting things inside scope containers because it can be nice to have the tests fail in a way that all variables are already available at the REPL for diagnosing the failure. OTOH this is is not the first time this has caused problems, so perhaps it's time to just fix this. (Ref #217).

timholy:
Wow, that's a big change! I confess I didn't read it all carefully; as long as there aren't any tests where you've made it less stringent, I have no problems with this, and look forward to seeing how it works in practice.

I restarted a failing build, let's see if that helps.

test/runtests.jl
```diff
@@ -1,16 +1,34 @@
+module ImagesTests
+
+using Compat, FactCheck, Base.Test, Images, Color, FixedPointNumbers, Graphics
```
timholy: You'll want to delete the Graphics from this line, so you rely on the next ones.

rsrock: Good catch

rsrock:
Good point about working at the REPL. I usually fire off the tests from the command line or using Watcher.jl.  One thing that works with the new layout is just copying all of the lines from a `context` block to the REPL.  In most cases, that gets everything that you need to explore.  If not, there may be a few lines at the top of the file at the start of the `facts` block got grab as well.

If we keep the `context` blocks small, this may not be so bad (most are pretty small already, and some others could be carved up further).

Looks like I need to fix up a couple of merge conflicts.  I'll do that now, and then after tonight I'll test again when the new nightlies are out.

timholy:
I usually use the shell too, until some change I make breaks a test. Happens surprisingly often :smile:. But I agree with the idea of just copy/pasting a context block, and indeed that might make it easier in some ways. So I look forward to this!

Sorry about the merge conflicts, and thanks for fixing.

rsrock:
Any clue what's going wrong with the parallel test?  I get this locally too.  Something to do with wrapping all tests in a single module?

timholy:
Does it happen for you on master?

rsrock:
Thanks, I figured out the parallel issue after comparing with master.

This is now merged in ad4318dc434824

kmsquire:
@rsrock, thanks for doing this!

Issue #316, opened 2015-07-24
Pkg.add("Images") fails on Windows 7


JPhillips4:
I'm getting started with Julia 0.3.10 under Windows 7. The language installs and runs, but I can't install the Images package. When I type:

Pkg.add("Images")

I get output that looks OK, but then the lines below. It looks as if Inno Setup has detected a version problem, but I'm not sure whose version is problematic. I've tried deleting everything "julia" and re-downloading, to no avail. Also, I've searched the web for various pieces of this error message, also stackoverflow; nothing appears to be relevant.

After Pkg.add("Images"), a few dozen normal-looking lines, then:

INFO: Building Images
INFO: Installing ImageMagick library
INFO: Attempting to Create directory C:\Users\Jim\.julia\v0.3\Images\deps\downloads
INFO: Attempting to Create directory C:\Users\Jim\.julia\v0.3\Images\deps\usr\lib\x64
INFO: Attempting to Create directory C:\Users\Jim\.julia\v0.3\Images\deps\downloads
INFO: Directory C:\Users\Jim\.julia\v0.3\Images\deps\downloads already created
INFO: Downloading file [link at imagemagick] download/binaries/ImageMagick-6.9.1-9-Q16-x64-dll.exe
INFO: Done downloading file [...]ImageMagick-6.9.1-9-Q16-x64-dll.exe
INFO: Attempting to Create directory C:\Users\Jim\.julia\v0.3\Images\deps\downloads
INFO: Directory C:\Users\Jim\.julia\v0.3\Images\deps\downloads already created
INFO: Downloading file [link at julialang.s3.amazonaws.com]/bin/winnt/extras/innounp.exe
INFO: Done downloading file [link at julialang.s3.amazonaws.com]/bin/winnt/extras/innounp.exe
INFO: Changing Directory to C:\Users\Jim\.julia\v0.3\Images\deps\downloads

Here's where I see the first sign of trouble:

Signature detected: Inno Setup Setup Data (5.5.6) (u)
This is not directly supported, but i'll try to unpack it as version 5500; Version detected: 5506 (Unicode)
Critical error: The setup files are corrupted. Please obtain a new copy of the program.

Unpacking failed. This version is not supported.

Thanks,
Jim

timholy:
I'm no Windows guru, unfortunately. @ihnorton or @tkelman? (Sorry)

tkelman:
[...]

PR #317, opened 2015-07-24
WIP: Use portable zip download for imagemagick on windows

tkelman:
for #316


tkelman:
apparently there's no libwand there - that's not so useful

PR #320, opened 2015-07-27
Fix Travis FactCheck failures for new parsing of =>

timholy:
Ironically, right after the switch to FactCheck, 0.4 changed the priority of `=>`.


rsrock:
That's a relief, I was wondering what the heck I did wrong there.  LGTM.

PR #325, opened 2015-08-05
make the reading from a stream test unix_only

### LINKED ISSUE #324: io.jl test failure on Windows
Continuing debugging from #316 #323 

```
julia> using Images, Compat

julia> Images.LibMagick.libwand
"C:\\Users\\Tony\\.julia\\Images\\deps\\usr\\lib\\x64\\CORE_RL_wand_.DLL"

julia> wand = Images.LibMagick.MagickWand()
MagickWand(Ptr{Void} @0x000000000017ad10)

julia> workdir = joinpath(tempdir(), "Images")
"C:\\Users\\Tony\\AppData\\Local\\Temp\\Images"

julia>  stream = open(joinpath(workdir, "2by2.png"))
IOStream(<file C:\Users\Tony\AppData\Local\Temp\Images\2by2.png>)

julia> Compat.Libc.FILE(stream).ptr
Ptr{Void} @0x000007feff2c2b10

julia> status = ccall((:MagickReadImageFile, Images.LibMagick.libwand), Cint, (P
tr{Void}, Ptr{Void}), wand.ptr, Compat.Libc.FILE(stream).ptr)
```

And that just crashes Julia. This functionality looks like a relatively recent addition from 3b0d46eb8777c067e08f76c0756dfd3714b22a16. I don't think it's valid to pass FILE objects across shared library boundaries on Windows when the shared libraries were built against different C runtimes - the ImageMagick that we're downloading is built against a recent MSVC runtime (msvcr120.dll), Julia is built against mingw and the old msvcrt.dll that's always present.

Switching to ImageMagick from WinRPM may fix this, because Julia and libwand would be using the same C runtime. I still haven't gotten a very clear story about what happened the last time that was tried, if it ever was. Another option would be to disable that test on windows.

tkelman:
closes #324

### DIFF
--- a/test/io.jl
+++ b/test/io.jl
@@ -127,7 +127,7 @@ facts("IO") do
         @fact B => map(Ufixed8, A)
     end
     
-    context("Reading from a stream (issue #312)") do
+    @unix_only context("Reading from a stream (issue #312)") do
         fn = joinpath(workdir, "2by2.png")
         io = open(fn)
         img = Images.imread(io, Images.ImageMagick)
--- a/test/readremote.jl
+++ b/test/readremote.jl
@@ -49,7 +49,7 @@ facts("Read remote") do
         @fact ndims(img) => 2
         @fact colordim(img) => 0
         @fact eltype(img) => Images.ColorTypes.GrayAlpha{Ufixed8}
-        @osx? nothing : begin
+        @linux_only begin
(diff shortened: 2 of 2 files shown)


timholy:
The remaining test failures seem to be fixed by #326. OK to merge, right?

PR #327, opened 2015-08-05
Expand the credits

timholy:
Also a nice chance to see if we're green across the board.


timholy:
Oh, so nice to see everything passing again!

tkelman:
:cake:

timholy:
You should eat it :smile:

PR #328, opened 2015-08-05
Fixed imread for new parsing of =>

Cody-G:
Due to the recent change in parsing of => , imread was failing to create properties dict.  Simple fix.


timholy:
Thanks!

PR #330, opened 2015-08-11
Fix #329

### LINKED ISSUE #329: Error updating on Windows
```LoadError: type Void has no field match
while loading C:\j\.julia\v0.4\Images\deps\build.jl, in expression starting on line 37```

This happens because of the ```>``` char in 

```pattern = ">ImageMagick-6.9.*-Q16-$(OS_ARCH)-dll.exe"```

If remove it, than

```julia> magick_exe = convert(ASCIIString, m.match)```

"ImageMagick-6.9.1-10-Q16-x64-dll.exe"

joa-quim:
Update the parsing of download page to account for new change in imagemagick URL

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -32,9 +32,9 @@ end
     magick_base = "http://www.imagemagick.org/download/binaries"
     binariesfn = download(magick_base)
     str = readall(binariesfn)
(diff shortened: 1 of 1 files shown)


timholy:
I don't understand the remaining Windows error, but this is clearly progress. Many thanks, @joa-quim!

PR #338, opened 2015-08-16
disable precompilation on windows

tkelman:
may help with the RegistryKeyLookupFailed `CoderModulesPath' error?

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -1,4 +1,4 @@
-VERSION >= v"0.4.0-dev+6521" && __precompile__()
+VERSION >= v"0.4.0-dev+6521" && __precompile__(@unix? true : false)
 
 module Images
 


timholy:
Yay, thanks a bunch!

timholy:
(Bummer, of course.)

timholy:
Hmm, just saw https://github.com/JuliaLang/julia/issues/12638. Once that's available, we might try reverting this?

tkelman:
The commit that ran, at least on win32, was https://github.com/JuliaLang/julia/commit/6becc96 so would have had that fix. win64 said it was running on https://github.com/JuliaLang/julia/commit/58408ff which is a bit confusing since that was a commit from inside a PR? Anyway good reason to always have `versioninfo()` somewhere in the CI logs.

Can always try reverting this periodically and see if it helps. I suspect trying the WinRPM version of imagemagick may be more effective though, if the current library is trying to access something from the registry in its initialization? Maybe unfixable though, I'm not really sure.

flicaflow:
I just enabled precompilation manually on Windows for the Images package. I do not see any errors, tests are succeeding . Maybe this change could be reverted?

SimonDanisch:
I did that as well on my fileio branch... Seems to work fine for my use cases as well!

timholy:
Keep an eye on #358 and we'll see how it goes.

PR #340, opened 2015-08-17
Rebase on Colors/ColorTypes/ColorVectorSpace

timholy:
This is a heads-up PSA.

I'm in the process of cleaning up our handling of colors---we had some incompatibilities with different packages using different implementations of the basic color definitions, and `Color.jl` itself was due for an overhaul (more about that elsewhere, later). I've got a replacement (not yet posted) called `Colors.jl`, and I am farming out the colorspace-math to `ColorVectorSpace.jl` (also not yet posted). But I wanted to put up this fairly-massive patch in part to get a second copy out there somewhere, and while at it I thought I'd give folks fair warning about what's coming. You can see the new color names I'm proposing [over here](https://github.com/timholy/ColorTypes.jl/tree/teh/traits) (which is currently a pull request against [ColorTypes.jl](https://github.com/SimonDanisch/ColorTypes.jl), which was split out from the internal `ColorTypes` module in Images.).


timholy:
Combined testing here: https://travis-ci.org/timholy/Dummy.jl. Assuming this is on-track tomorrow, I'll merge.

PR #342, opened 2015-08-20
Add additional libmagickwand package name for ubuntu 15.04

c42f:
I'm in no way a BinDeps expert, but this allowed me to satisfy the build process for Images.jl on ubuntu 15.04 with julia-0.3.10.  The proliferation of version numbers in ubuntu package names is a bit confusing, given that 15.04 doesn't seem to provide any alternatives to libmagickwand-6.q16-2.  Maybe I'm just doing something wrong.


timholy:
Sounds good to me! CC @SimonDanisch.

PR #343, opened 2015-08-25
Add ColorVectorSpace to REQUIRE and fix OSX issues

timholy:


timholy:
This turned into quite an OSX festival---OSX wasn't properly tested with the Colors switch because failures occurred for "dumb" reasons (dependent packages also need to be merged/tagged) earlier in the build, so I dismissed those failures.

I'm going to merge because much of this is self-evident, but pinging @rsrock in case he sees any points of concern. e4c72980a2f7c03b0857d55293310b3b0b51cb5f is particularly worth taking a glance at, since we now have `GrayA` and `AGray` types.

rsrock:
Looks good to me.  All tests pass here, so it's good to the same extent that our tests are good :smile:.

Do we have an example of an `AGray` image?  I could never find one for testing, but I suppose I could make one myself.

I like the new color names, by the way.  Much less verbose.  I was getting tired of seeing things like  `AlphaColorValue{RGB{Ufixed16}, Ufixed16}`.

timholy:
> I like the new color names, by the way. Much less verbose. I was getting tired of seeing things like AlphaColorValue{RGB{Ufixed16}, Ufixed16}

Glad you like them. Previously one probably could have called that `RGBA{Ufixed16}`. But it was a `typealias` for the longer form, and typealiases continue to have some annoyances. https://github.com/JuliaLang/Color.jl/issues/101#issuecomment-132431179.

timholy:
No, I don't have an AGray image.

PR #360, opened 2015-09-16
Fix julia 0.4 type deprecations

timholy:


timholy:
Now with the addition of the `Union` deprecation, the logs are quite bad: https://travis-ci.org/timholy/Images.jl/jobs/80997593

Fixing that in a 0.3-compatible way would require an additional 55 `@compat`s. I think it's time to branch for 0.4 and leave julia-0.3 in the dust.

Awaiting comments.

SimonDanisch:
I'm all in =)

rsrock:
Sounds good to me.

I have a question, should we be excluding the next major release in our REQUIRES file?  Something like `julia 0.4 0.5`?  That does allow `julia 0.5-`, I think. It's tough to predict the future, but I imagine that our current code won't work with future versions of Julia, and it might be best to spell that out with each commit.

I ask because I've never seen that feature used in other packages.  It seems quite useful.

lucasb-eyer:
:+1:

timholy:
Sounds like a quorum to me.

@rsrock, it's an interesting idea. I haven't really thought about it. Since the package manager will always default to the most recent version, and I'm presuming we'll eventually have a version that _will_ run on 0.5, it seems like the old ones should never be chosen. Can you further clarify the advantage?

yuyichao:
> Fixing that in a 0.3-compatible way would require an additional 55 @compats. I think it's time to branch for 0.4 and leave julia-0.3 in the dust.

I've been going in the `add 55 @compat` route and it's not **that** bad, but I think branching isn't too bad either.

Issue #279, opened 2015-04-09
Creating an AbstractImage from raw png bytes


habemus-papadum:
Hi -- Is there a way to create an Image from png data that is stored in memory (without dumping the data to a temp file)?  Thanks!

timholy:
Yes, see https://github.com/timholy/Images.jl/blob/master/doc/function_reference.md#image-construction.

lilinjn:
My apologies, I'm still not getting it -- I've looked at the docs as well as the code...

To clarify, my in memory data is in PNG format in an Uint8 array and I am looking to convert it to a matrix of colors i.e. an Image.  

As far as I can tell, imread is the function I should be using, and there are methods that take filenames (which I'm hoping to avoid) and streams which would be great as I could use IOBuffer, except the stream version needs a filetype and there doesn't seem to be an ImageFileType for PNG (the only supported file types for reading via stream seem to be Imagine and SIF)

A self contained example of what I am hoping might be possible:

````
g = "digraph foo {1 -> 2;3 -> 2;3 -> 1}"
o,i,p = readandwrite(`dot -Tpng`)  ## use grpahviz to create a png
write(i,g)                                        ## write the input
close(i)
image = imread(o,??What Goes Here??)
````


For reference (but has nothing to do with Images)  the following works perfectly fine in IJulia

````
g = "digraph foo {1 -> 2;3 -> 2;3 -> 1}"
o,i,p = readandwrite(`dot -Tpng`)
write(i,g)
close(i)

d = readbytes(o)
display("image/png",d)
````

(This is quick and dirty example just for clarification)

Thanks (fyi, I used Images to create my github avatar!)

timholy:
Now I get it---sorry.

Probably the easiest would be `img = imread(o, Images.ImageMagick)`. If you're on OSX you could alternatively use the OSXnative reader.

BTW, nice avatar!

lilinjn:
Hi -- Unfortunately, that method does not exist:
`image = imread(o,Images.ImageMagick)`
````
`imread` has no method matching imread(::Pipe, ::Type{ImageMagick})
while loading In[57], in expression starting on line 5
````

And similarly for OSXNative.  

[...]

PR #365, opened 2015-09-23
Allow imread(::Array, Images.ImageMagick) to read from array of bytes

benmoran:
I came across issue #279 since I head the same problem so I implemented @timholy's suggested fix.  
The patch allows imread to accept an Array as well as String and IO - I'm not sure whether this is too general though, do you have any opinion on tightening it?


timholy:
Maybe insist that it be a `Vector{UInt8}`, just so it seems more like a buffer?

Otherwise, this looks great!

timholy:
Thanks!

PR #367, opened 2015-09-26
Add @test_approx_eq_sigma_eps for approximate image testing

timholy:
CC @tbreloff

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -94,6 +94,9 @@ export # types
     ScaleSigned,
     SliceData,
 
+    # macros
+    @test_approx_eq_sigma_eps,
+
     # constants
     palette_fire,
     palette_gray32,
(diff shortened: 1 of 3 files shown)


tbreloff:
Cool!

tbreloff:
Should I copy/paste this code into Plots for now?  I see the PR is still open, and I'd prefer not to have to track this branch to use it.  Thanks!

timholy:
Thanks for the reminder, I forgot all about it.

tbreloff:
Cool thanks! Now I have to actually write some tests...

tbreloff:
I'm just starting to put some test code together... do you have any guidelines as to how I might set sigma/eps?

timholy:
Nothing special. I'd personally set them to zero until I start seeing test failures, then study the specific failure. Once I understand the reason for the failure (small intensity change? slight change in antialiasing that leads to blurring?) then I'd have a principled choice.

That said, it sounds hard to go wrong blurring by a single pixel. The `eps` threshold is much harder to guess generically.

PR #370, opened 2015-09-26
Workaround upstream weak type inference

### LINKED ISSUE #356: Type-unstable `similar` and `copy`
For example, doing `code_warntype(copy, (Images.Image{Float64,2,Array{Float64,2}},))`, yields an abstract `::Images.Image{T,N,A<:AbstractArray{T,N}}` return type.

This also happens with the `similar` method, and seems to be due to a type-unstable `Image` constructor.

cdsousa:
Workaround https://github.com/JuliaLang/julia/issues/13082.
Trying to partially fix #356.

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -11,7 +11,7 @@ type Image{T,N,A<:AbstractArray} <: AbstractImageDirect{T,N}
     data::A
     properties::Dict{ASCIIString,Any}
 end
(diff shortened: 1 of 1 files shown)


timholy:
Ah, I hadn't read the second part carefully enough. Thanks!

timholy:
Oh, shoot, just noticed the AppVeyor failures (one is so common, I've started ignoring it, but this looks different).

timholy:
https://github.com/JuliaLang/julia/issues/13327.

If this isn't fixed quickly, we may need to revert.

PR #371, opened 2015-09-27
Merge master into sd/fileio

timholy:
CC @SimonDanisch Note this is against your branch; I thought I'd help with the merge conflict.


SimonDanisch:
Oh great! Yeah I totally messed up my rebase ;) It was my second rebase and I still had some misconceptions about what it really does.

timholy:
I noticed :smile: I just decided to merge rather than rebase because it was looking messy.

SimonDanisch:
Yeah cleaning that mess up was on the list for today...

timholy:
Hmm, this is fairly messed up too...it seems to have restored io.jl and test/io.jl. Will investigate.

SimonDanisch:
I think it'll take less time to make a new attempt fresh from master.. The only thing I did besides deleting files is the `grayim` stuff. If that works for you ;)

timholy:
I noticed the grayim stuff. There's one problem: several of the tests fail, of which the only really concerning one is [this test](https://github.com/timholy/Images.jl/blob/945bf2ec503e7c36a645be93b60725a278f6556d/test/algorithms.jl#L132). It would seem there's no way to define a `BitArray{Gray{Bool}}` since BitArrays don't have an element type.

I'm not quite sure what to do about that. A special exception in `grayim`, which preserves BitArrays intact?

timholy:
If you're going to rebase, I'll close this.

PR #372, opened 2015-09-27
second attempt with FileIO

SimonDanisch:
Didn't include the `grayim` change, because it introduced some conflict I wasn't sure how to fix (different values for `peakval` in edge->`Nonmax suppression vertical edge`.
But there seem to be a different `==` operator for Images, so that `Images{Gray{T}} == Images{T}` works, so I'm all good for FileIO.


SimonDanisch:
Okay besides that SIF doesn't have any tests and Imagine needs to be incorperated in some better way this is good to go, I think!
One other thing to do is to create some meta package for image IO tests. 
I'd like to have very thorough tests there, but its not really a blocking issue.

timholy:
See #373. The most important addition was deprecations for `imread` and `imwrite`.

timholy:
Superseded by #373.

PR #374, opened 2015-09-29
Add ImageMagick and (on OSX) QuartzImageIO to REQUIRE

timholy:


REQUIRE
```diff
@@ -6,3 +6,5 @@ SIUnits
 Zlib
 Graphics 0.1
 FileIO
+ImageMagick
```
rsrock: How about making this `@linux` and/or `@windows`?  The problem is that the `QuartzImageIO` tests require `Images`, `Images` requires this, which them means that Travis pulls down Homebrew and everything else that will never be used.  

Even worse is when ImageMagick is used, when `QuartzImageIO` fails for some reason. `ImageMagick` takes over and I never notice the problem!

timholy: We can do that.

timholy: BUT...OSX users will be left without a method for writing images.

timholy: Didn't you deliberately fall back to ImageMagick before, when OSXnative.jl failed?

rsrock: True, for the moment.  But we can always `Pkg.add` and `using ImageMagick` separately.

timholy: Let me make sure I understand:
> ImageMagick takes over and I never notice the problem!

For that one file, or from that point forward?

rsrock: Yep.  I think the way that it will work now is if QuartzImageIO fails, it will try ImageMagick if the user has added it.  If the user hasn't added it, they will get an error.

We can give errors that recommend a "Pkg.add("ImageMagick")", etc., if folks run into trouble.

timholy: Related reading:
- https://github.com/JuliaIO/FileIO.jl/pull/16. I'm still not sure about this, which is why I haven't merged it yet. An alternative is to define `load(format"ImageMagick", filename)` or something. But I'm beginning to wonder if we need one function (`load`) that _is_ overloaded by multiple packages, and a different one (`load_`) that is private to each module.
- https://github.com/JuliaIO/FileIO.jl/pull/40. Unlikely to be relevant, but just in case.

rsrock: (interleaved conversation here...)

ImageMagick takes over for that one file, I believe (haven't tested-- we don't emit info messages telling us which one is used).

Actually, I have a PR for FileIO that I'm just about to send over that will fix this particular issue.  Simon pointed me in the right direction there.

PR #380, opened 2015-10-04
move writemime to Images

SimonDanisch:
goes along with JuliaIO/ImageMagick.jl#10

### DIFF
--- a/src/writemime.jl
+++ b/src/writemime.jl
@@ -2,8 +2,8 @@
 ### writemime
 ###
 # only mime writeable to PNG if 2D (used by IJulia for example)
-mimewritable(::MIME"image/svg+xml", img::AbstractImage)             = false
-mimewritable(::MIME"image/png", img::AbstractImage) 				= sdims(img) == 2 && timedim(img) == 0
+mimewritable(::MIME"image/svg+xml", img::AbstractImage) = false
+mimewritable(::MIME"image/png", img::AbstractImage) = sdims(img) == 2 && timedim(img) == 0
 mimewritable{C<:Colorant}(::MIME"image/png", img::AbstractArray{C}) = sdims(img) == 2 && timedim(img) == 0
 
 
@@ -27,7 +27,7 @@ function writemime(io::IO, mime::MIME"image/png", img::AbstractImage; mapi=mapin
     if isdefined(:writemime_) && applicable(CurrentMod.writemime_, io, mime, imgcopy)
         return CurrentMod.writemime_(io, mime, imgcopy)
     else
-        error("No IO library loaded for writemime $mime with $(typeof(img)). 
+        error("No IO library loaded for writemime $mime with $(typeof(img)).
             Please consider putting \"using ImageMagick\" in your script"
         )
     end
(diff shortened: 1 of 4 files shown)


timholy:
I'm not a fan of this:
- It ties Images back to ImageMagick again, and will make it much harder to experiment with other loaders/savers.
- With FileIO, Images doesn't worry about I/O at all, and writemime is an I/O method
- Currently Images doesn't load ImageMagick but ImageMagick loads Images; consequently, ImageMagick can implement all of the necessary functionality by calling methods in Images.

If you think this is important, can you provide more justification?

SimonDanisch:
I'm not a big fan either :(
But we need to put the full signature in a package that is always loaded
before calling writemime, as it can't be loaded on demand.. So it can only
live in FileIO or Images. It shouldn't live in FileIO, because FileIO can't
fully qualify the signature without loading Images. So Images is the only
package left. I tried to tie it not too closely to ImageMagick and I can
further reduce it. We can call the `getblob/image2magickwand` `image2png`
and then let other libraries implement this function as well... We can also
try to be smarter with the loading of the library. Maybe with some logic
from FileIO?!

timholy:
How about having Images define
```jl
Base.writemime(io::IO, ::MIME"image/png", img::AbstractImage; mapi=nothing, minpixels=10^4, maxpixels=10^6) = error("Please load an I/O package via `using IOPkg`, where `IOPkg` would be ImageMagick or QuartzImageIO")
```

`ImageMagick.jl` and `QuartzImageIO.jl` can define a method with the identical signature and it will overwrite the one in Images.

lucasb-eyer:
Though you'd also want to add a hint that it should be loaded *after* `Images` in that message, right?

timholy:
It always will be, because ImageMagick looks like this:
```jl
module ImageMagick
using Images
# stuff
function Base.writemime(...)
   # blah
end

end  # module ImageMagick
```

lucasb-eyer:
Ah makes sense, nevermind me then.

SimonDanisch:
Does `QuartzImageIO` actually support this functionality? Haven't seen anything to write png's to streams, but haven't looked closely :D

rsrock:
Not yet, but I've looked into it and it should be easy to add.

(Once I finish writing this manuscript, that is...)

timholy:
I always write code before I write manuscripts. What's your problem? :wink: Oh, wait, maybe I'm the one with the problem...

timholy:
I guess the one concern is https://github.com/JuliaLang/julia/issues/265; people may have to restart their IJulia kernel when they hit this. Still, it's not like this is the only thing forcing that on users, and it's not 100% obvious this will be a problem---if the mime type is created in the same function that dispatches to `writemime`, then we should be saved by type-instability.

SimonDanisch:
Haha :D 
Okay... I'm not too happy, but I get the feeling that there won't come much happiness out of this issue eitherway...
I'd like to keep this part in Images, though, as it seems to be shared ultimately:
```Julia
   assert2d(img)
    A = data(img)
    nc = ncolorelem(img)
    npix = length(A)/nc
    while npix > maxpixels
        A = restrict(A, coords_spatial(img))
        npix = length(A)/nc
    end
    if npix < minpixels
        fac = ceil(Int, sqrt(minpixels/npix))
        r = ones(Int, ndims(img))
        r[coords_spatial(img)] = fac
        A = repeat(A, inner=r)
   end
```
Maybe we can keep this part of writemime in images and then do this at the end:
```Julia
if applicable(_writemime, io, mime, raw_data) 
_writemime(io, mime, raw_data) # _writemime, because otherwise we get into conflict with Array{Uint8,3}
else
error("Please load an I/O package via `using IOPkg`, where `IOPkg` would be ImageMagick or QuartzImageIO")
end
end

**Comment by SimonDanisch** [comment]
(or rather do the check at the beginning to avoid useless computations)

**Comment by rsrock** [comment]
> I always write code before I write manuscripts. What's your problem? :wink: Oh, wait, maybe I'm the one with the problem...

I'm with you on that.  Unfortunately, no Julia at all in this manuscript, so I can't even kid myself.

**Comment by timholy** [comment]
Definitely fine to keep any common parts in Images. The only part I object to is Images loading ImageMagick.

**Comment by SimonDanisch** [comment]
Good... Yeah, all this dynamic loading is terrible and best restricted to
one place

**Comment by SimonDanisch** [comment]
This should be good to merge!
Fixing REQUIRE maybe better in another PR.

File: src/writemime.jl [hunk 1]
```diff
@@ -0,0 +1,56 @@
+###
+### writemime
+###
+# only mime writeable to PNG if 2D (used by IJulia for example)
+mimewritable(::MIME"image/svg+xml", img::AbstractImage)             = false
+mimewritable(::MIME"image/png", img::AbstractImage) 				= sdims(img) == 2 && timedim(img) == 0
```
vchuravy: There are some whitespace issues here

SimonDanisch:
Thanks @vchuravy!
If no one protests, I'm going to merge this soon :)

Best,
Simon

timholy:
Sorry to be out of touch, I am traveling and short on time. By all means, go for it!

PR #383, opened 2015-10-12
don't install IO deps

SimonDanisch:


timholy:
Thanks! I'm in the process of dealing with all the deprecations in the color world from https://github.com/JeffBezanson/FixedPointNumbers.jl/pull/27 (which is a great PR, I'm not complaining!), and then I'll hope to tag a new version. Ping @rsrock re QuartzImageIO.

rsrock:
Will test all of this soon with an empty `.julia`.  Thanks!

PR #385, opened 2015-10-14
modified:   src/algorithms.jl:328 sumsq->abs2

hustf:
Removes Pkg.test deprecation warning.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -325,7 +325,7 @@ function imlog(sigma::Number=0.5)
 end
 
 # Sum of squared differences and sum of absolute differences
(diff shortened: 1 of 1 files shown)


kmsquire:
Personally, I think I'd rather this be renamed in Base to `sumsq`.

timholy:
I think it used to be `sumsq`, and got changed to `abs2`. In this case it's being used on a scalar, so the `sum` part of that name would indeed have been confusing.

timholy:
Thanks!

kmsquire:
(That's what happens when I don't actually read the PR.  Sorry for any confusion!)

PR #388, opened 2015-10-22
add exports for ufixedsc and ufixed8sc

rened:
The deprecation warning of `uint8sc` points to `ufixed8sc`, but that is not exported yet.


timholy:
Thanks!

PR #390, opened 2015-10-25
added test_approx_eq_sigma_eps function and difftype fix

tbreloff:
Based on your macro of the same name, I made a function version with some additional types, comments, and error messages, and I also added a `difftype` method that allows it to work in all the cases I tested so far.

I'm sure you had a good reason to make it a macro, but it seems to work ok as a function.  I'm happy to change that, or anything else you want.


src/algorithms.jl
```diff
@@ -383,6 +384,31 @@ macro test_approx_eq_sigma_eps(A, B, sigma, eps)
     end
 end
 
+# image difference testing (@tbreloff's, based on the macro)
+#   A/B: images/arrays to compare
+#   sigma: tuple of ints... how many pixels to blur
+#   eps: error allowance
+# returns: percentage difference on match, error otherwise
+function test_approx_eq_sigma_eps(A::AbstractMatrix, B::AbstractMatrix,
+                                  sigma::AbstractVector{Int} = [1,1],
```
timholy: Doesn't have to be `Int`. Also, you're making a strong assumption that the images are 2-dimensional. Better would be `ones(ndims(A))`.

src/algorithms.jl
```diff
@@ -383,6 +384,31 @@ macro test_approx_eq_sigma_eps(A, B, sigma, eps)
...
+                                  eps::@compat(AbstractFloat) = 1e-2)
+    if size(A) != size(B)
+        error("Arrays differ: size(A): $(size(A)) size(B): $(size(B))")
+    end
+    if length(sigma) != 2
```
timholy: `length(sigma) != ndims(A)`

timholy:
It's fine to have it as a function. The only differences are that (1) the macro allows one in principle to display the expressions that generated the failure (but if you always use them as `@test_approx_eq_sigma_eps A B`, that's not very informative), (2) consistency with `@test_approx_eq` in `Base.Test`, (3) the macro requires new code to be compiled every time, whereas the function won't be. The last is certainly an advantage for the function.

Needs tests before I can merge this, see e.g., #367.

tbreloff:
I added tests and generalized the signature.  Based on your comments, I think it makes sense to have both the function and macro versions, although we could consider having the macro just wrap the function.  Let me know if you want any other changes.

timholy:
Yeah, it probably would be best to have the macro wrap the function. You willing to tackle this, too?

At some later point I can add the expression-printing; I realized that my original version doesn't even take advantage of this.

src/algorithms.jl
```diff
@@ -344,6 +344,7 @@ difftype{T<:Integer}(::Type{T}) = Int
 difftype{T<:Real}(::Type{T}) = Float32
 difftype(::Type{Float64}) = Float64
 difftype{CV<:Colorant}(::Type{CV}) = difftype(CV, eltype(CV))
+difftype{CV<:Colorant, T<:Real}(::Type{CV}, ::Type{T}) = RGBA{T}
```
timholy: Hmm, it occurs to me that this could be problematic, given that now `UFixed` numbers are closed under arithmetic:
```
julia> UFixed8(0.1) - UFixed8(0.2)
UFixed8(0.906)
```
That's why I introduced the two-argument versions like those below.

timholy:
Bump. I was just testing Plots for you and got asked about a whole heap of "do you want to replace this image?" queries.

tbreloff:
Thanks for the bump.  How do you want to handle this:
>  this could be problematic, given that now UFixed numbers are closed under arithmetic

I'm not sure if you've looked at my test code in Plots, but this is the methodology (I'm open to improvements):

- For each example plot:
  - Run the commands and save a PNG to a temporary location
  - Call `test_approx_eq_sigma_eps` with this temporary image and a saved reference image
  - If they do **not** match:
    - if `isinteractive()`, pull up a Gtk window with the images side-by-side and a dialog asking if we should save the temporary image as the new reference
    - otherwise, rethrow the error

So when you run the tests yourself it goes through the GUI process and gives you a chance to visually verify that a new reference image is appropriate.  When run by travis, it simply errors.  This has worked fairly well so far... travis has caught some issues which normally I would have been searching for myself.

timholy:
Turns out most of the errors are because `Plots/test/imgcomp` uses this function, which has not yet been merged. If I add the `@` symbol in front of `test_approx_eq_sigma_eps`, then most gadfly tests pass. I'll file an issue with the exceptions for both pyplot and gadfly.

timholy:
My main point regarding

> this could be problematic, given that now UFixed numbers are closed under arithmetic

is that a better definition would probably be

```jl
difftype{CV<:RGBA,T<:Real}(::Type{CV}, ::Type{T}) = RGBA{Float32}
difftype{CV<:RGBA}(::Type{CV}, ::Type{Float64}) = RGBA{Float64}
```
to prevent the overflow.

timholy:
When you fix these, this will need rebasing, unfortunately.

tbreloff:
@timholy Understood.  Do you have a cheat sheet on proper rebasing commands?  I want to do it right the first time to avoid too much back and forth.

Also, I will include the difftype signatures for BGRA.

timholy:
http://docs.julialang.org/en/stable/manual/packages/#making-changes-to-an-existing-package

PR #393, opened 2015-11-02
Improve type stability of Image constructor

mbauman:
Inference is run before the inlining pass, so when inference runs on constructors of the form `T{ndims(A)}(...)`, it does't see the inlined constant; it just sees an Int.  The fix is to just use the type parameters directly.  I've added this as an additional constructor to make sure it doesn't break any existing code with very poorly typed arrays.

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -12,6 +12,7 @@ type Image{T,N,A<:AbstractArray} <: AbstractImageDirect{T,N}
     properties::Dict{ASCIIString,Any}
 end
 Image(data::AbstractArray, props::Dict) = Image{eltype(data),ndims(data),typeof(data)}(data,props)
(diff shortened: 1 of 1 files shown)


mbauman:
Aha, I just noticed https://github.com/timholy/Images.jl/pull/370, which was reverted due to Windows failures.  Looking at the AppVeyor logs there, it failed due to a missing method for the `Image` constructor.  Unlike #370, this keeps the old constructor around, so it will still work even in the absence of good typing.  AppVeyor already [passed once](https://ci.appveyor.com/project/timholy/images-jl/build/1.0.422), and I re-ran it just to make sure.

(Interesting that inference behaved differently on Windows in #370, though…)

timholy:
Hey, that's great! Interesting that neither the author of #370 nor myself thought of keeping the old one around as a backup. :+1:

PR #394, opened 2015-11-03
Testing fixes

timholy:
Fixes an occasional fail-to-fail issue (e.g., https://ci.appveyor.com/project/timholy/images-jl/build/1.0.424/job/kinjj8svl0dxf3gp, https://github.com/JuliaLang/julia/pull/13834#issuecomment-152822134). CC @tkelman. Also tests that #356 really is fixed now thanks to #393. CC @cdsousa, @mbauman.

### DIFF
--- a/src/map.jl
+++ b/src/map.jl
@@ -164,18 +164,18 @@ similar{T,F,To,From,S}(mapi::ScaleMinMax{To,From,S}, ::Type{T}, ::Type{F}) = Sca
 # Implementation
 function map{To<:Union{Real,AbstractGray},From<:Union{Real,AbstractGray}}(mapi::ScaleMinMax{To,From}, val::From)
     g = gray(val)
-    t = ifelse(g  < mapi.min, zero(From), ifelse(g  > mapi.max, mapi.max-mapi.min, g-mapi.min))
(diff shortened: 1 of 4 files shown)


timholy:
https://travis-ci.org/timholy/Images.jl/jobs/88983419 finally captured the infamous off-by-one bug. Remarkably, it seems due to a difference between `x/s` and `x*(1/s)`. Hopefully fixed in the 3rd commit in this PR. CC @SimonDanisch,

timholy:
@tkelman, since I'm looking at the tests carefully: I noticed that the 64-bit AppVeyor tests take much longer than the 32-bit, see https://ci.appveyor.com/project/timholy/images-jl/build/1.0.430. After the installation of julia itself, one step that seems to take a long time (3-5 minutes) is going from `INFO: Testing Images` to seeing any kind of output for the first test. On 32-bit it's not speedy either, but it's more like 1 minute. I'm guessing it's precompiling during that time, but the performance difference might be a source of concern?

tkelman:
Yes it might, but I don't think I'm equipped in either time or skill to get to the bottom of that one particularly soon.

PR #400, opened 2015-11-14
"Don't pad in imfilter_fft if size(kernel)==size(img)"

andferrari:
Here is a (perhaps too simplistic) « fix ».


timholy:
Looks good. Seems like we didn't have any tests for "circular", so I won't make you add one :smile:

timholy:
Thanks for the contribution!

timholy:
Tests added in #401

PR #398, opened 2015-11-12
Docstrings!

timholy:
```
julia> using Images

help?> separate
search: separate

  imgs = separate(img) separates the color channels of img, for example returning an m-by-n-by-3 array from an m-by-n array of RGB.
```

This is a WIP because there are still many files to go, but I wanted to post this here now because of one concern: presumably it would be best to generate the [function reference](http://timholy.github.io/Images.jl/function_reference.html) from the docstrings. I haven't plumbed into the guts of the new documentation system, so: does anyone know how to do this?

Particular CC to @rsrock, since he invested a lot in the github.io docs.

PR #403, opened 2015-11-15
[WIP] Lexicon-based API doc generation

rsrock:
I've started to look at a follow-up to #398 using Lexicon.jl. It does a good job generating the API docs.  The main change here is to include a `build.jl` script in the `doc` dir that runs Lexicon. 

Now we have to work out the last steps of generating the package, and there are two potential routes.

1) Stick with dcjones's Judo.jl.  It currently hangs because Judo tries to run any code blocks (fenced with triple backticks or indented by 4-spaces).  We would have to decorate all of these with `{.julia execute="false"}` (except any that we'd like to run as examples, of course).

Judo now has a branch that uses Julia's markdown processing to ditch the `pandoc` dependency.  That's nice.

2) Switch to `mkdocs`, which is recommended by Lexicon. One nice feature that it has is the ability to deploy docs to gh-pages with a single command.  But it is an additional dependency for package writers.

Any thoughts?


timholy:
Sweet!

I don't have strong opinions here, and am happy to be guided by your preferences (if you have them). I do have python installed, and I imagine I can figure out any problems in getting `mkdocs` working on my machine, if that's the direction you're leaning in.

Presumable this is something that won't be a concern for PRs, right? It would be just you, me, and a few others who might generate the docs from time to time?

rsrock:
Hmm, it looks like I accidentally updated the docs for this repo while working on this.  I was trying to test on my fork, but it looks like the `mkdocs gh-deploy` command pushes to `origin` instead of what's set in the `mkdocs.yml` configuration.

That said, I think the result looks pretty nice.  To get here, it's just a matter of `julia docs/build.jl; mkdocs gh-deploy --clean`.  I will package this up in just a bit, hopefully this evening.

rsrock:
See #431

PR #404, opened 2015-11-16
Fix tests for scalar index change on 0.5

yuyichao:
This is the bare minimum change to make the tests pass. I think there shouldn't be much (any?) others that needs updating given `Images` mostly delegate to the base version for indexing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
     - osx
 julia:
     - 0.4
-#    - nightly
+    - nightly
 notifications:
     email: false
 script:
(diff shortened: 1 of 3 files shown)


timholy:
Thanks, @yuyichao! Since this is on your own fork, can I ask you to also remove the comments from [this line](https://github.com/timholy/Images.jl/blob/4594c7363e9fba30e1befa5f8e0c5fec56cf746f/.travis.yml#L7) and [this one](https://github.com/timholy/Images.jl/blob/4594c7363e9fba30e1befa5f8e0c5fec56cf746f/appveyor.yml#L5-L6) so the tests get run on 0.5?

yuyichao:
Ahh, I didn't check that. Sure.

yuyichao:
Done and tests all passed.

P.S. I like this PR number, even though apparently no one can find it. ;-p.

timholy:
:smile:

Thanks. You're doing great stuff over a wide swath of julia, and we are all grateful.

Issue #402, opened 2015-11-15
Morphology-based Operations


pratapvardhan:
I was looking for mathematical morphology support in Images.jl and found base operations (dialtion, erosion, opening, closing).

Before I take a stab at additional morphology functions, wanted to check, if the community is open for these to be added to the library. I've just started using/looking into Images.jl

List of morphological functions:
- [x] [morphgradient](https://en.wikipedia.org/wiki/Morphological_gradient) -  Morphological gradient
- [x] [morphlaplace](http://www.mif.vu.lt/atpazinimas/dip/FIP/fip-Morpholo.html#Heading107) - morphologically-based Laplacian filter
- [x] [tophat](https://en.wikipedia.org/wiki/Top-hat_transform) - Top-hat filtering
- [x] [bothat](https://en.wikipedia.org/wiki/Top-hat_transform) - Bottom-hat filtering

timholy:
Yes, these would be great.

If you're looking at code in Images, you might notice that there's a lot of use of Base.Cartesian. You might still need it for some operations, but I have on my agenda to rewrite a lot of those algorithms using the CartesianIndex iterators in julia 0.4. Now many more arbitrary-dimensional algorithms can be implemented efficiently without resorting to metaprogramming.

That said, if your intention was to implement them just for 2d, that's also fine. My own work usually involves 3d and 4d (space+time) images, but I don't want to make that a barrier for others for whom that might be irrelevant.

juliohm:
I think this issue can be closed.

PR #405, opened 2015-11-16
ENH: Add morphological tophat, bothat, gradient, laplacian

pratapvardhan:
In continuation with discussion from #402 

1. Added morphological tophat, bothat, gradient, laplacian functions
2. Basic test cases
3. Docs

I'm quite not sure if I've made proper/any use of copy, array properties to save memory or other optimizations for these functions. I thought, I'll push this is up and take feedback for the modifications/improvements if any needed. And, the documentation for these is a bit rusty too.

### DIFF
--- a/doc/function_reference.md
+++ b/doc/function_reference.md
@@ -1048,6 +1048,42 @@ perform the `opening` and `closing` morphology operations. `opening` does first
 in the oposite way. The region parameter is passed to `erode` and `dilate` and describes
 the kernel size over which these operations are performed.
 
+## tophat
+```{.julia execute="false"}
+tophat(img, [region])
+```
+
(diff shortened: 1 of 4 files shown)


timholy:
Thanks! Looks fine to me. Eventually the function reference should be autogenerated (see #403), but for now it's very nice to have it added manually.

pratapvardhan:
Nice! I was thinking, if #402 should be kept open for some more time? Where I'll update the todo list of Morphology functions, tick-up/strike off the implemented functions or would you want me to close that and start a fresh issue?

timholy:
That would be fine; good idea to organize the effort.

Thanks again for the contributions!

PR #406, opened 2015-11-16
added test_approx_eq_sigma_eps function and difftype fix

tbreloff:
@timholy Please let me know if I rebased/squashed correctly, and if there's anything else to change.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -388,6 +388,10 @@ difftype{T<:Integer}(::Type{T}) = Int
 difftype{T<:Real}(::Type{T}) = Float32
 difftype(::Type{Float64}) = Float64
 difftype{CV<:Colorant}(::Type{CV}) = difftype(CV, eltype(CV))
+difftype{CV<:RGBA,T<:Real}(::Type{CV}, ::Type{T}) = RGBA{Float32}
+difftype{CV<:RGBA}(::Type{CV}, ::Type{Float64}) = RGBA{Float64}
+difftype{CV<:BGRA,T<:Real}(::Type{CV}, ::Type{T}) = BGRA{Float32}
+difftype{CV<:BGRA}(::Type{CV}, ::Type{Float64}) = BGRA{Float64}
 difftype{CV<:AbstractGray,T<:Real}(::Type{CV}, ::Type{T}) = Gray{Float32}
 difftype{CV<:AbstractGray}(::Type{CV}, ::Type{Float64}) = Gray{Float64}
 difftype{CV<:AbstractRGB,T<:Real}(::Type{CV}, ::Type{T}) = RGB{Float32}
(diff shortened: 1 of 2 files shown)


test/algorithms.jl
```diff
@@ -97,6 +97,16 @@ facts("Algorithms") do
         a = colorim(rand(3,15,15))
         @fact (Images.@test_approx_eq_sigma_eps a a [1,1] 0.01) --> nothing
         @fact_throws ErrorException (Images.@test_approx_eq_sigma_eps a colorim(rand(3,15,15)) [1,1] 0.01)
+
+        a = rand(5,5)
+        @fact_throws ErrorException Images.test_approx_eq_sigma_eps(a, rand(3,5), [1,1], 0.1)
+        @fact_throws ErrorException Images.test_approx_eq_sigma_eps(a, rand(5,5), [1,1], 0.1)
```
timholy: Looks like this line failed to throw an error. I think I've bumped the eps up to 0.01 in the other tests, that should probably fix this too.

timholy: Travis failure here: https://travis-ci.org/timholy/Images.jl/jobs/91455248

tbreloff: Hmm ok it passed for me.  Yeah 10% seems like a high threshold.  You can wait to merge if you want.

tbreloff:
I also had the thought that it might be a good addition to the algorithm to expand the sizes of the matrices to the largest of each dimension, so [these comparisons](https://github.com/tbreloff/Plots.jl/issues/70) might still pass the test if sufficiently similar otherwise.  Do you have thoughts/opinions on that?  I could try to include that functionality before merging.

test/algorithms.jl
timholy: Most did pass; it depends on the specific random numbers generated.

timholy: I will wait, just because I recently put in a certain amount of time to make sure the tests produced consistent results (#394). Inconsistent test results make it hard to evaluate contributions, unfortunately.

timholy:
> I also had the thought that it might be a good addition to the algorithm to expand the sizes of the matrices to the largest of each dimension, so these comparisons might still pass the test if sufficiently similar otherwise.

That's a clever idea. I have no idea, but it seems worth a try!

test/algorithms.jl
tbreloff: That make sense. The easiest solution is to call srand at the beginning so that the tests are deterministic. Unless you have objections I'll add that, along with the matrix-resizing addition.

timholy: That's fine.

tbreloff:
I'm going to reopen... comments welcome.  This version has another option to `expand_arrays`, which will expand the smaller of each dimension with `zero(eltype(A))` in those expanded positions.  I'll probably add one more commit with some additional tests for this functionality.

src/algorithms.jl
```diff
@@ -432,6 +436,42 @@ macro test_approx_eq_sigma_eps(A, B, sigma, eps)
     end
 end
 
+# image difference testing (@tbreloff's, based on the macro)
+#   A/B: images/arrays to compare
+#   sigma: tuple of ints... how many pixels to blur
+#   eps: error allowance
+# returns: percentage difference on match, error otherwise
```
timholy: Now there are docstrings in Images. But I can fix this myself, as I've asked you to revise this lots of times already.

timholy:
Thanks!

tbreloff:
Cool... Thanks!

PR #407, opened 2015-11-18
Add corner detection functionality

yhenon:
This adds corner detection based on the Harris algorithm and the Shi-Tomasi algorithm. These are pretty common image processing tasks, so I thought they may be useful to others.

### DIFF
--- a/src/corner.jl
+++ b/src/corner.jl
@@ -1,4 +1,4 @@
-function imcorner(img::AbstractArray, method::AbstractString="harris", border::AbstractString="replicate", blockSize::Int=3, k::Float64=0.04)
+function imcorner(img::AbstractArray; method::AbstractString="harris", border::AbstractString="replicate", blockSize::Int=3, k::Float64=0.04)
     # Performes corner detection using the Harris method or the Shi-Tomasi method
 
     (grad_x,grad_y) = imgradients(img, "sobel",border);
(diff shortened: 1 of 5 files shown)


src/corner.jl
```diff
@@ -0,0 +1,28 @@
+function imcorner(img::AbstractArray, method::AbstractString="harris", border::AbstractString="replicate", blockSize::Int=3, k::Float64=0.04)
```
timholy: Should any of these be keyword arguments?

yhenon: I think that's a good idea, everything but `img` could be a keyword.

timholy:
Need any help with this? Not sure why it was closed. The Travis failures?

yhenon:
Hi Tim,
I was just dealing with a Git issue on my side and was afraid of messing up. I've reopened it now. I've noticed I should add a docstring to the function imcorner() - any other comments ?

timholy:
I totally understand git issues. It was the biggest barrier for me getting started with contributing to julia.

Adding the docstring would be great. Otherwise, everything looks really good!

timholy:
FYI I don't get an email when you update this, so if you make changes and want me to look at it, just post a little comment.

If you click on the red "x" you'll see there's a test failure.

yhenon:
Thanks for all your assistance Tim! I've pushed an update that should fix the test failures.

timholy:
Thanks for the contribution!

yhenon:
Thanks for everything Tim. I hope to be making some more contributions in the fitire.

PR #413, opened 2015-11-22
Add extrema_filter

codles:
Added function and tests for 2d and 4d case

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1603,23 +1603,6 @@ function extrema_filter{T <: Number}(a::AbstractArray{T}, window::Int)
     return minval, maxval
 end
 
-
-function min_filter(a::AbstractArray, window::Int)
-
-    minval, maxval = minmax_filter(a, window)
-
-    return minval
-end
-
-
-function max_filter(a::AbstractArray, window::Int)
-
-    minval, maxval = minmax_filter(a, window)
-
-    return maxval
-end
-
-
 function wedgeisempty(X::Wedge)
     X.n <= 0
 end
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1476,8 +1476,6 @@ extr(order::Ordering, x::Color, y::Color, z::Color) = extr(order, convert(RGB, x
 
 # Min max filter
 
-type Wedge
-    buffer::AbstractArray
 # This is a port of the Lemire min max filter as implemented by Bruno Luong
 # http://arxiv.org/abs/cs.DS/0610046
 # http://lemire.me/
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1478,6 +1478,13 @@ extr(order::Ordering, x::Color, y::Color, z::Color) = extr(order, convert(RGB, x
 
 type Wedge
     buffer::AbstractArray
+# This is a port of the Lemire min max filter as implemented by Bruno Luong
+# http://arxiv.org/abs/cs.DS/0610046
+# http://lemire.me/
+# http://www.mathworks.com/matlabcentral/fileexchange/24705-min-max-filter
+
+type Wedge{A <: AbstractArray}
+    buffer::A
     size::Int
     n::Int
     first::Int
@@ -1512,8 +1519,8 @@ for N = 2:4
             end
 
             # Circular shift the dimensions
-            maxval_temp = permutedims(maxval_temp, mod([1:$N], $N)+1)
-            minval_temp = permutedims(minval_temp, mod([1:$N], $N)+1)
+            maxval_temp = permutedims(maxval_temp, mod(collect(1:$N), $N)+1)
+            minval_temp = permutedims(minval_temp, mod(collect(1:$N), $N)+1)
 
         end
 
@@ -1547,12 +1554,12 @@ function extrema_filter{T <: Number}(a::AbstractArray{T}, window::Int)
 
     for i = 2:n
         if i > window
-            if ~wedgeisempty(U)
+            if !wedgeisempty(U)
                 maxval[i-window] = a[getfirst(U)]
             else
                 maxval[i-window] = a[i-1]
             end
-            if ~wedgeisempty(L)
+            if !wedgeisempty(L)
                 minval[i-window] = a[getfirst(L)]
             else
                 minval[i-window] = a[i-1]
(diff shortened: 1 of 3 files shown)


codles:
Initial add following [METADATA](https://github.com/JuliaLang/METADATA.jl/pull/4036) discussion.

src/algorithms.jl
```diff
@@ -1473,6 +1473,171 @@ extr(order::Ordering, x::Color, y::Color) = extr(order, convert(RGB, x), convert
...
+                # Create index for full array (fa) length
+                @nexprs $(N)   j->(fa_{j} = 1:size(maxval_temp)[j])
+                @nexprs $(N-1) j->(fa_{j} = i_j)
+
+                # Create index for short array (sa) length
+                @nexprs $(N)   j->(sa_{j} = 1:size(maxval_temp)[j] - window[dim] + 1)
+                @nexprs $(N-1) j->(sa_{j} = i_j)
+
+                # Filter the last dimension
+                (@nref $N minval_temp sa) = min_filter(vec( @nref $N minval_temp fa), window[dim])
```
timholy: This doesn't seem to be defined (likewise `max_filter`)

src/algorithms.jl
```diff
@@ -1473,6 +1473,171 @@ extr(order::Ordering, x::Color, y::Color) = extr(order, convert(RGB, x), convert
 extr(order::Ordering, x::Color, y::Color, z::Color) = extr(order, convert(RGB, x), convert(RGB, y), convert(RGB, z))
 
 
+
+# Min max filter
+
+type Wedge
+    buffer::AbstractArray
```
timholy: For good performance, you're going to want to declare this to be a leaf type. If you need to handle arbitrary array types, then something like
```jl
type Wedge{A<:AbstractArray}
    buffer::A
    size::Int
    ...
end
```
will give you much better performance.

timholy: (See http://docs.julialang.org/en/stable/manual/faq/#how-do-abstract-or-ambiguous-fields-in-types-interact-with-the-compiler)

src/algorithms.jl
```diff
@@ -1473,6 +1473,171 @@ extr(order::Ordering, x::Color, y::Color) = extr(order, convert(RGB, x), convert
...
+
+    # Initialise the internal wedges
+    # U[1], L[1] are the location of the global maximum and minimum
+    # U[2], L[2] are the maximum and minimum over (U1, inf)
+    L = Wedge(zeros(Int,1,window+1), window+1, 0, 1, 0, 0)          # Min
+    U = Wedge(zeros(Int,1,window+1), window+1, 0, 1, 0, 0)
+
+    for i = 2:n
+        if i > window
+            if ~wedgeisempty(U)
```
timholy: ! rather than ~? (Matlab-ism?)

src/algorithms.jl
```diff
@@ -1473,6 +1473,171 @@ extr(order::Ordering, x::Color, y::Color) = extr(order, convert(RGB, x), convert
...
+
+    return minval, maxval
+end
+
+function wedgeisempty(X::Wedge)
+    X.n <= 0
+end
+
+function pushback(X::Wedge, v)
+    X.last = mod(X.last, X.size) + 1
```
timholy: AFAICT, this is just a circular buffer and loops 1,2,3,...,size,1,2,3,...size,.... It turns out that `mod` is slow (it relies on integer division, which is one of the slowest operations on a CPU), so you could probably speed this up with
```jl
X.last += 1
if X.last > X.size
    X.last = 1
end
```
or something.

src/algorithms.jl
```diff
@@ -1473,6 +1473,171 @@ extr(order::Ordering, x::Color, y::Color) = extr(order, convert(RGB, x), convert
...
+    X.buffer[X.last] = v
+    X.n = X.n+1
+    X.mxn = maximum([X.mxn, X.n])
```
timholy: `max(X.mxn, X.n)` will be orders-of-magnitude faster.

src/algorithms.jl
```diff
@@ -1473,6 +1473,171 @@ extr(order::Ordering, x::Color, y::Color) = extr(order, convert(RGB, x), convert
...
+    end
+
+    return minval, maxval
+end
+
+function wedgeisempty(X::Wedge)
+    X.n <= 0
+end
+
+function pushback(X::Wedge, v)
```
timholy: Maybe call this `pushback!`? Functions which mutate their inputs conventionally end in `!`.

codles:
Thanks @timholy. Ill make your suggested changes. And a few changes of my own.
Should I build on this pull request? Or make the changes and open a new pull request?

src/algorithms.jl
```diff
@@ -1473,6 +1473,171 @@ extr(order::Ordering, x::Color, y::Color) = extr(order, convert(RGB, x), convert
...
+
+                # Create index for full array (fa) length
+                @nexprs $(N)   j->(fa_{j} = 1:size(maxval_temp)[j])
+                @nexprs $(N-1) j->(fa_{j} = i_j)
+
+                # Create index for short array (sa) length
+                @nexprs $(N)   j->(sa_{j} = 1:size(maxval_temp)[j] - window[dim] + 1)
+                @nexprs $(N-1) j->(sa_{j} = i_j)
+
+                # Filter the last dimension
```
timholy: It seems that you're actually filtering on the first dimension? Julia arrays are column-major.

timholy:
You can just build on this one. When you push new commits to `codles:master`, this will be automatically updated.

I don't get an email each time you push a commit, so when you're ready for me to check this again just post a comment here.

src/algorithms.jl
codles: Thanks

src/algorithms.jl
codles: Thanks. Old habits die hard

src/algorithms.jl
codles: I didn't know this. Thanks

src/algorithms.jl
codles: Noted, thanks

src/algorithms.jl
codles: Again I was unaware of this. Thanks

codles:
@timholy I think I have addressed your concerns above and added tests for all lines of code. Please let me know of any additional changes you want made.

timholy:
Yay, thanks!

PR #416, opened 2015-11-24
Fix #415: Add entropy for grayscale (intensity) images

### LINKED ISSUE #415: Entropy for grayscale (intensity) images
Hi Tim,

Do you consider the entropy useful to have in `Images.jl`? http://www.mathworks.com/help/images/ref/entropy.html

What do you think of this implementation? May I send a PR?

```julia
function entropy(img::AbstractArray)
  boolean = isa(img, AbstractArray{Bool})
  if boolean # quick count
    p = sum(img) / length(img)
    (0 < p < 1) ? - p*log2(p) - (1-p)*log2(1-p) : 0
  else # check histogram
    _, counts = hist(img[:], 256)
    p = counts / length(img)
  -sum((p.!=0).*p.*log2(p))
end

entropy(img::AbstractImageDirect) = entropy(data(img))
```

juliohm:


src/algorithms.jl
```diff
@@ -128,6 +128,45 @@ meanfinite(img::AbstractImageIndexed, region) = meanfinite(convert(Image, img),
...
+
+function entropy(img::AbstractArray{Bool}; kind=:shannon)
+  logᵦ = _log(kind)
+
+  p = sum(img) / length(img)
+
+  (0 < p < 1) ? - p*log2(p) - (1-p)*log2(1-p) : zero(p)
+end
+
+entropy{C<:AbstractGray}(img::AbstractArray{C}; kind=:shannon) = entropy(raw(img), kind)
```
timholy: This should be `kind=kind`.

test/algorithms.jl
```diff
@@ -39,6 +39,10 @@ facts("Algorithms") do
         imgm = mean(img)
         imgn = img/imgm
         @fact reinterpret(Float64, Images.data(imgn)) --> roughly(convert(Array{Float64}, A/mean(A)))
+        img = rand(1:10,10,10)
+        img2 = rand(1:2,10,10)
+        @fact all([entropy(img, kind=kind) for kind in [:shannon,:nat,:hartley]] .≥ 0) --> true
+        @fact all([entropy(img2, kind=kind) for kind in [:shannon,:nat,:hartley]] .≥ 0) --> true
```
timholy: Maybe add a test for an `Image{Gray}`, since that's what gets returned by `load`? You can test this with
```jl
img = reinterpret(Gray{U8}, grayim(rand(UInt8, 3,3)))
```

src/algorithms.jl
juliohm: Thanks, I missed that one.

test/algorithms.jl
juliohm: Will do.

juliohm:
Let me know if I can improve the PR further, I'm leaving my PC for a while, will check it later.

timholy:
It's great, thanks a lot!

juliohm:
I just saw a typo in the specialization for Boolean arrays, it should use logb instead of log2, will fix it in a quick patch through GitHub's interface...

PR #418, opened 2015-11-24
Fix typo and replace log2 by logb in entropy

juliohm:


missperovaz:
this is a really good add, thanks

carolina
     __o
  _`\<,_
(+) / (+)

www.carolinabrum.com
(avast users: turn it off pls)

PR #424, opened 2015-12-11
Fix #420: Add modified Hausdorff distance

### LINKED ISSUE #420: Hausdorff distance
Would you be interested in having the Hausdorff distance in `Images.jl`? http://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=7053955

It is a famous distance used for object-matching. We can apply it to binary images where active pixels are the points in the set.

I also would like to ask you if you know any other distance that can be applied to continuous images? What is the state of the art in computing similarity between images in computer graphics? Any well-known library (C, C++, Fortran, ...) for that?

juliohm:


src/distances.jl
```diff
@@ -0,0 +1,38 @@
...
+"""
+`MHD(imgA, imgB)` is the modified Hausdorff distance between binary
+images (or point sets).
+
+References
+----------
+Dubuisson, M-P; Jain, A. K., 1994. A Modified Hausdorff Distance for
+Object-Matching.
+"""
+function MHD(imgA::AbstractArray, imgB::AbstractArray)
```
timholy: In general, types should be uppercase and functions lowercase. I also suspect that `MHD` is not very discoverable. Could this be called `haussdorff_distance` or something similar?

juliohm: Sure Tim, I'll rename it.

-Júlio

timholy:
Very nice, thanks a lot!

PR #432, opened 2016-01-28
Fixed deprecation warning

hofmannmartin:
A very tiny change to fix the following warning...

```WARNING: [a] concatenation is deprecated; use collect(a) instead
 in depwarn at deprecated.jl:73
 in oldstyle_vcat_warning at ./abstractarray.jl:29
 in ctranspose at /home/hofmann/.julia/v0.4/Images/src/core.jl:1300```


timholy:
Thanks!

PR #436, opened 2016-02-07
Fix properties when using sliceim with Colons

timholy:
This also fixes a potential overwriting of original data with `imfilter_gaussian`.


timholy:
Failures on nightly are unrelated, so I'll merge.

PR #438, opened 2016-02-17
added component properties

bjarthur:
convenience functions to compute the bounding boxes, areas, and pixels in labelled components.  should work with N-D data.


timholy:
These implementations are `O(N*M)`, where `N` is the number of voxels and `M` the number of components. I think it's possible to do all this in `O(N+M)` without too much more effort. Would you be willing to give this a shot?

bjarthur:
oi.  good call.  this push is definitely faster.  is there more that could be done?

timholy:
:+1:, nice work!

timholy:
Curious that Travis didn't run on this, but since AppVeyor passed I'm merging.

timholy:
Ah, I just realized one question: is it worth documenting that the first element of the output arrays corresponds to label 0?

bjarthur:
that's easy enough.  or, should label 0 stats just not be included?  i could go either way.

timholy:
In my thinking, 0 is "background" and therefore seems on balance like it should be omitted. But like you, I could go either way. If it's hard to decide, one could always use a keyword to give either one at user discretion.

bjarthur:
it's easy enough for users to [2:end] the return value if they don't want the background label, so i just went ahead and updated the docs to make sure they knew to expect it to be included

timholy:
Great, works for me.

PR #439, opened 2016-02-18
typo fix in sample code

siddhantjain:
Just a small fix. Can the author please review and see if this is what they intended?


timholy:
Indeed I think so. @kmsquire?

kmsquire:
Yep

timholy:
Thanks!

PR #441, opened 2016-02-20
Implementing Canny Edge Detection

siddhantjain:
Can someone please review it? I felt the following things might need a review:

1.) Naming conventions for the variables
2.) Use of label_components for hystersis
3.) parameters for the function (should I add any other variables?)

### DIFF
--- a/src/edge.jl
+++ b/src/edge.jl
@@ -287,7 +287,7 @@ g = rgb2gray(rgb_image)
 gx, gy = imgradients(g)
 mag, grad_angle = magnitude_phase(gx,gy)
 mag[mag .< 0.5] = 0.0  # Threshold magnitude image
-thinned, subpix =  thin_edges_subpix(mag, grad_angle)
+thinned, subpix =  thin_edges_subpix(mag, gradient)
 ```
 """
 thin_edges{T}(img::AbstractArray{T,2}, gradientangles::AbstractArray, border::AbstractString="replicate") =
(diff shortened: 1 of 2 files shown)


src/edge.jl
```diff
@@ -287,7 +287,7 @@ g = rgb2gray(rgb_image)
 gx, gy = imgradients(g)
 mag, grad_angle = magnitude_phase(gx,gy)
 mag[mag .< 0.5] = 0.0  # Threshold magnitude image
-thinned, subpix =  thin_edges_subpix(mag, grad_angle)
+thinned, subpix =  thin_edges_subpix(mag, gradient)
```
kmsquire: Curious why you changed the name here, but not two lines up?

siddhantjain: I am sorry. I didn't review my changelist properly. This was fixed with [this commit](https://github.com/timholy/Images.jl/commit/c9b633caaeee604f49b46158d186ebfc9a2b2ab0), but my local version didn't reflect the change. This change should not be there.

timholy:
@siddhantjain, if you need help on how to work with git, [see here](http://docs.julialang.org/en/stable/manual/packages/#making-changes-to-an-existing-package) or just ask about specific sticking points.

yhenon:
I believe this does not properly implement canny edge detection. You have this:

	#double thresholding
	img_nonMaxSup[img_nonMaxSup .<lowerThreshold] = 0.0
	img_nonMaxSup[img_nonMaxSup .>upperThreshold] = 1.0
	#hysteris
	labels = label_components(img_nonMaxSup,trues(3,3)) 

In your implementation, component lying between the min and max threshold become part of the final image no matter what. It's actually necessary to check if those components are connected to a value above the max threshold. See [here](http://docs.opencv.org/trunk/da/d22/tutorial_py_canny.html#gsc.tab=0) for example.

mronian:
@siddhantjain Are you still working on this?

siddhantjain:
@mronian Go ahead and finish it off if you are planning to work on it right now.

timholy:
Closed by #495.

PR #443, opened 2016-02-23
fixing JuliaStats/DataArrays.jl#168

SimonDanisch:
Behold:
```Julia
julia> using Images, Gadfly
# <----- SO MUCH EMPTY SPACE
```
Now that I'm part of the `DataArrays`/`Images` club due to: https://github.com/jverzani/Mustache.jl/issues/26 and the gadfly backend in https://github.com/JuliaGL/GLVisualize.jl/pull/43, I decided it's finally time to tackle JuliaStats/DataArrays.jl#168 ....
I tried the solution that @timholy sketched out but run into destroying precompilation and creating circular references. If I just didn't do it right, it should now be way easier for someone with the right import/eval magic skills to implement a nicer solution ;)

I created a solution that utilizes a macro to stay as close as possible to the solution printed by the warning and to make it easier to change anything without the need of redoing the regex foo.

Best,
Simon


src/Images.jl
```diff
@@ -32,6 +32,13 @@ const is_little_endian = ENDIAN_BOM == 0x04030201
 immutable TypeConst{N} end  # for passing compile-time constants to functions
 
 include("core.jl")
+
+if VERSION > v"0.4.4"
```
timholy: What's the difference between 0.4.4 and prior 0.4 versions?

timholy:
Wow, this is amazing! Sorry my not-paying-attention caused me to completely space on this important contribution.

I'm not entirely sanguine on requiring DataArrays unless it benefits Images directly, so I may mull this over for a bit. I'm a bit swamped with other things though. How eager are you to have this merged?

vchuravy:
Is there anyway of doing a conditional import based on the fact that `DataFrames` is installed?

timholy:
Demo, with notation: IC = ImagesCore, DAC = DataArraysCore, IDA = ImagesDataArrays, I = Images, DA = DataArrays. Tested using `JULIA_PKGDIR=/tmp/fakepackages julia` and creating all of these as fake packages (so it's a pretty realistic demo).

First, the usage:
```jl
julia> using I

julia> using DA

julia> a = Image(rand(3,2));

julia> b = DataArray(rand(3,2));

julia> c = rand(3,2);

julia> a+c;

julia> b+c;

julia> a+b;
ERROR: not defined
 in + at /tmp/fakepackages/v0.4/IDA/src/IDA.jl:7
```

Definitions are in [this gist](https://gist.github.com/timholy/e81825a29d0124d098307af853ecc5a6).

timholy:
I should have also explained why this works:
- With precompilation, any "logic" in your source code runs at precompilation time. Consequently, any conditionals apply only to the state of your system when you built the package, not to when you're using it.
- The `__init__` function runs at the end of module initialization. It's the one place you can put runtime-logic.
- You need the types to be defined before you can define error stubs
- Once the ambiguity-generating methods get "loaded," it's too late to generate the stubs.

So as a consequence, the solution is: split the type-definitions into one module, and the ambiguity-generating methods in another. At the end of the type-definition module loading, check to see if any conflicting packages are loaded; if so, that means the types are defined, so you can generate the stubs.

You don't have to do this in different repos, you can do all of this in one repo with two (or more) sub-modules.

See also https://github.com/JuliaLang/julia/issues/11098.

src/Images.jl
SimonDanisch: It's a different set of ambiguity warnings, where the non intersecting parts create warnings in the the other version.
Might have changed by now!

timholy: Wow, didn't know that. I'll have to upgrade my julia-0.4 and check it out.

Also, FYI I'm thinking of trying to fix the Images tests on 0.5 over the next few days. It could possibly affect things like warnings.

timholy:
Superseded by https://github.com/JuliaLang/julia/pull/16125

PR #445, opened 2016-02-27
Rework writemime to use save(stream, img)

timholy:
This reworks `writemime` to address https://github.com/timholy/TestImages.jl/issues/11, leveraging FileIO's `save(stream, img)` to avoid the need for a `writemime_` method exported by FileIO. Requires https://github.com/JuliaIO/ImageMagick.jl/pull/29. (Come to think about it, this PR will fail Travis until we tag a new version of ImageMagick. I'll do that once https://github.com/JuliaIO/ImageMagick.jl/pull/30 merges.) CC @SimonDanisch, @tlnagy, @Evizero.

See also https://github.com/JuliaIO/QuartzImageIO.jl/pull/16

PR #448, opened 2016-02-28
Rework writemime, take 2

timholy:
A redo of #445, as GitHub closed the issue due to a commit to Colors that referenced the PR in Images.


timholy:
Passes for all julia-0.4 platforms. 0.5 has some issues that need fixing there.

PR #454, opened 2016-03-16
Minor doc fix for `imstretch`.

lucasb-eyer:
This makes the documentation correspond to the code: `m` is taken to the power too.

Since I wasn't sure whether the comment or the code was right, I tried to find a reference for this formula, but my google-fu fails me. Is there a reference? If you link it, I could add it to this PR before merge.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1342,7 +1342,7 @@ end
 `imgs = imstretch(img, m, slope)` enhances or reduces (for
 slope > 1 or < 1, respectively) the contrast near saturation (0 and 1). This is
 essentially a symmetric gamma-correction. For a pixel of brightness `p`, the new
-intensity is `1/(1+m/(p+eps)^slope)`.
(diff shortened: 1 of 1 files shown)


timholy:
Thanks!

As for a reference, I don't have one. This function pre-dated my own involvement with julia (there was an `images.jl` file in a directory called `examples/` in the main julia repository).

lucasb-eyer:
Wow git (and your memory) truly is awesome. It was easy to find that this was added by user @stefan-k in JuliaLang/julia@07117df16f17396865093a76a36522c91a5ce8c9. Stefan, any chance you remember a source for this formula?

stefan-k:
Hellooo again!
I don't remember the source for this formula, but I guess it is from the "Image Processing in Matlab" (or something like that) book. The likelihood that I made a mistake is pretty high. I don't recall ever using it.
Hope this helps! :)

Issue #453, opened 2016-03-14
GSoC Proposal


mronian:
I am interested in applying to Julia with a proposal for Images.jl

The ideas page on julialang.org has a [project proposal](http://julialang.org/soc/ideas-page.html#writing-high-performance-multithreaded-kernels-for-image-processing) for writing high performance multithreaded kernels for image processing, more specifically multithreaded versions of `imfilter` and `imfilter_gaussian`. 

I am also interested in adding functionality to Images.jl like Image Stitching, Face Detection, Object Tracking, Feature Descriptors or Drawing Functions. 

Which project would be higher in the priority list?

timholy:
Thanks for your interest @mronian. You raise some good questions. Overall I tend to think that some features, like face detection, are sufficiently specialized that they would like best be in a separate repository. https://github.com/JuliaImages is not yet populated, but perhaps that should change.

mronian:
Is https://github.com/JuliaImages planned to be a umbrella organisation for all image processing repositories?

Am I correct in interpreting JuliaImages as equivalent to OpenCV while Images.jl is equivalent to the imgproc component of OpenCV?

Maybe I could work on populating JuliaImages with the various functions taking a cue from other popular image processing libraries.

lucasb-eyer:
Thanks @timholy for this policy, I fully agree. I've used OpenCV a lot and I don't think it's an example to follow here.

mronian:
I would like to work on 2D Feature extraction for Images.jl, more specifically the following algorithms : 

- Canny
- Corner Harris
- Shi-Tomasi Corner Detector
- SIFT Descriptor
- SURF Descriptor
- ORB Descriptor
- GLCM Matrix
- Descriptor Matching Function

@timholy Does this seem like a good project to work on?

timholy:
Yes, absolutely! But some of this may already be available. See `src/corner.jl`, for example. Also see #441.

bjarthur:
@mronian  you should coordinate with [ImageRegistration.jl](https://github.com/seung-lab/ImageRegistration.jl).  they specifically mention SURF/SIFT.

mronian:
[...]

PR #455, opened 2016-03-18
added blob_LoG

bjarthur:
might be a better place to put feature extraction algorithms in future as discussed in https://github.com/timholy/Images.jl/issues/453 and https://github.com/seung-lab/ImageRegistration.jl/issues/2, but thought i'd submit a PR for this blob detector i recently made before @mronian gets to work

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1029,6 +1029,65 @@ function copytail!(dest, A, coloffset, strd, len)
     dest
 end
 
+"""
+`blob_LoG(img, sigmas) -> Vector{Tuple}`
+
+Find "blobs" in an N-D image using Lapacian of Gaussians at the specifed
+sigmas.  Returned are the local maxima's heights, radii, and spatial coordinates.
+
+See Lindeberg T (1998), "Feature Detection with Automatic Scale Selection",
+International Journal of Computer Vision, 30(2), 79–116.
+
+Note that only 2-D images are currently supported due to a limitation of `imfilter_LoG`.
+"""
+function blob_LoG{T,N}(img::AbstractArray{T,N}, sigmas)
+    img_LoG = Array(Float64, length(sigmas), size(img)...)
+    for isigma in eachindex(sigmas)
+        img_LoG[isigma,:] = sigmas[isigma] * imfilter_LoG(img, sigmas[isigma])
+    end
+
+    radii = sqrt(2)*sigmas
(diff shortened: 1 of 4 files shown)


src/algorithms.jl
```diff
@@ -1029,6 +1029,43 @@ function copytail!(dest, A, coloffset, strd, len)
...
+        @nloops $(N+1) i d->(2:size(img_LoG,d)-1) begin
+            ismax = true
+            img_LoG_I = (@nref $(N+1) img_LoG i)
+            @nloops $(N+1) j d->(i_{d}-1:i_{d}+1) begin
+                if (@nref $(N+1) img_LoG j) > img_LoG_I
+                    ismax = false
+                    break
+                end
+            end
+            ismax && push!(blobs,
```
timholy: Is this part just a 3d 3pt maximum-finding algorithm? We do have a maximum-filter already, but maybe not a maximum-finder (I can't remember).

 If we don't have something like this already, it seems worth splitting this out into a separate function (more composable design that way, i.e., other algorithms could use it). It could return integer indexes, which you could then convert to `(radius, ix, iy)`.

timholy:
Very nice. This can definitely live here until ImageFeatures is ready.

src/algorithms.jl
bjarthur: yes, it's a N-D maximum-finding algorithm, so 27 points in 3-D.  is the maximum filter you refer to `dilate!`.  not sure i can leverage that.  but happy to split out the maximum finder currently inside of `blob_LoG`.

timholy: Yeah, basically "if there isn't one in there already" (which I agree there probably isn't) then it would be lovely to split this one out. I can imagine you might have many customers using it for other things.

bjarthur:
i am undecided whether `regions` and `edges` of `findlocalmaxima` should be optional or keyword arguments.

bjarthur:
also...  might `findlocalmaxima` be general enough to include in `Base`?

timholy:
Wow, really nice!

timholy:
I don't think it needs to be included in base (base is going to shrink---it might even lose, e.g., sparse matrix algebra), but there's always a question of which of Image's algorithms should be packaged in some other way. I don't have any particular suggestions, so laziness suggests leaving it here, but if you have other preferences I'm comfortable with whatever you choose. (It's fine to make Images depend on other packages, for example.)

PR #456, opened 2016-03-18
Add flipdim with dimension as string parameter

Evizero:
Allows `Base.flipdim` to be called with a string parameter denoting the dimension. Tests are included

For example, the following code flips the x dimension, independent of the storage order:

```julia
img = grayimg(UInt8[200 150; 50 1])
img2 = flipdim(img, "x")
```

Alternatively, additional shortcut functions are provided for the three most common spatial dimensions x, y, and z. These shortcut functions are called `flipx`, `flipy`, and `flipz` respectively.

For example, the following code is equivalent to the example provided above:

```julia
img = grayimg(UInt8[200 150; 50 1])
img2 = flipx(img)
```

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -174,6 +174,11 @@ export # types
     ufixed8sc,
     ufixedsc,
 
+    # flip dimensions
+    flipx,
+    flipy,
+    flipz,
+
     # algorithms
     ando3,
     ando4,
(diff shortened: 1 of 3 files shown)


timholy:
Sorry I overlooked this for so long. Thanks!

PR #458, opened 2016-03-22
Histogram Equalisation

mronian:
This commit adds a histogram function to help with histogram equalisation.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1239,6 +1239,77 @@ end
 
 imhist{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins, minval, maxval) = imhist(img, nbins, convert(T, minval), convert(T, maxval))
 
+"""
+```
+hist_equalised_img = histeq(img, nbins)
+hist_equalised_img = histeq(img, nbins, minval, maxval)
+```
+
+Returns a grayscale histogram equalised image with a granularity of `nbins` number of bins. If minval and maxval are specified then
+only the intensities in the range (minval, maxval) are equalised.
+
+"""
+
+RET_TYPE = Dict("8bit" => Gray{U8}, "16bit" => Gray{FixedPointNumbers.UFixed{UInt16,16}})
+Y_MIN = 16
+Y_MAX = 235
+
+function _prep_image_for_histeq(img::AbstractArray, dtype)
+    img_shape = size(img)
+    if dtype == "8bit"
+        img = [convert(base_colorant_type(c){FixedPointNumbers.UFixed{UInt8,8}}, c) for c in img]
+    elseif dtype == "16bit"
+        img = [convert(base_colorant_type(c){FixedPointNumbers.UFixed{UInt16,16}}, c) for c in img]
+    end
+    img = reshape(img, img_shape)
+    img
+end
+
+histeq{T<:Colorant}(img::AbstractArray{T}, nbins, dtype = "8bit") = histeq(_prep_image_for_histeq(img, dtype), nbins, zero(YCbCr), zero(YCbCr))
+histeq{T<:Colorant}(img::AbstractArray{T}, nbins, minval, maxval, dtype = "8bit") = histeq(_prep_image_for_histeq(img, dtype), nbins, minval, maxval)
+
+function histeq{T<:Colorant}(img::AbstractArray{T}, nbins, minval, maxval)
+    img_ycbcr = convert(Array{YCbCr}, data(img))
+    colons = ntuple(d->Colon(), ndims(img))
+    Y = separate(img_ycbcr)[colons..., 1]
+    if maxval == zero(YCbCr{Float32})
+        eq_Y = histeq(Y, nbins, Y_MIN, Y_MAX)
+    else
+        eq_Y = histeq(Y, nbins, minval, maxval)
+    end
+    hist_equalised_img = [YCbCr(eq_Y[i], col.cb, col.cr) for (i,col) in enumerate(img_ycbcr)]
+    hist_equalised_img = reshape(hist_equalised_img, size(img_ycbcr))
+    hist_equalised_img = convert(Array{T}, hist_equalised_img)
+end
+
+function histeq(img::AbstractImage, nbins, dtype = "8bit")
+    hist_equalised_img = histeq(data(img), nbins, dtype)
+    hist_equalised_img = shareproperties(img, hist_equalised_img)
+    hist_equalised_img
+end
+
+function histeq(img::AbstractImage, nbins, minval, maxval, dtype = "8bit")
+    hist_equalised_img = histeq(data(img), nbins, minval, maxval, dtype)
+    hist_equalised_img = shareproperties(img, hist_equalised_img)
+    hist_equalised_img
+end
+
+histeq{T<:Gray}(img::AbstractArray{T}, nbins, dtype = "8bit") = histeq(_prep_image_for_histeq(img, dtype), nbins, ColorVectorSpace.typemin(RET_TYPE[dtype]), ColorVectorSpace.typemax(RET_TYPE[dtype]))
+
+function histeq{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins::Int, minval::T, maxval::T)
+    bins, histogram = imhist(img, nbins, minval, maxval)
+    cdf = cumsum(histogram[2:end-1])
+    img_shape = size(img)
+    cdf_length = size(cdf)[1]
+    hist_equalised_img = [max(1,Int(ceil((x-minval)*cdf_length/(maxval-minval)))) for x in img]
+    hist_equalised_img = [minval + ((cdf[x]-cdf[1])*(maxval-minval)/(cdf[end]-cdf[1])) for x in hist_equalised_img]
+    hist_equalised_img = reshape(hist_equalised_img, img_shape)
+    hist_equalised_img
+end
+
+histeq{T<:Number}(img::AbstractArray{T}, nbins) = histeq(img, nbins, minfinite(img), maxfinite(img))
+
+histeq{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins, minval, maxval) = histeq(img, Int(nbins), convert(T, minval), convert(T, maxval))
 
 """
 `imgr = restrict(img[, region])` performs two-fold reduction in size
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1280,18 +1280,27 @@ end
 histeq{T<:Colorant}(img::AbstractArray{T}, nbins, dtype = "8bit") = histeq(_prep_image_for_histeq(img, dtype), nbins, zero(YCbCr), zero(YCbCr))
 histeq{T<:Colorant}(img::AbstractArray{T}, nbins, minval, maxval, dtype = "8bit") = histeq(_prep_image_for_histeq(img, dtype), nbins, minval, maxval)
 
+function recompose_y(c::Color, eq_Y::Float32)
+    c_ycbcr = convert(YCbCr, color(c))
+    ret_c = YCbCr(eq_Y, c_ycbcr.cb, c_ycbcr.cr)
+    ret_c
+end
+
+function recompose_y(c::TransparentColor, eq_Y::Float32)
+    c_ycbcr = convert(YCbCrA, color(c))
+    ret_c = YCbCrA(eq_Y, c_ycbcr.cb, c_ycbcr.cr, c_ycbcr.alpha)
+    ret_c
+end
+
 function histeq{T<:Colorant}(img::AbstractArray{T}, nbins, minval, maxval)
-    img_ycbcr = convert(Array{YCbCr}, data(img))
-    colons = ntuple(d->Colon(), ndims(img))
-    Y = separate(img_ycbcr)[colons..., 1]
+    Y = [convert(YCbCr, color(c)).y for c in img]
     if maxval == zero(YCbCr{Float32})
         eq_Y = histeq(Y, nbins, Y_MIN, Y_MAX)
     else
         eq_Y = histeq(Y, nbins, minval, maxval)
     end
-    hist_equalised_img = [YCbCr(eq_Y[i], col.cb, col.cr) for (i,col) in enumerate(img_ycbcr)]
-    hist_equalised_img = reshape(hist_equalised_img, size(img_ycbcr))
-    hist_equalised_img = convert(Array{T}, hist_equalised_img)
+    hist_equalised_img = reshape([convert(T, recompose_y(c, eq_Y[i])) for (i, c) in enumerate(img)], size(img))
+    hist_equalised_img
 end
 
 function histeq(img::AbstractImage, nbins, dtype = "8bit")
@@ -1304,6 +1313,13 @@ function histeq(img::AbstractImage, nbins, minval, maxval, dtype = "8bit")
     hist_equalised_img = histeq(data(img), nbins, minval, maxval, dtype)
     hist_equalised_img = shareproperties(img, hist_equalised_img)
     hist_equalised_img
+end
+
+function histeq{T<:TransparentGray}(img::AbstractArray{T}, args...)
+    opaque_img = [color(c) for c in img]
+    hist_equalised_img = histeq(opaque_img, args...)
+    hist_equalised_img = reshape([convert(T, AGray(hist_equalised_img[i], alpha(c))) for (i, c) in enumerate(img)], size(img))
+    hist_equalised_img    
 end
 
 histeq{T<:Gray}(img::AbstractArray{T}, nbins, dtype = "8bit") = histeq(_prep_image_for_histeq(img, dtype), nbins, ColorVectorSpace.typemin(RET_TYPE[dtype]), ColorVectorSpace.typemax(RET_TYPE[dtype]))
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1241,12 +1241,24 @@ imhist{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins, minval, maxval) = im
 
 """
 ```
-hist_equalised_img = histeq(img, nbins)
-hist_equalised_img = histeq(img, nbins, minval, maxval)
+hist_equalised_img = histeq(img, nbins, dtype = "8bit")
+hist_equalised_img = histeq(img, nbins, minval, maxval, dtype = "8bit")
 ```
 
-Returns a grayscale histogram equalised image with a granularity of `nbins` number of bins. If minval and maxval are specified then
-only the intensities in the range (minval, maxval) are equalised.
+Returns a histogram equalised image with a granularity of approximately `nbins` 
+number of bins. An optional `dtype` argument (defaulting to 8bit) can be specified
+to choose the number of bits of the returned image. 
+
+The `histeq` function can handle a variety of input types. The returned image depends 
+on the input type. If the input is an `Image` then the resulting image is of the same type
+and has the same properties. 
+
+For coloured images, the input is converted to YCbCr type and the Y channel is equalised. This
+is the combined with the Cb and Cr channels and the resulting image converted to the same type 
+as the input.
+
+If minval and maxval are specified then only the intensities in the range 
+(minval, maxval) are equalised.
 
 """
 
--- a/docs/src/function_reference.md
+++ b/docs/src/function_reference.md
@@ -282,6 +282,7 @@ findlocalminima
 
 ```@docs
 imhist
+histeq
 ```
 
 # Filtering kernels
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -217,6 +217,7 @@ export # types
     imgaussiannoise,
     imgradients,
     imhist,
+    histeq,
     imlaplacian,
     imlineardiffusion,
     imlog,
@@ -280,7 +281,7 @@ Algorithms:
     - Resizing: `restrict`, `imresize` (not yet exported)
     - Filtering: `imfilter`, `imfilter_fft`, `imfilter_gaussian`, `imfilter_LoG`, `imROF`, `ncc`, `padarray`
     - Filtering kernels: `ando[345]`, `guassian2d`, `imaverage`, `imdog`, `imlaplacian`, `prewitt`, `sobel`
-    - Exposure : `imhist`
+    - Exposure : `imhist`, `histeq`
     - Gradients: `backdiffx`, `backdiffy`, `forwarddiffx`, `forwarddiffy`, `imgradients`
     - Edge detection: `imedge`, `imgradients`, `thin_edges`, `magnitude`, `phase`, `magnitudephase`, `orientation`
     - Corner detection: `imcorner`


mronian:
@bjarthur For Histogram Equalisation, I will need the maximum value for each datatype. For eg. in 8 bit Uint it would be 255 so I will have bins from 0-255 for the histogram. What is the best way to find this for all the types of Images?

timholy:
This was an excellent choice for a first pull request: it's a simple function, which gives you a chance to learn more about julia, Images, and the process of contributing to packages. :+1: for style.

I'll comment on the details in the code. (I was preparing a response to your previous version, and I noticed you just pushed a new version that addressed some of my concerns. Good timing!)

src/algorithms.jl
```diff
@@ -1029,6 +1029,30 @@ function copytail!(dest, A, coloffset, strd, len)
...
+"""
+```
+hist=imhist(img, 0:0.2:1)
+```
+
+Calculates a histogram of the given single channel image with the given range as bins. 
+The histogram is calculated on the flattened image. For multi dimensional images, use the function 
+separately on each channel to obtain a histogram for each.  
+"""
+function imhist{T}(img::AbstractArray{T,2}, nbins::Integer)
```
timholy: Is there a strong reason to limit this to 2d? Why couldn't I supply a 3d grayscale image?

src/algorithms.jl
```diff
@@ -1029,6 +1029,30 @@ function copytail!(dest, A, coloffset, strd, len)
...
+    bins=1/nbins:1/nbins:1
+    w,h=size(img)
+    histogram=zeros(Integer,nbins)
+    for i=1:w
+        for j=1:h
+            index=searchsortedfirst(bins, img[i,j])
```
timholy: Since `bins` is regularly spaced, `searchsortedfirst` isn't necessary: you should be able to compute the bin index directly from the value.

src/algorithms.jl
```diff
@@ -1029,6 +1029,30 @@ function copytail!(dest, A, coloffset, strd, len)
...
+```
+
+Calculates a histogram of the given single channel image with the given range as bins. 
+The histogram is calculated on the flattened image. For multi dimensional images, use the function 
+separately on each channel to obtain a histogram for each.  
+"""
+function imhist{T}(img::AbstractArray{T,2}, nbins::Integer)
+    bins=1/nbins:1/nbins:1
+    w,h=size(img)
+    histogram=zeros(Integer,nbins)
```
timholy: `Integer` is an abstract type, which means that performance will be poor. `Int` would be a better choice.

src/algorithms.jl
timholy: What about an RGB image? What should this do in that case?

timholy:
And now for a more general comment: do we need a separate function for this? Couldn't we just call `hist`?

timholy:
(Note that if you try `hist` on an image with element type `Gray`, you'll get an error. I'll be happy to coach you through how to interpret the error, where to supply the fix, etc.)

src/algorithms.jl
mronian: No reason as such. I will extend it to N dims.

mronian:
Yes. ```hist``` was giving errors for the Datatypes involved so I thought of doing it this way. I would like to learn how to use it with Images.

I am getting this error for Gray datatype.
```
ERROR: MethodError: no method matching histrange(::Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}, ::Int64)
Closest candidates are:
  histrange{T<:AbstractFloat,N}(::AbstractArray{T<:AbstractFloat,N}, ::Integer)
  histrange{T<:Integer,N}(::AbstractArray{T<:Integer,N}, ::Integer)
 in hist(::Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}, ::Int64) at ./statistics.jl:732
 in eval(::Module, ::Any) at ./boot.jl:267

```

I'll work on the other comments and resubmit my PR.

src/algorithms.jl
mronian: For an RGB Image, there can be multiple ways. We could calculate the histogram for each channel or  calculate the histogram of RGB values itself.

timholy:
So, what that error means is that you might need to define a new `histrange` method. You do so like this:
```jl
Base.histrange{T}(A::AbstractArray{Gray{T}}, n) = ...
```
where `...` is something I'm leaving for you to figure out. We should consider whether such a function goes here or in `ColorVectorSpace`.

One of the challenges we might face is whether to only compute histograms over `[0,1]`, or whether it should be over `[Gray(typemin(T)), Gray(typemax(T))]` (what about `T = Float64`, though?), or possibly `[minfinite(A), maxfinite(A)]`. Thoughts?

mronian:
I think we can add it to ColorVectorSpace as it has many other operations defined over the various Colors. I'll open an issue there and submit a PR.

I tried the ```hist(img[:], edg::AbstractVector)``` function. It works in this case. The other two ones where number of bins are given or just the data is given don't work due to the absence of a ```histrange``` function over ColorTypes. The ```hist(A, n)``` function anyways gives a range over ```[minfinite(A), maxfinite(A)]``` so that case is covered if we add the missing ```histrange``` function. For ```histeq``` we should use the ```hist(A, edg::AbstractVector)``` function with bins between [0,1] i.e. (1/n:1/n:1) where n is the number of bins desired. The fact that all Images are expressed using ColorTypes makes it even more easier as all Images are between [0,1].

mronian:
I was going through Color.jl and came to know that apart from RGB and XYZ, the others dont support the FixedPointNumbers package. This means we will have to take the histogram from ColorType(typemin(T)) to ColorType(typemax(T)) depending on the colorspace. Am I Right?

mronian:
Submitted PR JuliaGraphics/ColorVectorSpace.jl#33.

mronian:
PR JuliaGraphics/ColorVectorSpace.jl#35 adds zero() function for initialising cdf.

timholy:
> I was going through Color.jl ColorTypes.jl and came to know that apart from RGB and XYZ, the others dont support the FixedPointNumbers package.

Most types require `Fractional`, which is defined [here](https://github.com/JuliaGraphics/ColorTypes.jl/blob/2b0f939c11cafc1c0a266d780d45474d94c61380/src/ColorTypes.jl#L7). So they take fractional.

The problem with `typemin` and `typemax` is they give `Inf` for things like `Float32` and `Float64`. That probably won't be useful.

Here are some considerations: `RGB{U8}` doesn't allow ranges outside of 0, 1, but `RGB{Float32}` does. This is sometimes useful if, for example, you want to accumulate a value (e.g., in taking a mean). On the other hand, it's not _meaningful_ to have values outside the range 0, 1, because they won't be displayed any differently than their clamped values. So on balance I favor restricting the range to 0, 1 for `RGB`.

For other color types, you're right that the range might need to be different. Overall it seems best to use the [gamut](https://en.wikipedia.org/wiki/Gamut), but I'm not always sure how you find out what that is in all cases. (It's pretty obvious in some, but in others less so.) I suppose one option is to define `histrange` only for those cases where the answer is clear, and leave it undefined for the rest.

src/algorithms.jl
```diff
@@ -1029,6 +1029,42 @@ function copytail!(dest, A, coloffset, strd, len)
...
+# Histogram Equalisation Function
+"""
+```
+hist_equalised_img = histeq(img, nbins)
+
+Returns a grayscale histogram equalised image with a granularity of `nbins` number of bins.
+```
+
+"""
+function histeq{T,N}(img::AbstractArray{T,N}, nbins::Int)
```
timholy: Might want `histeq{T<:Union{Real,Gray},N}` to restrict to grayscale?

timholy: Oh, I see, you're doing it by conversion to grayscale, so never mind. To avoid creating too many intermediaries, you might want to dispatch to two functions:

```jl
histeq{T<:Real,N}(...
    # this one calls maxfinite and minfinite, and does the scaling
    histeq_rescale(imgv)
end

histeq{T<:Colorant,N}(...
    imgv = [convert(Gray, p) for p in img]
    histeq_rescale(imgv)
end
```
or something.

src/algorithms.jl
```diff
@@ -1029,6 +1029,42 @@ function copytail!(dest, A, coloffset, strd, len)
...
+    _,histogram = hist(imgv, 0:1/nbins:1)
+    cdfmin = 0
+    cdf = zeros(Int64, nbins)
+    for i = 1:nbins
+        cdf[i:nbins] += histogram[i]
+        if cdf[i] == 0
+            cdfmin += 1
+        end
+    end
+    retimg = [(imgv[i] == 0)? 1:Int(ceil(nbins*imgv[i])) for i = 1:length(imgv)]
```
timholy: You could use `[max(1, ceil(Int, nbins*x)) for x in imgv]` here.

src/algorithms.jl
```diff
@@ -1029,6 +1029,42 @@ function copytail!(dest, A, coloffset, strd, len)
...
+    end
+    _,histogram = hist(imgv, 0:1/nbins:1)
+    cdfmin = 0
+    cdf = zeros(Int64, nbins)
+    for i = 1:nbins
+        cdf[i:nbins] += histogram[i]
+        if cdf[i] == 0
+            cdfmin += 1
+        end
+    end
```
timholy: Putting the branch inside the loop will probably be no better than calling `cumsum` followed by `findfirst`.

src/algorithms.jl
```diff
@@ -1029,6 +1029,42 @@ function copytail!(dest, A, coloffset, strd, len)
...
+    retimg = [(imgv[i] == 0)? 1:Int(ceil(nbins*imgv[i])) for i = 1:length(imgv)]
+    retimg = [newmaxintensity*((cdf[retimg[i]]-cdfmin)/(length(imgv)-cdfmin)) for i = 1:length(retimg)]
```
timholy: Not sure I understand the denominator here. Isn't it `nbins - cdfmin`?

src/algorithms.jl
```diff
@@ -1029,6 +1029,42 @@ function copytail!(dest, A, coloffset, strd, len)
...
+    retimg = reshape(retimg, size(img))
```
timholy: This line and the next would make `retimg` type-unstable: it's a `Vector` above but now it will generally be multidimensional. Use a different name.

src/algorithms.jl
mronian: length(imgv) is the size of the image i.e. the last value of cdf. I'll change it to cdf[-1] to make it simpler. I am scaling the intensity to between 0 and 1.

mronian:
I am getting this error sometimes if I use the julia notebook to display images. 
```
WARNING: MethodError(Images.mapinfo_writemime_,(
Gray Images.Image with:
  data: 256x256 Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}
  properties:
    timedim: 0
    spatialorder:  y x
    pixelspacing:  1.0 1.0,))
 in #mapinfo_writemime#67(::Int64, ::Any, ::Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}}) at /home/jarvis/.julia/v0.5/Images/src/writemime.jl:40
 in #save_#10(::Array{Any,1}, ::ImageMagick.#save_, ::FileIO.Stream{FileIO.DataFormat{:PNG},Base64EncodePipe}, ::Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}}) at /home/jarvis/.julia/v0.5/ImageMagick/src/ImageMagick.jl:151
 [inlined code] from ./boot.jl:331
 in #save#6(::Array{Any,1}, ::Any, ::FileIO.Stream{FileIO.DataFormat{:PNG},Base64EncodePipe}, ::Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}}, ::Vararg{Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}}}) at /home/jarvis/.julia/v0.5/ImageMagick/src/ImageMagick.jl:67
 [inlined code] from ./boot.jl:331
 in #save#18(::Array{Any,1}, ::Any, ::FileIO.Stream{FileIO.DataFormat{:PNG},Base64EncodePipe}, ::Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}}, ::Vararg{Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}}}) at /home/jarvis/.julia/v0.5/FileIO/src/loadsave.jl:95
 [inlined code] from ./boot.jl:331
 in #writemime#64(::Images.MapNone{ColorTypes.RGB{FixedPointNumbers.UFixed{UInt8,8}}}, ::Int64, ::Int64, ::Any, ::Base64EncodePipe, ::MIME{symbol("image/png")}, ::Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}}) at /home/jarvis/.julia/v0.5/Images/src/writemime.jl:30
 in base64encode(::Any, ::MIME{symbol("image/png")}, ::Vararg{Any}) at ./base64.jl:159
 in stringmime(::MIME{symbol("image/png")}, ::Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}}) at ./multimedia.jl:84
 in display_dict(::Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}}) at /home/jarvis/.julia/v0.5/IJulia/src/execute_request.jl:32
 in execute_request_0x535c5df2(::ZMQ.Socket, ::IJulia.Msg) at /home/jarvis/.julia/v0.5/IJulia/src/execute_request.jl:212
 [inlined code] from ./dict.jl:766
 in eventloop(::ZMQ.Socket) at /home/jarvis/.julia/v0.5/IJulia/src/IJulia.jl:141
 in (::IJulia.##24#30)() at ./task.jl:431
MethodError: no method matching mapinfo_writemime_(::Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}})
```

mronian:
There is a problem with scaling the image to values between [0-1] for the histogram. If there are intensities equal to 0, the histogram function will miss out on the 0 if we use the range 0:1/nbins:1. I feel it would be better to consider the bin centres instead i.e. take the range from 1/(2*nbins):1/nbins:(1+1/(2*nbins)). Is there a better way to do this?

timholy:
Bin centers seem like a natural way to solve this. If you like that strategy, go for it.

If you prefer edge-based, then (given your observation) I have no objection to the "code duplication" of writing your own histogram algorithm. Your observation here is convincing incentive. So I'd say choose whichever path you like better.

timholy:
Regarding the error in the notebook (which is a display error, not a "computation" error), I'm puzzled. Why isn't it dispatching to [this method](https://github.com/timholy/Images.jl/blob/e6af9e1497bad82306383d3a5035b5f4d2c53acf/src/writemime.jl#L49)?

mronian:
We could do it this way (It combines some of the suggestions discussed above)- 

For any image, get the following attributes -
```
minval=minfinite(raw(img))
maxval=maxfinite(raw(img))
tmax=typemax(eltype(raw(img)))
```
Calculate the histogram using the range ```(0-tmax+1)``` divided into ```nbins```. Basically I am converting the image back to the original datatype and then calculating the histogram over that range for the best result. The range ```(0-tmax+1)``` will consider all possible values (0 too) by making the bins left aligned. 
As you suggested above, we use the histogram from ```(minval, maxval)``` and stretch it to the above range. 

The returned image will also be grayscale, based on the original datatype instead of UInt8 always. I'll implement these functions and add a commit.

mronian:
@timholy In the ```histeq{T<:Colorant, N}``` function, if I use -

```
    img = convert(Image{ColorTypes.Gray}, img)
    _histeq(img, nbins)
```

I get the following error - 

```
ERROR: MethodError: Cannot `convert` an object of type Images.Image{ColorTypes.RGB{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.RGB{FixedPointNumbers.UFixed{UInt8,8}},2}} to an object of type Images.Image{ColorTypes.Gray{T<:Union{AbstractFloat,FixedPointNumbers.FixedPoint{T<:Integer,f}}},N,A<:AbstractArray{T,N}}
This may have arisen from a call to the constructor Images.Image{ColorTypes.Gray{T<:Union{AbstractFloat,FixedPointNumbers.FixedPoint{T<:Integer,f}}},N,A<:AbstractArray{T,N}}(...),
since type constructors fall back to convert methods.
Closest candidates are:
  convert{C<:ColorTypes.Colorant{T,N}}(::Type{Images.Image{C<:ColorTypes.Colorant{T,N},N,A<:AbstractArray{T,N}}}, ::Images.Image{C<:ColorTypes.Colorant{T,N},N,A<:AbstractArray{T,N}})
  convert{T}(::Type{Images.Image{T,N,A<:AbstractArray{T,N}}}, ::Images.Image{T,N,A<:AbstractArray{T,N}})
  convert{Cdest<:ColorTypes.Colorant{T,N},Csrc<:ColorTypes.Colorant{T,N}}(::Type{Images.Image{Cdest<:ColorTypes.Colorant{T,N},N,A<:AbstractArray{T,N}}}, ::Images.Image{Csrc<:ColorTypes.Colorant{T,N},N,A<:AbstractArray{T,N}})
  ...
 in histeq(::Images.Image{ColorTypes.RGB{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.RGB{FixedPointNumbers.UFixed{UInt8,8}},2}}, ::Int64) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1089 (repeats 2 times)
 in eval(::Module, ::Any) at ./boot.jl:267

```

The ```convert``` function works from the console on the same image though - 
```
julia> convert(Image{ColorTypes.Gray}, img)
Gray Images.Image with:
  data: 256x256 Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}
  properties:
    colorspace: sRGB
    spatialorder:  x y
```

I've amended the previous commit with the imhist and histeq functions for T<:Colorant. There is an issue with using the same functions for T<:Real as typemax might be really huge as compared to Image values. We could either have the user specify the typemax i.e. the range he wants to spread the histogram to or we could spread it in ```(minval, maxval)```

timholy:
Sorry I've been absent, my julia time has been taken up by my efforts on the general array infrastructure.

> maxval=maxfinite(raw(img))

A problem with using `raw` is that it converts `U8->UInt8`, in which case `maxfinite` might yield 255 != 1. Better to keep the innate representation; `separate` might be the better choice.

Regarding the `convert` error, I do find that puzzling; I'll try to look at it myself. (Is this branch the version you want me to test?) Another implementation would be to define `histeq{T<:Union{Number,Gray},N}` as the "core function" and then
```jl
imhist{T<:Colorant,N}(img::AbstractImage{T,N}, nbins::Int = 400) = imhist(convert(Image{Gray}, img), nbins)
imhist{T<:Colorant,N}(img::AbstractArray{T,N}, nbins::Int = 400) = imhist(convert(Array{Gray}, img), nbins)
```

timholy:
I have a suggested alternative implementation of `imhist`:
```jl
imhist{T<:Colorant}(img::AbstractArray{T}, nbins=400) = imhist(convert(Array{Gray}, data(img)), nbins)

function imhist{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins = 400)
    minval = minfinite(img)
    maxval = maxfinite(img)
    imhist(img, nbins, minval, maxval)
end

function imhist{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins, minval::T, maxval::T)
    bins = minval:(maxval-minval)/(nbins-1):maxval
    histogram = zeros(Integer,nbins+2)
    for val in img
        index = searchsortedlast(bins, val)
        histogram[index+1] += 1
    end
    bins, histogram
end
imhist{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins, minval, maxval) = imhist(img, nbins, convert(T, minval), convert(T, maxval))
```

There are several things to like about this implementation:
- It allows the user to specify the histogram limits, if desired
- It uses two extra bins to accumulate any pixels less than the minimum or greater than the maximum
- It doesn't perform any unnecessary conversion/copying
- It ensures that the range is commensurate with the data type of the image itself

The extra two bins could be trouble for your histogram equalization algorithm; I didn't even look at that.

timholy:
@mronian, did you want to keep going with this? Would you prefer me to commit this version of `imhist`?

mronian:
Yes I would like to think over it for some time before you merge it. I'll add a commit here with either your version or a modified one by tomorrow. Please go ahead if it is urgently required though :)

timholy:
No urgency.

mronian:
Really excited to start with the coding period :)

@timholy I am implementing the following functions for ```histeq``` after giving it some thought and the discussions on ```imhist```. Some of them will require additional functions for ```imhist```.

For ```T<:Colorant```, I found a really nice way of doing histogram equalization [online](http://opencv-srf.blogspot.in/2013/08/histogram-equalization.html). They were converting the ColorSpace to YCbCr and then equalizing only the ```Y``` channel and returning the resulting YCbCr image with the original ```Cb, Cr``` channels. We can return the image in the original datatype too. 

For ```T<:Gray``` but ```N>2```, we can allow the user to specify which channel should be equalized. If nothing is specified, then we equalize each channel separately.

For ```T<:Gray``` and ```N=2```, we equalize the image like before. The only addition being, if a (minval, maxval) range is specified then we equalise the values only in that range. This applies to the above two cases also.

One thing which I removed is equalizing the histogram to a user specified range. Due to the different datatypes involved, it will not be possible to offer just one standard function to suit all. It will be easier to return the equalized image in a standard type (Say two options UInt8 and UInt16. OpenCV returns an 8 bit image). The user can then easily scale it as required by a simple multiplication. I'll submit a PR with the above code by tonight. It requires #490 to be fixed though since most of the functions deal with proper Colorspaces and not just Number types.

```imhist``` can also be extended in the same way to produce histogram for the intensity channel in a color image instead of converting to Grayscale. Similarly for N channel images we can produce a stack of histograms. I am not sure of this though and would like to know your thoughts.

**Comment by mronian** [comment]
Tests for ```histeq``` and functions without the (minval, maxval) range remain.

File: src/algorithms.jl [hunk 9]
```diff
@@ -1231,6 +1231,56 @@ end
...
+
+function histeq{T<:Gray, N}(img::AbstractArray{T, N}, nbins, minval, maxval)
+    hist_equalised_stack = []
+    for i in 1:1:N
+        push!(hist_equalised_stack, histeq(img, nbins, minval, maxval))
+    end
+    hist_equalised_stack
+end
+
+histeq{T<:Gray, N}(img::AbstractArray{T, N}, nbins, minval, maxval, nchannel) = histeq(img[:,:,nchannel], nbins, minval, maxval)
```
timholy: What's the use here? Why couldn't the user just pass `img[:,:,nchannel]` if that's what (s)he wants?

src/algorithms.jl
```diff
@@ -1231,6 +1231,56 @@ end
...
+function histeq{T<:Colorant}(img::AbstractArray{T}, nbins, minval, maxval)
+    YCbCr = convert(Array{ColorTypes.YCbCr}, data(img))
+    Y = separate(YCbCr )[:,:,1]
+    eq_Y = []
+    if maxval == 0
+        eq_Y = histeq(Y, nbins)
+    else 
+        eq_Y = histeq(Y, nbins, minval, maxval)
+    end
+    hist_equalised_img = [ColorTypes.YCbCr(eq_Y[i], YCbCr[i].cb, YCbCr[i].cr) for i in 1:1:length(YCbCr)]
```
timholy: Just use `1:length(YCbCr)`, it makes it slower if you use a step parameter when you don't need it.

src/algorithms.jl
```diff
@@ -1231,6 +1231,56 @@ end
...
+    hist_equalised_img = reshape(hist_equalised_img, size(YCbCr))
+    hist_equalised_img = convert(Array{T}, hist_equalised_img)
+end
+
+# histeq{T<:Gray, N}(img::AbstractArray{T, N}, nbins) = histeq(img, nbins, typemin(img), 0)
+
+function histeq{T<:Gray, N}(img::AbstractArray{T, N}, nbins, minval, maxval)
+    hist_equalised_stack = []
+    for i in 1:1:N
+        push!(hist_equalised_stack, histeq(img, nbins, minval, maxval))
```
timholy: I think there's a bug here, and that you probably meant to supply `i` as a `nchannel` (see the function below this one).

However, I find this fairly problematic. What if this is a 3d image, e.g., a CT or MRI scan? One would almost always want the 3rd dimension treated identically to the other two.

timholy: It's also easy to do what you're asking via `mapslices(histeq, img, 3)`.

src/algorithms.jl
timholy: Actually, even better will be
```jl
[ColorTypes.YCbCr(eq_Y[i], col.cb, col.cr) for (i,col) in enumerate(YCbCr)]
```
It saves an extra array lookup.

src/algorithms.jl
```diff
@@ -1231,6 +1231,56 @@ end
...
+    end
+    hist_equalised_stack
+end
+
+histeq{T<:Gray, N}(img::AbstractArray{T, N}, nbins, minval, maxval, nchannel) = histeq(img[:,:,nchannel], nbins, minval, maxval)
+
+function histeq{T<:Union{Gray,Number}}(img::AbstractArray{T, 2}, nbins, minval::T, maxval::T)
```
timholy: Don't limit to 2d.

src/algorithms.jl
```diff
@@ -1231,6 +1231,56 @@ end
...
+Returns a grayscale histogram equalised image with a granularity of `nbins` number of bins. If minval and maxval are specified then
+only the intensities in the range (minval, maxval) are equalised.
+
+"""
+
+histeq{T<:Colorant}(img::AbstractArray{T}, nbins) = histeq(img, nbins, 0, 0)
+
+function histeq{T<:Colorant}(img::AbstractArray{T}, nbins, minval, maxval)
+    YCbCr = convert(Array{ColorTypes.YCbCr}, data(img))
+    Y = separate(YCbCr )[:,:,1]
```
timholy: Hmm, this also assumes the input image is 2d. You can use `colons = ntuple(d->Colon(), ndims(img))` and then `[colons..., 1]`.

src/algorithms.jl
```diff
@@ -1231,6 +1231,56 @@ end
...
+    eq_Y = []
```
timholy: Just delete this; it will make `eq_Y` type-unstable, and then the comprehension below will be really, really slow.

src/algorithms.jl
```diff
@@ -1231,6 +1231,56 @@ end
...
+    if maxval == 0
+        eq_Y = histeq(Y, nbins)
+    else 
+        eq_Y = histeq(Y, nbins, minval, maxval)
+    end
+    hist_equalised_img = [ColorTypes.YCbCr(eq_Y[i], YCbCr[i].cb, YCbCr[i].cr) for i in 1:1:length(YCbCr)]
+    hist_equalised_img = reshape(hist_equalised_img, size(YCbCr))
+    hist_equalised_img = convert(Array{T}, hist_equalised_img)
```
timholy: Should we restore the `Image` wrapper if it was supplied? You can do that by adding an extra method for `AbstractImage` that makes use of `copyproperties` or `shareproperties`.

src/algorithms.jl
mronian: Agreed. I'll remove it.

src/algorithms.jl
mronian: What would be the best way to do ```histeq``` for an N dimensional image? I have not worked much with N>3 grayscale images so I dont know what operations are required usually. By treating the 3rd dimension like the other two, do you mean that the histogram should be calculated over all the channels and then equalised? As in, taking the probability distribution of intensities over all channels? If that is the case, should we keep it as the only option or is the current one useful in some cases?

src/algorithms.jl
mronian: I guess this should be merged with the function for N dimensional images if we choose to calculate the histogram over all channels.

src/algorithms.jl
mronian: I hadn't thought about 3D in colorspaces. I will try and add some test images to ```TestImages.jl``` to so that I can test easily. If possible could you direct me to some sources?

mronian: I hadn't thought about 3D in colorspaces. I will try and add some test images to ```TestImages.jl``` to so that I can test easily. If possible could you direct me to some sources?

src/algorithms.jl
mronian: Sounds great. I'll do that in and push a new version for review after doing the changes mentioned above.

mronian:
@timholy Just wanted to confirm if you get notifications on this thread when I comment or should I tag you like you had asked me to on some other thread?

src/algorithms.jl
timholy: I searched briefly, but such images often tend to be so large people must not post them much. Some 2d renderings are shown, e.g., https://www.google.com/search?q=confocal+microscopy+image+stack+color+example+images&client=ubuntu&espv=2&biw=1920&bih=1080&tbm=isch&tbo=u&source=univ&sa=X&ved=0ahUKEwjWws_a9fXMAhUSNFIKHbDbBFMQsAQIGw, but that's not really useful for this purpose. (If you want to search more, you can see the terms I used in that URL. "confocal microscopy" is frequently multicolor, and "image stack" means a 3d image.)

You could just generate them.

src/algorithms.jl
timholy: What do you mean "channels"? I usually think of "channels" as meaning "color channels," but if the image has more than one channel it's not `Gray`. `Gray` means the image is really, truly, a grayscale image, so if it has more than 2 dimensions it's either a movie (over time) or a stack (over z). (If one is using `separate` to separate color channels, then the element type should be a subtype of `Number`, not a `Colorant`.)

The way you have it here, in a 3d image each 2d slice will be separately equalized. The problem is that this may effectively cause the contrast to vary from one slice to the next. You'd probably be unhappy, too, if your 2d image were equalized separately on each row of the image.

If you just take out all the 2d-specific parts of your implementation, then it handles a 3d image as a whole, and this ensures a consistent contrast change.

src/algorithms.jl
timholy: It's useful to have a separate method for grayscale; converting grayscale to YCbCr will triple the size of the image unnecessarily. I think if you literally get rid of that `, 2` in the type declaration of `img`, you'll be fine.

timholy:
Yes, I get notification anytime you comment; I *don't* get notifications if you add more commits to the PR, so if you do and you want me to review, add some kind of comment.

mronian:
Is there an easy way to convert the inner datatype of the value in ColorType (UInt, Float etc) to another type?

This is required to convert the equalised image to a standard 8bit/16bit format so that in cases where (minval, maxval) is not given, we can take the (typemin, typemax) of the datatype without problems like ```typemax=infinity``` (Float) etc.

timholy:
The low-level syntax is:
```jl
convert(base_colorant_type(c){Float32}, c)
```
`base_colorant_type` is one of the low-level "trait-functions" of ColorTypes, which is the category of operations that gives ColorTypes most of its power. That syntax basically allows you to swap out the number-type for any one of your choosing, but keeping the outer wrapper type.

For arrays, Images exports the old `float32` and `float64` to make this easier.

mronian:
All functionalities have been implemented. Some things are not working for me though, I am commenting on the code to mark those.

src/algorithms.jl
```diff
@@ -1239,6 +1239,74 @@ end
...
+end
+
+histeq{T<:Colorant}(img::AbstractArray{T}, nbins, dtype = "8bit") = histeq(_prep_image_for_histeq(img, dtype), nbins, zero(YCbCr), zero(YCbCr))
+histeq{T<:Colorant}(img::AbstractArray{T}, nbins, minval, maxval, dtype = "8bit") = histeq(_prep_image_for_histeq(img, dtype), nbins, minval, maxval)
+
+function histeq{T<:Colorant}(img::AbstractArray{T}, nbins, minval, maxval)
+    YCbCr = convert(Array{ColorTypes.YCbCr}, data(img))
+    colons = ntuple(d->Colon(), ndims(img))
+    Y = separate(YCbCr )[colons..., 1]
+    if maxval == zero(YCbCr)
```
mronian: This does not work for some reason. I tried reproducing it on the console and it works there - 
```
julia> maxval = zero(YCbCr)
YCbCr{Float32}(0.0f0,0.0f0,0.0f0)

julia> maxval == zero(YCbCr)
true
```

src/algorithms.jl
```diff
@@ -1239,6 +1239,74 @@ end
...
+        eq_Y = histeq(Y, nbins, minfinite(Y), maxfinite(Y))
```
mronian: Instead of minfinite and maxfinite, I'd like to take the entire range of the Y Channel possible. This depends on the type of conversion used though, so I wanted to check with you. In [this](https://en.wikipedia.org/wiki/YCbCr) it uses the range (16,235) for the Y Channel in 8 bits. Is this the correct one?

src/algorithms.jl
```diff
@@ -1239,6 +1239,74 @@ end
...
+    else
+        eq_Y = histeq(Y, nbins, minval, maxval)
+    end
+    hist_equalised_img = [ColorTypes.YCbCr(eq_Y[i], col.cb, col.cr) for (i,col) in enumerate(YCbCr)]
+    hist_equalised_img = reshape(hist_equalised_img, size(YCbCr))
+    hist_equalised_img = convert(Array{T}, hist_equalised_img)
+end
+
+function histeq(img::AbstractImage, nbins, dtype = "8bit")
```
mronian: For some reason, images of type ```Image``` are not directed to this(and the next) function. These seem to give wrong answers too -
```
julia> img1
RGB Images.Image with:
  data: 512×512 Array{ColorTypes.RGB{FixedPointNumbers.UFixed{UInt8,8}},2}
  properties:
    colorspace: sRGB
    spatialorder:  x y

julia> typeof(img1)
Images.Image{ColorTypes.RGB{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.RGB{FixedPointNumbers.UFixed{UInt8,8}},2}}

julia> typeof(img1) <: Images.Image
false

julia> typeof(img1) <: Images.AbstractImage
false

julia> typeof(img1) <: AbstractImage
false
```

src/algorithms.jl
```diff
@@ -1239,6 +1239,74 @@ end
...
+    hist_equalised_img = Image(histeq(data(img), nbins, dtype))
+    shareproperties(img, hist_equalised_img)
```
mronian: ```shareproperties``` & ```copyproperties``` is not working for me. I tried this on the console - 
```
julia> img2
Gray Images.Image with:
  data: 256×256 Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}
  properties:
    colorspace: Gray
    spatialorder:  x y

julia> properties(img2)
Dict{String,Any} with 3 entries:
  "timedim"      => 0
  "spatialorder" => String["y","x"]
  "pixelspacing" => [1.0,1.0]

julia> ret
Gray Images.Image with:
  data: 256×256 Array{ColorTypes.Gray{Float64},2}
  properties:

julia> copyproperties(img2, ret)
Gray Images.Image with:
  data: 256×256 Array{ColorTypes.Gray{Float64},2}
  properties:

julia> shareproperties(img2, ret)
Gray Images.Image with:
  data: 256×256 Array{ColorTypes.Gray{Float64},2}
  properties:
```

src/algorithms.jl
timholy: Dunno, can you show an example of it failing? If I had to guess blindly, I'd say it probably comes from the fact that `YCbCr` is a parametric type:
```jl
julia> zero(YCbCr{Float64}) == zero(YCbCr)
false
```
It seems to default to `Float32` if you don't specify the type.

src/algorithms.jl
timholy: Best way to check would be to look at the `convert` methods for YCbCr in `Colors.jl`. This could surely use documentation, if it isn't documented already.

src/algorithms.jl
timholy: Are you sure you don't have a stale module problem? All those evaluate to `true` for me. Try a fresh julia session.

src/algorithms.jl
timholy: Works for me. Again, maybe try in a fresh julia session?

src/algorithms.jl
mronian: Gosh. Had no idea about this issue. It worked in the new session.

src/algorithms.jl
mronian: Again, worked in the new session. Thanks!

mronian:
Finally Done! It works very well with multiple types of Images and ranges. The YCbCr conversion method used is the one from wikipedia. One slight change I have made is that while calculating the corrected intensity, it takes ```hist_length``` instead of ```nbin``` since the histogram calculated by ```StatsBase``` doesnt always return an histogram of length exactly ```nbins``` which was causing out-of-bounds exceptions.

mronian:
Need to add some tests now. How are things like this which are very subjective and depend on the approach used usually tested? I mean, if someone were to modify the function a bit, the values returned would be different. Should I put tests to check if the return types etc are correct instead of testing the data returned?

src/algorithms.jl
timholy: It's quite subtle, but the way to think about it is that if you load a module `Mod`, and then `reload` or `include` it again, then any objects created before reloading have an `OldMod` type (even though the name still prints as `Mod`). For your tests, you have to define new objects for the new types.

src/algorithms.jl
```diff
@@ -1239,6 +1239,77 @@ end
 
 imhist{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins, minval, maxval) = imhist(img, nbins, convert(T, minval), convert(T, maxval))
 
+"""
+```
+hist_equalised_img = histeq(img, nbins)
+hist_equalised_img = histeq(img, nbins, minval, maxval)
+```
+ColorVectorSpace.
```
timholy: What is this line for?

mronian: Thats crept in by mistake when I was switching between the terminal and my editor. Updating now.

src/algorithms.jl
```diff
@@ -1239,6 +1239,77 @@ end
...
+    end
+    img = reshape(img, img_shape)
+    img
+end
+
+histeq{T<:Colorant}(img::AbstractArray{T}, nbins, dtype = "8bit") = histeq(_prep_image_for_histeq(img, dtype), nbins, zero(YCbCr), zero(YCbCr))
+histeq{T<:Colorant}(img::AbstractArray{T}, nbins, minval, maxval, dtype = "8bit") = histeq(_prep_image_for_histeq(img, dtype), nbins, minval, maxval)
+
+function histeq{T<:Colorant}(img::AbstractArray{T}, nbins, minval, maxval)
+    YCbCr = convert(Array{ColorTypes.YCbCr}, data(img))
```
timholy: Is YCbCr not exported by ColorTypes?

src/algorithms.jl
```diff
@@ -1239,6 +1239,77 @@ end
...
+    hist_equalised_img = histeq(data(img), nbins, dtype)
+    hist_equalised_img = shareproperties(img, hist_equalised_img)
+    hist_equalised_img
+end
+
+function histeq(img::AbstractImage, nbins, minval, maxval, dtype = "8bit")
+    hist_equalised_img = histeq(data(img), nbins, minval, maxval, dtype)
+    hist_equalised_img = shareproperties(img, hist_equalised_img)
+    hist_equalised_img
+end
```
timholy: Can you merge this and the previous into `histeq(img::AbstractImage, args...)`?

timholy:
Looking very good!

src/algorithms.jl
mronian: It was giving an error without that.

mronian: Oh. It was because I am naming my images as YCbCr. Silly mistake :)

src/algorithms.jl
mronian: Result of merging the two methods gives the following -
```
julia> detect_ambiguities(Images)
Skipping Images.indexedcolor
Skipping Images.iterate_spatial
Skipping Images.lut
Skipping Images.palette_fire
Skipping Images.palette_gray32
Skipping Images.palette_gray64
Skipping Images.palette_rainbow
Skipping Images.uint16sc
Skipping Images.uint32sc
Skipping Images.uint8sc
4-element Array{Tuple{Method,Method},1}:
 (histeq(img::Images.AbstractImage, nbins, minval, maxval) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1292,histeq{T<:Union{ColorTypes.Gray{T<:Union{AbstractFloat,Bool,FixedPointNumbers.FixedPoint}},Number}}(img::AbstractArray{T,N<:Any}, nbins::Int64, minval::T, maxval::T) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1300)
 (histeq(img::Images.AbstractImage, args...) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1358,histeq{T<:Union{ColorTypes.Gray{T<:Union{AbstractFloat,Bool,FixedPointNumbers.FixedPoint}},Number}}(img::AbstractArray{T,N<:Any}, nbins::Int64, minval::T, maxval::T) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1372)              
 (histeq{T<:Union{ColorTypes.Gray{T<:Union{AbstractFloat,Bool,FixedPointNumbers.FixedPoint}},Number}}(img::AbstractArray{T,N<:Any}, nbins::Int64, minval::T, maxval::T) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1300,histeq(img::Images.AbstractImage, args...) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1358)              
 (histeq(img::Images.AbstractImage, nbins, minval, maxval) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1292,histeq{T<:Union{ColorTypes.Gray{T<:Union{AbstractFloat,Bool,FixedPointNumbers.FixedPoint}},Number}}(img::AbstractArray{T,N<:Any}, nbins::Int64, minval::T, maxval::T) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1372)
```

Without merging I get - 
```
julia> detect_ambiguities(Images)
Skipping Images.indexedcolor
Skipping Images.iterate_spatial
Skipping Images.lut
Skipping Images.palette_fire
Skipping Images.palette_gray32
Skipping Images.palette_gray64
Skipping Images.palette_rainbow
Skipping Images.uint16sc
Skipping Images.uint32sc
Skipping Images.uint8sc
2-element Array{Tuple{Method,Method},1}:
 (histeq(img::Images.AbstractImage, nbins, minval, maxval) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1292,histeq{T<:Union{ColorTypes.Gray{T<:Union{AbstractFloat,Bool,FixedPointNumbers.FixedPoint}},Number}}(img::AbstractArray{T,N<:Any}, nbins::Int64, minval::T, maxval::T) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1372)
 (histeq{T<:Union{ColorTypes.Gray{T<:Union{AbstractFloat,Bool,FixedPointNumbers.FixedPoint}},Number}}(img::AbstractArray{T,N<:Any}, nbins::Int64, minval::T, maxval::T) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1300,histeq(img::Images.AbstractImage, nbins, minval, maxval) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1364)
```

mronian:
@timholy I forgot about Transparent Images. Is there a way to strip the entire alpha channel of a Colorant and then rejoin it after equalisation is done? I know about the ```alpha``` function but I think it can only be used on an element. Also, is there a way to know the non-transparent version of a colorspace given a TransparentColor element? This will help to handle all the different types with one function.

src/algorithms.jl
timholy: Looks easiest to solve the ambiguity if you don't merge the methods. Just make sure the last 3 arguments in the two ambiguous methods have the same signature constraints, and you should be fine.

timholy:
Actually, in general it looks like a better pattern would be
```
Y = [convert(YCbCr, color(c)).y for c in img]   # this strips out the alpha
eq_Y = ...
histeq_img = reshape([recompose_y(c, eq_Y[i]) for (i,c) in enumerate(img)], size(img))
```
and define separate `recompose_y` methods for transparent vs opaque colors. This way you only end up creating three arrays: the output array, one temporary `Y`, and one temporary `eq_Y`. This might make up for the fact that `recompose_y` probably has to compute the YCbCr conversion all over again. (From a performance perspective, in principle the `convert(YCbCr, color(c)).y` might allow the compiler to skip calculation of the other channels of `YCbCr`, so this may not be as much redundant work as it seems. Having that work in practice may depend on inlining, however.)

timholy:
In general, lots of "convert this array to that array" operations are characteristic of high-level languages that don't have good performance: it's how you "batch" work so that the (fast) underlying C code can do the real work. In julia, there's no need for that: focus on designing a good API for handling individual elements, chain all of your operations together on the elements, and then rebuild arrays as the "outermost" step.

mronian:
@timholy I've made the changes. Please Review :)

timholy:
Sorry for the long delay, the gigantic https://github.com/JuliaLang/julia/pull/16260 was all-consuming.

This looks good, I'm merging. At some point we should add some tests, though.

PR #465, opened 2016-04-06
Fix some conversion bugs & rebase on ccolor

timholy:
In the course of considering #458, I noticed some missing methods and errors in converting to grayscale. I took the opportunity to rebase some of the conversion methods on `ccolor` from ColorTypes; this greatly simplifies the selection of a concrete color type. (There are probably more places where this could be leveraged.)

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -323,6 +323,14 @@ end
 
 
 ## convert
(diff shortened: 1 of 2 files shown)


mronian:
Thanks! ```convert``` works now in ```histeq```.

Issue #463, opened 2016-04-04
Error on save method


mronian:
I get the following error on images returned from the histeq function in PR #458. The ```save``` method does not work with Images of the ```Image``` type. If I do ```save("test.jpg", res.data)``` then it works. 

```
julia> res = Images.histeq(img);

julia> save("test.jpg", res)
WARNING: MethodError(reinterpret,(UInt8,
Gray Images.Image with:
  data: 256x256 Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}
  properties:
    timedim: 0
    spatialorder:  x y
    pixelspacing:  1.0 1.0,

(256,256)))
 [inlined code] from ./array.jl:17
 in to_explicit(::Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}}) at /home/jarvis/.julia/v0.5/ImageMagick/src/ImageMagick.jl:231
 in image2wand(::Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}}, ::Images.MapNone{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}}}, ::Void, ::Bool) at /home/jarvis/.julia/v0.5/ImageMagick/src/ImageMagick.jl:186
 in #save_#9(::Images.MapNone{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}}}, ::Void, ::Any, ::UTF8String, ::Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}}, ::Bool) at /home/jarvis/.julia/v0.5/ImageMagick/src/ImageMagick.jl:146 (repeats 2 times)
 [inlined code] from ./boot.jl:331
[...]

Issue #460, opened 2016-03-25
Overlay doesn't work with PyPlot


bjarthur:
i can `ImageView.view` an `Overlay` just fine, but i cannot `PyPlot.imshow`.  is it because `raw` and `separate` don't work?

```
julia> using Images, ColorTypes, TestImages

julia> mandril = testimage("mandril_gray")
Gray Images.Image with:
  data: 512x512 Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}
  properties:
    imagedescription: <suppressed>
    spatialorder:  x y
    pixelspacing:  1 1

julia> ovr = Overlay((mandril,mandril),(RGB(1,0,0),RGB(0,1,0)));

julia> using PyPlot

julia> imshow(ovr)
2016-03-25 08:53:03.909 julia[5481:41933] setCanCycle: is deprecated.  Please use setCollectionBehavior instead
2016-03-25 08:53:03.918 julia[5481:41933] setCanCycle: is deprecated.  Please use setCollectionBehavior instead
ERROR: PyError (:PyObject_Call) <type 'exceptions.TypeError'>
TypeError(u'Image data can not convert to float',)
  File "/Users/arthurb/.julia/v0.4/Conda/deps/usr/lib/python2.7/site-packages/matplotlib/pyplot.py", line 3022, in imshow
    **kwargs)
  File "/Users/arthurb/.julia/v0.4/Conda/deps/usr/lib/python2.7/site-packages/matplotlib/__init__.py", line 1811, in inner
    return func(ax, *args, **kwargs)
  File "/Users/arthurb/.julia/v0.4/Conda/deps/usr/lib/python2.7/site-packages/matplotlib/axes/_axes.py", line 4943, in imshow
    im.set_data(X)
  File "/Users/arthurb/.julia/v0.4/Conda/deps/usr/lib/python2.7/site-packages/matplotlib/image.py", line 449, in set_data
    raise TypeError("Image data can not convert to float")

 [inlined code] from /Users/arthurb/.julia/v0.4/PyCall/src/exception.jl:81
 in pycall at /Users/arthurb/.julia/v0.4/PyCall/src/PyCall.jl:344
 in imshow at /Users/arthurb/.julia/v0.4/PyPlot/src/PyPlot.jl:460

julia> raw(ovr)
ERROR: auto_unbox: unable to determine argument type
 in raw at /Users/arthurb/.julia/v0.4/Images/src/core.jl:318

[...]

PR #467, opened 2016-04-06
Fix raw and separate for general AbstractArrays (incl. Overlays). Fix…

timholy:
…es #460 

CC @bjarthur. I can't easily test the PyPlot part because I don't have it installed, but this should at least fix `raw` and `separate` for overlays.


timholy:
2nd commit fixes https://github.com/timholy/Images.jl/issues/463#issuecomment-206397952, CC @vchuravy.

vchuravy:
Thank you Tim, I really enjoy working with `Images.jl` (and you other packages)

timholy:
Thanks, very kind of you!

PR #468, opened 2016-04-07
Run ImageMagick & QuartzImageIO tests on travis

timholy:
Since changes to Images have the possibility of breaking the IO packages (see https://github.com/JuliaIO/ImageMagick.jl/pull/34), we should run those packages' tests when testing Images.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,13 @@ julia:
     - nightly
 notifications:
     email: false
-script:
+before_script:
     - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
     - julia -e 'Pkg.clone(pwd()); Pkg.build("Images")'
+    - julia -e 'Pkg.add("ImageMagick"); Pkg.add("QuartzImageIO")'
+script:
     - julia -e 'Pkg.test("Images", coverage=true)'
+    - julia -e 'Pkg.test("ImageMagick", coverage=true); @osx_only Pkg.test("QuartzImageIO", coverage=true)'
 after_success:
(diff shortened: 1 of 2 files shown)


timholy:
Looks like this has inadvertently revealed ways in which QuartzImageIO lags behind ImageMagick. Perhaps more importantly, I'm surprised that the QuartzImageIO method is getting called by `ImageMagick.save`.

rsrock:
Looks like the test run on OS X that errored out with giant log is because the ImageMagick tests are being run using QuartzImageIO.  See these lines (around line 450, iirc)

```
INFO: Testing ImageMagick
IO
  > Binary png
WARNING: MethodError(QuartzImageIO.save,(FileIO.File{FileIO.DataFormat{:PNG}}
```

I do like the idea behind this, though.  I think it's a great idea to couple the tests.

timholy:
Right, but what has me puzzled is that the test [looks like this](https://github.com/JuliaIO/ImageMagick.jl/blob/360c947635655d4f375f2b5e74eecb90421e0662/test/constructed_images.jl#L14).

...and hovering doubtfully over the comment button can occasionally be useful, giving one a chance to think a little more. The answer: looks like we need to tag a new version of ImageMagick, that may solve the problem.

rsrock:
That is a curveball right there.  Hopefully the tagging fixes it!

rdeits:
Is this still something we want? I can fix the conflicts and get this going again.

timholy:
I'd say yes. Best to know if we do anything here that breaks the ability to load images. :+1:

johnnychen94:
I assume this and #667 are outdated now?

PR #469, opened 2016-04-08
Fixes for julia-0.5

timholy:
This eliminates a lot of warnings, as well as a few bugs. Likely to require https://github.com/JuliaLang/julia/pull/15810 (UPDATE: more importantly, the part moved to https://github.com/JuliaLang/julia/pull/15814), but let's try it anyway.


timholy:
The AppVeyor failure was some random webclient thing, and only affected one build.

Nice to see all this green!

PR #474, opened 2016-04-19
Switch to using Documenter.jl

rsrock:
This PR will switch us over to the new documentation system based on Documenter.jl.  I think I have managed to set up everything so that Travis will update the docs based on master after each commit, without any intervention from us.  That meant that I had to enter a Github token on Travis (as outlined in the Documenter docs), but it seems that I had permission to do that.

I was able to get this to work on my fork of Images.jl (with some modifications).  But the only way to tell if it will work here is to merge this and keep our fingers crossed.


rsrock:
@timholy, any objections to merging this? These changes have to end up on master, because that's where Travis looks when generating the docs.

timholy:
Let's go for it! Many thanks for tackling this!

rsrock:
Thanks!  I'm watching Travis churn right now.

...and it's borked, but I know what the issue is.  Give me a minute here.

rsrock:
And it works! 

http://timholy.github.io/Images.jl/latest/

However, I need to fix the link in README.md, it's pointing to my fork at the moment.

timholy:
Yay! I *love* having the images back.

I noticed the function reference has a bunch of `@` symbols, though.

PR #478, opened 2016-04-26
Support ScaleMinMax for To<:Integer

timholy:
Integer-valued images are not generally recommended (deciding contrast is a mess), but it's worth having them work. Note this also improves overflow-safety (for `From<:Signed`).


timholy:
Failure on 0.5 is https://github.com/JuliaLang/julia/pull/15675.

PR #480, opened 2016-05-06
Adds Histogram Function

mronian:
@timholy I found your version to be better than what I had earlier. Just one doubt : As you said, the two extra bins will create an issue for the histeq function. I could hard code that function to ignore those two bins but I dont know if that would be the correct way to do this. I do not want to remove the two a bins too since it gives a very nice representation of the image if (minval, maxval) does not cover the entire range. I feel we can do this : Take a (minval, maxval) range for the histeq function as well and then equalise only in that range. The default will be the entire range. This way we can have both the things. Does this look fine?

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1197,6 +1197,42 @@ if isdefined(:restrict)
 end
 
 """
+```
+range, histogram = imhist(img, nbins)
+range, histogram = imhist(img, nbins, minval, maxval)
+```
+
+Generates a histogram for the image over nbins spread between (minval, maxval). If minval and maxval are not given, then the
+minimum and maximum values present in the image are taken.
+"""
+
+imhist{T<:Colorant}(img::AbstractArray{T}, nbins=400) = imhist(convert(Array{Gray}, data(img)), nbins)
+imhist{T<:Colorant}(img::AbstractArray{T}, nbins, minval, maxval) = imhist(convert(Array{Gray}, data(img)), nbins, minval, maxval)
+
+function imhist{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins = 400)
+    minval = minfinite(img)
+    maxval = maxfinite(img)
+    imhist(img, nbins, minval, maxval)
+end
+
+function imhist{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins, minval::T, maxval::T)
+    bins = minval:(maxval-minval+1)/(nbins):maxval
+    histogram = zeros(Integer,nbins+2)
+    for val in img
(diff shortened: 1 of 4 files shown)


test/algorithms.jl
```diff
@@ -158,6 +158,10 @@ facts("Algorithms") do
         @fact img2 --> roughly(reinterpret(Float32, img))
     end
 
+    context("Exposure") do
+
```
timholy: ?

timholy:
Sounds quite reasonable to me. Looks like you might have missed some intended tests?

mronian:
Sorry about that. I've added the tests now.

I've made some changes to the function. Earlier it was taking the range ```minval:(maxval-minval)/(nbins-1):maxval```. While this worked well in some cases, it was giving the wrong intervals in cases like -

```
img=1:1:10
bins, hist=Images.imhist(img, 2)
```
Here, the bins would be ```1:9:10 --> 1,10``` instead of ```1,6 i.e. intervals of 1-5 and 6-10```. This would also not work in cases where only 1 bin was required (divide by zero error) or in cases where a minval and maxval value was supplied -

```
img=reshape(0:1:99,10,10)
bins, hist=Images.imhist(img, 7, 25, 59)
```

Here we need 7 bins from 25,30,35...55 and each should have a value=5, with the two overflow bins at both ends holding leftbin(<minval)=25 and rightbin(>maxval)=40. With the previous approach we would have got a range of 25:5.66:59 which would be wrong. 

I have changed the bins to be ```minval:(maxval-minval+1)/nbins:maxval```. I may be wrong in this though and may have misunderstood your approach.

timholy:
Yay! Thanks so much!

src/algorithms.jl
```diff
@@ -1197,6 +1197,42 @@ if isdefined(:restrict)
...
+
+function imhist{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins = 400)
+    minval = minfinite(img)
+    maxval = maxfinite(img)
+    imhist(img, nbins, minval, maxval)
+end
+
+function imhist{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins, minval::T, maxval::T)
+    bins = minval:(maxval-minval+1)/(nbins):maxval
+    histogram = zeros(Integer,nbins+2)
```
timholy: Oops, I didn't notice this, but it should be `Int` and not `Integer`. `Integer` is an abstract type, and that basically means that the operations in the loop can't be optimized. Containers should, whenever possible, have concrete element types.

Don't worry about fixing it, I'm tackling this as part of fixing #490.

PR #483, opened 2016-05-12
Fix depwarn on 0.5

yuyichao:
`ImageMagick.jl` PR coming soon. The only depwarn left comes from base https://github.com/JuliaLang/julia/pull/15675#issuecomment-214703355

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,3 +6,4 @@ SIUnits
 Zlib
 Graphics 0.1
 FileIO
+Compat 0.7.15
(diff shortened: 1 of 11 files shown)


timholy:
Bless you, @yuyichao. I'll wait for CI, but this looks great.

yuyichao:
There's some error on master that is not happening on my one week old build... I can have a look later.

timholy:
I'll fix it.

timholy:
Thanks!

Issue #484, opened 2016-05-14
Base.showcompact for properties


Evizero:
I wonder. would it maybe be more in the spirit of its intention to use `Base.showcompact` here: https://github.com/timholy/Images.jl/blob/master/src/core.jl#L1420 . 

I guess it is a weird kind of "problem" I am having, but I am currently working on something that stores metadata into images and I would like julia to display short and concise text for when an image is printed to the console like the following text shows (see `operations`, which is a vector denoting a history of performed operations):

```
RGBA Images.Image with:
  data: 64x64 Array{ColorTypes.RGBA{FixedPointNumbers.UFixed{UInt8,8}},2}
  properties:
    operations:  Flip x-axis. Flip y-axis. Crop to 1.0 aspect ratio. Resize to 64x64.
    imagedescription: <suppressed>
    spatialorder:  x y
    pixelspacing:  1 1
```

 while on the other hand I would like a more involved `show` for when I am just looking at an individual element (i.e. operation). Currently it seems that I could trick my way into that functionality by using `print` for the compact version and `show` for the more detailed one, but that seems a little weird to me given that `showcompact` exists.

It's not the most critical of issues, but I found it worth discussing. If you don't feel like that change is a good fit for any reason then I shall just work around it.

PR #485, opened 2016-05-14
Use showcompact instead of print for metadata

Evizero:
See #484 

basically, for vectors it still prints Strings using `print` while using `showcompact` for other types


Evizero:
I don't think the travis error on 0.5 is related to this

timholy:
Thanks!

PR #486, opened 2016-05-15
tweak Compat.String usage

StefanKarpinski:
There are some failures on Julia master but I think they're unrelated.


yuyichao:
This is API breaking.

yuyichao:
@StefanKarpinski Please stop changing all the string types on 0.4 and earlier versions unless you are sure no users can see the difference. I know you really want to get rid of the `UTF8String` and `ASCIIString` types but they are not going anywhere before 0.4 support is dropped.

PR #487, opened 2016-05-18
Updates for Documenter.jl syntax changes

MichaelHatherly:
See https://github.com/MichaelHatherly/Documenter.jl/pull/51 for details.

### DIFF
--- a/docs/src/function_reference.md
+++ b/docs/src/function_reference.md
@@ -4,9 +4,10 @@ Below, `[]` in an argument list means an optional argument.
 
 # Image construction
 
-    {docs}
-    Image
-    ImageCmap
+```@docs
+Image
+ImageCmap
+```
 
 ```
 convert(Image, A)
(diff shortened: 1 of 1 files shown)


timholy:
Wow, thanks so much!

MichaelHatherly:
Seems to be a bit of a rendering problem near the end of http://timholy.github.io/Images.jl/latest/function_reference/ from `imfilter` onward. That's due to some missing docs:

```
 !! No docs found for 'forwarddiffx'. [src/function_reference.md]
 !! No docs found for 'forwarddiffy'. [src/function_reference.md]
 !! No docs found for 'backdiffx'. [src/function_reference.md]
 !! No docs found for 'backdiffy'. [src/function_reference.md]
```

Those errors are also showing up in the `stable` docs but rendered a little better. Are those four functions meant to be defined somewhere?

MichaelHatherly:
https://github.com/MichaelHatherly/Documenter.jl/commit/f81f74ea005131b75a333956da563c0841f50b0b should fix the rendering problem here, just needs a new travis build to get triggered here and things should be back to how they looked before.

timholy:
I restarted the travis build. It's better in some ways, but that whole section now just renders as a list without providing info about the documented functions.

MichaelHatherly:
Yes, that's expected the behaviour when a `@docs` block contains functions with missing docstrings. It could just filter out missing ones and display those that do exist, though I kind of prefer it failing on the entire block which provides some motivation to fix the `!! No docs found` warnings.

timholy:
Fair enough :smile:.

PR #491, opened 2016-05-28
Fix histogram deprecation and behavior

### LINKED ISSUE #490: Error on imhist when img data is ColorTypes
Getting this error with both RGB and Gray ColorTypes :

```
using TestImages, Images

img=testimage("mandrill")
img2=testimage("lena_gray_256")

imhist(img2)

ERROR: ArgumentError: StepRange should not be used with floating point
 in steprange_last(::ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}}, ::Float64, ::ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}}) at ./range.jl:27
 in imhist(::Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}}, ::Int64, ::ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}}, ::ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}}) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1219
 in imhist(::Images.Image{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}},2}}) at /home/jarvis/.julia/v0.5/Images/src/algorithms.jl:1213
 in eval(::Module, ::Any) at ./boot.jl:225
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46
```

timholy:
This makes subtle changes in how the histogram is computed. The key change is to clarify the meaning of the return values (see the changes to the docstring for a description), which necessitated changing some of the tests. While I understand that some of the return values may be worse from a DWIM perspective, from a mathematical standpoint this is more consistent.

Incidentally this fixes #490, and also some other problems on julia-0.5. It won't pass tests, however, until https://github.com/JuliaStats/StatsBase.jl/pull/166 is merged and tagged.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -149,24 +149,25 @@ The base b of the logarithm (a.k.a. entropy unit) is one of the following:
   `:hartley` (log base 10)
 """
 function entropy(img::AbstractArray; kind=:shannon)
-  logᵦ = _log(kind)
+    logᵦ = _log(kind)
 
-  _, counts = hist(img[:], 256)
-  p = counts / length(img)
-  logp = logᵦ(p)
+    hist = StatsBase.fit(Histogram, vec(img), nbins=256)
+    counts = hist.weights
+    p = counts / length(img)
+    logp = logᵦ(p)
 
-  # take care of empty bins
-  logp[isinf(logp)] = 0
+    # take care of empty bins
+    logp[isinf(logp)] = 0
 
-  -sum(p.*logp)
+    -sum(p.*logp)
 end
 
 function entropy(img::AbstractArray{Bool}; kind=:shannon)
-  logᵦ = _log(kind)
+    logᵦ = _log(kind)
 
-  p = sum(img) / length(img)
+    p = sum(img) / length(img)
 
-  (0 < p < 1) ? - p*logᵦ(p) - (1-p)*logᵦ(1-p) : zero(p)
+    (0 < p < 1) ? - p*logᵦ(p) - (1-p)*logᵦ(1-p) : zero(p)
 end
 
 entropy{C<:AbstractGray}(img::AbstractArray{C}; kind=:shannon) = entropy(raw(img), kind=kind)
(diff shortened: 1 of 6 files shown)


timholy:
CC @mronian

src/algorithms.jl
```diff
@@ -1217,17 +1224,21 @@ function imhist{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins = 400)
...
-    bins = minval:(maxval-minval+1)/(nbins):maxval
-    histogram = zeros(Integer,nbins+2)
+    edges = StatsBase.histrange(Float64(minval), Float64(maxval), nbins, :left)
+    histogram = zeros(Int, length(edges)+1)
     for val in img
-        if val>maxval
+        if val<edges[1]
+            histogram[1] += 1
+            continue
+        end
```
mronian: I think this is handled by ```searchsortedlast(edges, val)``` itself, since it returns ```0``` if the value is out of the range covered by ```edges```.

timholy: Good point!

mronian:
Thanks @timholy

mronian:
Could you merge this to master?

timholy:
Now there's a new breakage on julia-0.5 because of the recent removal of the 0.4 deprecations for `float32` and `float64`. I think those are useful functions for image processing, so we'll need to fix this by not trying to import them on julia-0.5. However, I've got a couple of other things to tackle before I hop on a plane across the Pacific, so it may be a while before I get to that.

mronian:
No Problem! I'll test them on julia-0.4 till then. I'll try to submit a fix for the breakage if I can.

PR #492, opened 2016-06-06
Adds Gamma Correction Function

mronian:


src/algorithms.jl
```diff
@@ -1239,6 +1239,81 @@ end
...
+on the input type. If the input is an `Image` then the resulting image is of the same type
+and has the same properties. 
+
+For coloured images, the input is converted to YCbCr type and the Y channel is gamma corrected. 
+This is the combined with the Cb and Cr channels and the resulting image converted to the same 
+type as the input.
+
+"""
+
+RET_TYPE = Dict("8bit" => Gray{U8}, "16bit" => Gray{FixedPointNumbers.UFixed{UInt16,16}})
```
timholy: Any time you pass type information via a string, you're going to have type-instability. I let this slide on the histogram equalization PR, but we should get away from that pattern ASAP. It's much better to pass the type as a type.

src/algorithms.jl
```diff
@@ -1239,6 +1239,81 @@ end
...
+end
+
+function recompose_y(c::TransparentColor, eq_Y::Float32)
+    c_ycbcr = convert(YCbCrA, color(c))
+    ret_c = YCbCrA(eq_Y, c_ycbcr.cb, c_ycbcr.cr, c_ycbcr.alpha)
+    ret_c
+end
+
+function adjust_gamma{T<:Color}(img::AbstractArray{T}, gamma::Number, dtype = "8bit")
+    Y = [convert(YCbCr, color(c)).y for c in img]
```
timholy: What's the point of creating this temporary array? gamma adjustment is a pointwise operation, meaning that in the end it can be implemented as a 1-liner, `map(gammafunc, data(img))`.

src/algorithms.jl
```diff
@@ -1239,6 +1239,81 @@ end
...
+
+function adjust_gamma{T<:TransparentGray}(img::AbstractArray{T}, gamma::Number, dtype = "8bit")
+    opaque_img = [color(c) for c in img]
+    gamma_corrected_img = adjust_gamma(opaque_img, gamma, dtype)
+    gamma_corrected_img = reshape([convert(T, AGray(gamma_corrected_img[i], alpha(c))) for (i, c) in enumerate(img)], size(img))
+    gamma_corrected_img    
+end
+
+function adjust_gamma{T<:Number}(img::AbstractArray{T}, gamma::Number, minval::Number, maxval::Number, dtype = "8bit")
+    rescaled_img = convert(Array{RET_TYPE[dtype]}, [((x-minval) / (maxval-minval)) for x in img])
```
timholy: If you simply define a single function for *scalar* gamma correction, you don't need to mess with all this complexity. Something like
```jl
gamma_tform(x::Number, gamma, minval, maxval) = minval + (maxval-minval)*((x-minval)/(maxval-minval))^gamma
```
and then just `map` this over the image. Note this also avoids the type-instability of `dtype`.

For `UFixed` types, you're right that pre-computing a table might be faster. You'll need to allocate an array to define the table, but after creation you can again define the whole transformation on a scalar value as a one-line function.

bjarthur:
what are we going to do with `imstretch`, which performs "symmetric gamma-correction" ?

mronian:
@timholy I have made the function pixel-wise and the code has reduced by a lot (just a line for color images, the chain of functions is a bit long though). Like ```histeq```, the ```dtype``` causes the return type to be marked red with ```@code_warntype```. Also, ```map()``` seem to be marked red ( they return a Union, eg ```Union{Array{Float64,2},Array{Union{},2}}```). Is this an error?

src/algorithms.jl
```diff
@@ -1337,6 +1331,52 @@ histeq{T<:Number}(img::AbstractArray{T}, nbins) = histeq(img, nbins, minfinite(i
...
+
+For coloured images, the input is converted to YCbCr type and the Y channel is gamma corrected. 
+This is the combined with the Cb and Cr channels and the resulting image converted to the same 
+type as the input.
+
+"""
+function adjust_gamma{T<:FixedPointNumbers.UFixed, D<:Union{U8, U16}}(img::AbstractArray{Gray{T}}, gamma::Number, dtype::Type{D} = U8)
+    raw_type = FixedPointNumbers.rawtype(dtype)
+    gamma_inv = 1.0 / gamma
+    table = [dtype((i / typemax(raw_type)) ^ gamma_inv) for i in 0:typemax(raw_type)]
```
timholy: This line is one source of type instability. It can be replicated with
```jl
The type instability can be replicated with
```jl
function prep_table{D}(::Type{D}, gamma_inv)
    raw_type = FixedPointNumbers.rawtype(D)
    [D((i / typemax(raw_type)) ^ gamma_inv) for i in zero(raw_type):typemax(raw_type)]
end
```
But put a `D` in front of the comprehension, and suddenly inference is happier. Probably https://github.com/JuliaLang/julia/issues/15276.

src/algorithms.jl
```diff
@@ -1337,6 +1331,52 @@ histeq{T<:Number}(img::AbstractArray{T}, nbins) = histeq(img, nbins, minfinite(i
 
 histeq{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins, minval, maxval) = histeq(img, Int(nbins), convert(T, minval), convert(T, maxval))
 
+adjust_gamma{T<:Colorant, D<:Union{U8, U16}}(img::AbstractArray{T}, gamma::Number, dtype::Type{D} = U8) = map(c -> convert(T, recompose_y(convert(base_colorant_type(typeof(c)){dtype}, c), gamma_pixel_rescale(convert(YCbCr, color(convert(base_colorant_type(typeof(c)){dtype}, c))).y, gamma, Y_MIN, Y_MAX))), img)
+
+function adjust_gamma{D<:Union{U8, U16}}(img::AbstractImage, gamma::Number, dtype::Type{D} = U8)
+    gamma_corrected_img = adjust_gamma(data(img), gamma, dtype)
+    gamma_corrected_img = shareproperties(img, gamma_corrected_img)
```
timholy: This is another source of type-instability (change the var name).

src/algorithms.jl
```diff
@@ -1337,6 +1331,52 @@ histeq{T<:Number}(img::AbstractArray{T}, nbins) = histeq(img, nbins, minfinite(i
 
 histeq{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins, minval, maxval) = histeq(img, Int(nbins), convert(T, minval), convert(T, maxval))
 
+adjust_gamma{T<:Colorant, D<:Union{U8, U16}}(img::AbstractArray{T}, gamma::Number, dtype::Type{D} = U8) = map(c -> convert(T, recompose_y(convert(base_colorant_type(typeof(c)){dtype}, c), gamma_pixel_rescale(convert(YCbCr, color(convert(base_colorant_type(typeof(c)){dtype}, c))).y, gamma, Y_MIN, Y_MAX))), img)
```
timholy: You can break this line somewhere, or create an intermediate function.

timholy:
If you're interested in pursuing this further, it would be good to address @bjarthur's question too.

mronian:
@bjarthur We could have a function ```adjust_gamma``` which takes in a method for symmetric gamma correction as an argument. The ```imstretch``` function has matrix dot style operations and needs to be changed anyways to a pixel wise one. Would it be better to do it as a part of ```adjust_gamma``` or a separate function is better?

mronian:
Made a lot of changes, simplified ```adjust_gamma``` on the lines of #516 and tests for the same. Tests fail because ```isapprox``` for ```ARGB, AGray``` is not merged/tagged(JuliaGraphics/ColorVectorSpace.jl#47).

mronian:
There is still some type instability issue with ```map```-
```jl
julia> function foo(i)
           i*i
       end
foo (generic function with 1 method)

julia> img = reshape(1:1:100, 10, 10)
10×10 Base.ReshapedArray{Int64,2,StepRange{Int64,Int64},Tuple{}}:
  1  11  21  31  41  51  61  71  81   91
  2  12  22  32  42  52  62  72  82   92
  3  13  23  33  43  53  63  73  83   93
  4  14  24  34  44  54  64  74  84   94
  5  15  25  35  45  55  65  75  85   95
  6  16  26  36  46  56  66  76  86   96
  7  17  27  37  47  57  67  77  87   97
  8  18  28  38  48  58  68  78  88   98
  9  19  29  39  49  59  69  79  89   99
 10  20  30  40  50  60  70  80  90  100

julia> @code_warntype map(foo, img)
Variables:
  #self#::Base.#map
  f::#foo
  A::Base.ReshapedArray{Int64,2,StepRange{Int64,Int64},Tuple{}}

Body:
  begin  # abstractarray.jl, line 1478:
      return $(Expr(:invoke, _collect(::Base.ReshapedArray{Int64,2,StepRange{Int64,Int64},Tuple{}}, ::Base.Generator{Base.ReshapedArray{Int64,2,StepRange{Int64,Int64},Tuple{}},#foo}, ::Base.EltypeUnknown, ::Base.HasShape), :(Base._collect), :(A), :($(Expr(:new, Base.Generator{Base.ReshapedArray{Int64,2,StepRange{Int64,Int64},Tuple{}},#foo}, :(f), :(A)))), :($(Expr(:new, :(Base.EltypeUnknown)))), :($(Expr(:new, :(Base.HasShape))))))
  end::Union{Array{Int64,2},Array{Union{},2}}

```

timholy:
That's been one of the most long-standing problems in julia: the possibility that the input would be an **empty** array made type-stable `map` impossible, because `map` needed at least one element in order to compute the type of the output.

Having been frustrated by this issue for at least 2 years now (e.g., https://github.com/JuliaLang/julia/issues/8027), I'm very pleased to report that...if you update your julia, you'll see this has (very recently) been fixed! There were some deep-guts improvements that made it finally possible to use type-inference to reliably infer the output type of `foo` even if there were no actual elements to play with.

src/algorithms.jl
```diff
@@ -1530,6 +1530,51 @@ function histeq(img::AbstractImage, nbins::Integer, minval::Union{Number,Gray},
...
+gamma_corrected_img = adjust_gamma(img, gamma)
+```
+
+Returns a gamma corrected image. 
+
+The `adjust_gamma` function can handle a variety of input types. The returned image depends 
+on the input type. If the input is an `Image` then the resulting image is of the same type
+and has the same properties. 
+
+For coloured images, the input is converted to YCbCr type and the Y channel is gamma corrected.
```
timholy: `YIQ`

timholy:
Aside from this tiny edit, LGTM. Seems to need a rebase before merging, however.

mronian:
Done :) Tests fail due to ```isapprox```.

PR #494, opened 2016-06-08
Adds Histogram Matching

mronian:
Completed the Basic Function.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1432,9 +1432,9 @@ This assumes the input `img` has intensities between 0 and 1.
 imstretch(img::AbstractArray, m::Number, slope::Number) = _imstretch(float(img), m, slope)
 
 
-imhist{T<:Colorant}(img::AbstractArray{T}, nbins=400) = imhist(convert(Array{Gray}, data(img)), nbins)
+imhist{T<:Colorant}(img::AbstractArray{T}, nbins::Integer = 400) = imhist(convert(Array{Gray}, data(img)), nbins)
 
-function imhist{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins = 400)
+function imhist{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins::Integer = 400)
     minval = minfinite(img)
     maxval = maxfinite(img)
     imhist(img, nbins, minval, maxval)
@@ -1456,8 +1456,12 @@ maximum values present in the image are taken.
 `count[end]` is the number satisfying `x >= edges[end]`. Consequently,
 `length(count) == length(edges)+1`.
 """
-function imhist(img::AbstractArray, nbins, minval::Union{Gray,Real}, maxval::Union{Gray,Real})
+function imhist(img::AbstractArray, nbins::Integer, minval::Union{Gray,Real}, maxval::Union{Gray,Real})
     edges = StatsBase.histrange([Float64(minval), Float64(maxval)], nbins, :left)
+    imhist(img, edges)
+end
+
+function imhist(img::AbstractArray, edges::Range)
     histogram = zeros(Int, length(edges)+1)
     o = Base.Order.Forward
     G = graytype(eltype(img))
@@ -1472,7 +1476,6 @@ function imhist(img::AbstractArray, nbins, minval::Union{Gray,Real}, maxval::Uni
     end
     edges, histogram
 end
-
 
 function _histeq_pixel_rescale{T<:Union{Gray,Number}}(pixel::T, cdf, minval, maxval)
     n = length(cdf)
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1579,6 +1579,47 @@ end
 
 adjust_gamma{T<:Number}(img::AbstractArray{T}, gamma::Number, minval::Number, maxval::Number) = map(i -> _gamma_pixel_rescale(i, gamma, minval, maxval), img)
 
+"""
+```
+hist_matched_img = histmatch(img, oimg, nbins)
+```
+
+Returns a grayscale histogram matched image with a granularity of `nbins` number of bins. `img` is the image to be 
+matched and `oimg` is the image having the desired histogram to be matched to. 
+
+"""
+histmatch(img::AbstractImage, oimg::AbstractArray, nbins::Integer = 400) = shareproperties(img, histmatch(data(img), oimg, nbins))
+
+_hist_match_pixel{T<:Union{Gray, Number}}(pixel::T, bins, lookup_table) = T(bins[lookup_table[searchsortedlast(bins, pixel)]])
+
+function _hist_match_pixel{T<:Color}(pixel::T, bins, lookup_table)
+    yiq = convert(YIQ, pixel)
+    y = _hist_match_pixel(yiq.y, bins, lookup_table)
+    convert(T, YIQ(y, yiq.i, yiq.q))
+end
+
+_hist_match_pixel{T<:TransparentColor}(pixel::T, bins, lookup_table) = base_colorant_type(T)(_hist_match_pixel(color(pixel), bins, lookup_table), alpha(pixel))
+
+function histmatch{T<:Colorant}(img::AbstractArray{T}, oimg::AbstractArray, nbins::Integer = 400)
+    el_gray = graytype(eltype(img))
+    oedges, ohist = imhist(oimg, nbins, zero(el_gray), one(el_gray))
+    _histmatch(img, oedges, ohist)
+end
+
+function _histmatch(img::AbstractArray, oedges::Range, ohist::AbstractArray{Int})
+    bins, histogram = imhist(img, oedges)
+    ohist[1] = 0
+    ohist[end] = 0
+    histogram[1] = 0 
+    histogram[end] = 0
+    cdf = cumsum(histogram)
+    cdf /= cdf[end]
+    ocdf = cumsum(ohist)
+    ocdf /= ocdf[end]
+    lookup_table = [indmin(abs(ocdf-val)) for val in cdf]
+    map(i -> _hist_match_pixel(i, bins, lookup_table), img)
+end
+
 # image gradients
 
 # forward and backward differences
--- a/docs/src/function_reference.md
+++ b/docs/src/function_reference.md
@@ -285,6 +285,7 @@ findlocalminima
 imhist
 histeq
 adjust_gamma
+histmatch
 ```
 
 # Filtering kernels
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -227,6 +227,7 @@ export # types
     imhist,
     histeq,
     adjust_gamma,
+    histmatch,
     imlaplacian,
     imlineardiffusion,
     imlog,
@@ -292,7 +293,7 @@ Algorithms:
     - Resizing: `restrict`, `imresize` (not yet exported)
     - Filtering: `imfilter`, `imfilter_fft`, `imfilter_gaussian`, `imfilter_LoG`, `imROF`, `ncc`, `padarray`
     - Filtering kernels: `ando[345]`, `guassian2d`, `imaverage`, `imdog`, `imlaplacian`, `prewitt`, `sobel`
-    - Exposure : `imhist`, `histeq`, `adjust_gamma`
+    - Exposure : `imhist`, `histeq`, `adjust_gamma`, `histmatch`
     - Gradients: `backdiffx`, `backdiffy`, `forwarddiffx`, `forwarddiffy`, `imgradients`
     - Edge detection: `imedge`, `imgradients`, `thin_edges`, `magnitude`, `phase`, `magnitudephase`, `orientation`, `canny`
     - Corner detection: `imcorner`
--- a/test/algorithms.jl
+++ b/test/algorithms.jl
@@ -229,6 +229,11 @@ facts("Algorithms") do
         @fact img == ret --> true
         @fact eltype(ret) == eltype(img) --> true
 
+        img = ones(Images.RGB{Images.U16}, 10, 10)
+        ret = histeq(img, 100)
+        @fact img == ret --> true
+        @fact eltype(ret) == eltype(img) --> true
+
         img = ones(Images.RGB{Float64}, 10, 10)
         ret = histeq(img, 100)
         @fact all(map((i, r) -> isapprox(i, r), img, ret)) --> true
@@ -314,7 +319,6 @@ facts("Algorithms") do
         img = ones(Images.ARGB{Images.U8}, 10, 10)
         ret = adjust_gamma(img, 1)
         @fact img == ret --> true
-        @fact eltype(ret) == eltype(img) --> true
 
         #Working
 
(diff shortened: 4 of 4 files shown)


src/algorithms.jl
```diff
@@ -1239,6 +1239,86 @@ end
...
+# function histmatch{T<:TransparentGray}(img::AbstractArray{T}, args...)
+#     opaque_img = [color(c) for c in img]
+#     hist_matched_img = histmatch(opaque_img, args...)
+#     hist_matched_img = reshape([convert(T, AGray(hist_matched_img[i], alpha(c))) for (i, c) in enumerate(img)], size(img))
+#     hist_matched_img    
+# end
+
+# histmatch{T<:Gray}(img::AbstractArray{T}, nbins, dtype = "8bit") = histmatch(_prep_image_for_histmatch(img, dtype), nbins, ColorVectorSpace.typemin(RET_TYPE[dtype]), ColorVectorSpace.typemax(RET_TYPE[dtype]))
+
+function histmatch{T<:Union{Gray,Number}}(img::AbstractArray{T}, ohist::AbstractArray{Int}, dtype = "8bit")
```
timholy: Nothing seems to call this?

mronian: This was just the base function. I was waiting for #458 to be finalized so that the other functions can be directly ported here. Will add the others and commit again.

src/algorithms.jl
```diff
@@ -1333,10 +1333,74 @@ function histeq{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins::Int, minval
...
+function histmatch{T<:Colorant}(img::AbstractArray{T}, args...)
+    Y = map(c -> Gray(convert(YCbCr, color(c)).y - Y_MIN) / (Y_MAX - Y_MIN), img)
+    eq_Y = histmatch(Y, args...)
+    hist_matched_img = map((eq, c) -> convert(T, recompose_y(c, Y_MIN + (Y_MAX - Y_MIN) * eq)), eq_Y, img)
+    hist_matched_img
+end
+
+function histmatch(img::AbstractImage, oimg, nbins, dtype = U8)
+    hist_matched_img = histmatch(data(img), oimg, nbins, dtype)
+    hist_matched_img = shareproperties(img, hist_matched_img)
```
timholy: type unstable

src/algorithms.jl
```diff
@@ -1333,10 +1333,74 @@ function histeq{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins::Int, minval
...
+    hist_matched_img
+end
+
+function histmatch(img::AbstractImage, oimg, nbins, minval, maxval, dtype = U8)
+    hist_matched_img = histmatch(data(img), oimg, nbins, minval, maxval, dtype)
+    hist_matched_img = shareproperties(img, hist_matched_img)
```
timholy: here too

src/algorithms.jl
```diff
@@ -1333,10 +1333,74 @@ function histeq{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins::Int, minval
     hist_equalised_img
 end
 
-histeq{T<:Number}(img::AbstractArray{T}, nbins) = histeq(img, nbins, minfinite(img), maxfinite(img))
+histeq{T<:Number}+(img::AbstractArray{T}, nbins) = histeq(img, nbins, minfinite(img), maxfinite(img))
```
timholy: ?

src/algorithms.jl
```diff
@@ -1333,10 +1333,74 @@ function histeq{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins::Int, minval
...
+
+Returns a grayscale histogram matched image with a granularity of `nbins` number of bins. `img` is the image to be 
+matched and `oimg` is the image having the desired histogram to be matched to. If `minval` and `maxval` are specified then
+only the intensities in the range (minval, maxval) are matched. The `minval` and `maxval` need to be in the range (0, 1).
+Optionally, a return type may be specified using the `dtype` argument with the default being UFixed8.
+
+"""
+function histmatch{T<:Union{Gray,Number}, D<:Union{U8, U16}}(img::AbstractArray{T}, ohist::AbstractArray{Int}, args...)
+    hist_length = length(ohist)
+    bins, histogram = imhist(img, hist_length, ColorVectorSpace.typemin(T), ColorVectorSpace.typemax(T))
```
timholy: How do you know the bin edges will be of the same value?

mronian: We dont require that the bin edges be the same values. We just need the number of bins to be same. Thanks to the way Gray works, normalizing both the histograms (Lines 1392, 1394) is enough to do histogram matching. The only problem is ```StatsBase.histrange``` may not always return a range which has exactly the number of bins requested. I had not intended for this function to be used directly (with proper documentation about what types of histograms are acceptable, this could be done too). If a person uses the function above (supplying two images) then since for both Images StatsBase gets the same range of values of Gray, it returns the same number of bins (may not be exactly equal to ```nbins```).

src/algorithms.jl
mronian: Since @code_warntype did not raise an error I thought this was fine. Will change that. 

Should we have a shareproperties!() function? Also, I think you mentioned sometime earlier about doing away with the properties dictionary for ImagesCore.jl. Is that still the plan?

src/algorithms.jl
mronian: Stray error before committing. Updating.

src/algorithms.jl
timholy: We can switch over to our own implementation of choosing the edges, if it's easier for you. But I am wondering what happens if the two images don't have the same range. What if they are of the same scene but taken on different days with different cameras? What if one of the cameras has a few pixels obscured by a fleck of dust? Then it seems the range could be quite different. (If you save the edges and reuse them, then you can guarantee the histograms "mean" the same thing.)

src/algorithms.jl
timholy: Yes, for most uses. There will still be a version that has the dictionary, but I bet 95% of cases won't use it.

timholy: > Should we have a shareproperties!() function?

That won't help if you're trying to add properties to a type that doesn't have any.

> Also, I think you mentioned sometime earlier about doing away with the properties dictionary for ImagesCore.jl. Is that still the plan?

An array type with a dictionary will still be supported, but I suspect it will rarely be used.

src/algorithms.jl
mronian: By range, do you mean that one image is 8Bit and the other is say Float? That shouldnt matter since we are dividing both ranges by the same number of bins i.e. if we had scaled down the float image to an 8Bit image, then the bin number for any pixel will be the same since the range is also scaled down by that amount. Since we normalise both ranges to be between (0,1) before doing the matching, it effectively doesnt make a difference. Is this what you meant or have I gone off track?

mronian: Or do you mean that if one image has all pixels in range (200-255) and the other has all dark pixels (0-100), then the edges will not be correct since we take (minfinite, maxfinite) values as our range by default in ```imhist```? Since I am specifying the range to be typemin and typemax here, the range will be same for both (0-255). This will help to darken the bright image by matching its histogram (weighing to the right) with the histogram of the dark image(weighing to the left).

timholy: You're definitely on the right track, so great questions. My answers touch on several distinct points:

- Note that this is an exported function. Even though you're not documenting this signature, someone might call it directly (it's easily discoverable with `methods`). So you have to be a little bit concerned with "what happens if users call this in ways I wasn't intending?" If you want it to be strictly internal, then perhaps call it `_histmatch` and don't export it---of course people can still call it, but you've done your part to mark this as something internal and so at that point it's the user's problem, not yours.
- Along those lines, I noticed that this docstring talks about `oimg` but the method it's attached to uses `ohist`. You should attach the docstring to a method that matches the signature of what you're documenting.
- This method gets called by the one above that takes `oimg` and computes `ohist`. In that method, the call to `imhist`  *doesn't* specify the min/max values, so they will be chosen automatically. If `oimg` only fills part of the range of intensities permitted by the type, then if I'm reading it correctly, it will choose boundaries based on `minfinite/maxfinite` rather than `typemin/typemax`, and consequently choose different edges from what's chosen here (I could be misreading, of course). To me it seems your best bet is to pass in the bin boundaries that you used when computing `ohist`. Another benefit to doing this is to support cases where the input "image" is 1-dimensional and happens to be an array of `Int`---adding an argument will help distinguish an image from a histogram.
- As you indeed guessed...since this method supports `Number`, I'm also worried about what happens when someone passes in an `Array{Float64}`, since `typemin/typemax` are `+/-Inf`. Now that I look more carefully, I see your "higher level" methods support only `Colorant` eltypes, so now I see why you say this shouldn't be a problem. However, note you can call this method directly, so you don't know what the user will do. In general I've also tried to support plain number eltypes, since those are often used when working with grayscale, although for the new Images perhaps we should reconsider that decision. In any event, if here you just want to support `Colorant` I won't complain.

mronian: > To me it seems your best bet is to pass in the bin boundaries that you used when computing ohist. Another benefit to doing this is to support cases where the input "image" is 1-dimensional and happens to be an array of Int---adding an argument will help distinguish an image from a histogram.

I got it now. Thanks :)

> As you indeed guessed...since this method supports Number, I'm also worried about what happens when someone passes in an Array{Float64}, since typemin/typemax are +/-Inf. Now that I look more carefully, I see your "higher level" methods support only Colorant eltypes, so now I see why you say this shouldn't be a problem. However, note you can call this method directly, so you don't know what the user will do. In general I've also tried to support plain number eltypes, since those are often used when working with grayscale, although for the new Images perhaps we should reconsider that decision. In any event, if here you just want to support Colorant I won't complain.

I agree. We should probably choose Images to have all values between (0, 1). As seen in many cases this not only simplifies things but is also intuitive and easy to understand. Since easy conversion methods exist to go back to the plain number type (i.e 0-255 for 8Bit Gray), I am inclined to go with a (0, 1) range. With proper documentation and tutorials I think this design choice will make it easier for the user and contributors. (Linked to JuliaGraphics/ColorVectorSpace.jl#44)

timholy: Along those lines, in ImagesCore I've been adding new *view* types that let you think about the same underlying array different ways. There's a `ChannelView` which presents an m-by-n `AbstractArray{RGB{U8}}` as a 3-by-m-by-n `AbstractArray{U8}`, and then also a `rawview` that presents an `AbstractArray{U8}` as if it were an `AbstractArray{UInt8}`. Combine these two, and you can go back and forth between the RGB 0-1 world and the 0-255 world using a *single underlying array* (meaning, changes you make to the thing that acts like an `Array{UInt8}` will propagate back all the way to the original `Array{RGB{U8}}`).

mronian: Awesome!

mronian:
Made a lot of changes, simplified functions and added pixel wise operations. The same problem on ```@code_warntype``` with the ```map``` exists as in #492. Also, there is a test failure because of ```isapprox```. I think there is some bug - 

```jl
img = ones(Images.RGB{Float64}, 10, 10)
julia> ret = Images.histmatch(img, img)
10×10 Array{ColorTypes.RGB{Float64},2}:
 RGB{Float64}(0.0,3.59379e-17,0.0)  RGB{Float64}(0.0,3.59379e-17,0.0)  …  RGB{Float64}(0.0,3.59379e-17,0.0)
 RGB{Float64}(0.0,3.59379e-17,0.0)  RGB{Float64}(0.0,3.59379e-17,0.0)     RGB{Float64}(0.0,3.59379e-17,0.0)
 RGB{Float64}(0.0,3.59379e-17,0.0)  RGB{Float64}(0.0,3.59379e-17,0.0)     RGB{Float64}(0.0,3.59379e-17,0.0)
 RGB{Float64}(0.0,3.59379e-17,0.0)  RGB{Float64}(0.0,3.59379e-17,0.0)     RGB{Float64}(0.0,3.59379e-17,0.0)
 RGB{Float64}(0.0,3.59379e-17,0.0)  RGB{Float64}(0.0,3.59379e-17,0.0)     RGB{Float64}(0.0,3.59379e-17,0.0)
 RGB{Float64}(0.0,3.59379e-17,0.0)  RGB{Float64}(0.0,3.59379e-17,0.0)  …  RGB{Float64}(0.0,3.59379e-17,0.0)
 RGB{Float64}(0.0,3.59379e-17,0.0)  RGB{Float64}(0.0,3.59379e-17,0.0)     RGB{Float64}(0.0,3.59379e-17,0.0)
 RGB{Float64}(0.0,3.59379e-17,0.0)  RGB{Float64}(0.0,3.59379e-17,0.0)     RGB{Float64}(0.0,3.59379e-17,0.0)
 RGB{Float64}(0.0,3.59379e-17,0.0)  RGB{Float64}(0.0,3.59379e-17,0.0)     RGB{Float64}(0.0,3.59379e-17,0.0)
 RGB{Float64}(0.0,3.59379e-17,0.0)  RGB{Float64}(0.0,3.59379e-17,0.0)     RGB{Float64}(0.0,3.59379e-17,0.0)

julia> isapprox(ret[1], zero(Images.RGB{Float64}))
false

julia> a = ret[1].g
3.5937919307116316e-17

julia> isapprox(a, 0)
false

julia> a+0.1
0.10000000000000005

julia> isapprox(a, 0, rtol = 0.1)
false

```

timholy:
Testing "approximately equal to 0" is fraught with subtle traps. The usual meaning of "x is approximately equal to y" is "the difference between them is tiny compared to their magnitude." Since the magnitude of 0 is 0, it's not obvious how anything can be "approximately equal to 0." Note that `isapprox` does take an `atol` keyword argument; see the `isapprox` docs for an explanation.

With regards to the histmatch result, that does seem problematic. I'll see if I can spot the problem.

mronian:
Thanks! Corrected the error.

src/algorithms.jl
```diff
@@ -1456,8 +1456,12 @@ maximum values present in the image are taken.
 `count[end]` is the number satisfying `x >= edges[end]`. Consequently,
 `length(count) == length(edges)+1`.
 """
-function imhist(img::AbstractArray, nbins, minval::Union{Gray,Real}, maxval::Union{Gray,Real})
+function imhist(img::AbstractArray, nbins::Integer, minval::Union{Gray,Real}, maxval::Union{Gray,Real})
     edges = StatsBase.histrange([Float64(minval), Float64(maxval)], nbins, :left)
+    imhist(img, edges)
+end
+
+function imhist(img::AbstractArray, edges::Range)
```
timholy: :+1:

src/algorithms.jl
```diff
@@ -1531,6 +1534,54 @@ end
...
+function _hist_match_pixel{T<:Color}(pixel::T, args...)
+    yiq = convert(YIQ, pixel)
+    y = _hist_match_pixel(yiq.y, args...)
+    convert(T, YIQ(y, yiq.i, yiq.q))
+end
+
+_hist_match_pixel{T<:TransparentColor}(pixel::T, args...) = base_colorant_type(T)(_hist_match_pixel(color(pixel), args...), alpha(pixel))
+
+function histmatch{T<:Colorant}(img::AbstractArray{T}, oimg::AbstractArray, nbins::Integer = 400)
+    T = graytype(eltype(img))
```
timholy: This `T` will conflict with the type parameter `T`. I'm surprised julia didn't throw an error for this! (That seems like a julia bug to me.) Did it give you any kind of warning, either when building the package or when calling the function?

timholy: Ah, I get
```jl
julia> function foo{T}(x::T)
           T = Float32
           return T
       end
WARNING: local variable T conflicts with a static parameter in foo at REPL[1]:2.
foo (generic function with 1 method)
```
so julia seems fine. (If you see this warning, it means you need to fix something.)

src/algorithms.jl
```diff
@@ -1531,6 +1534,54 @@ end
...
+
+function _histmatch(img::AbstractArray, oedges::Range, ohist::AbstractArray{Int}, args...)
+    hist_length = length(ohist)
+    bins, histogram = imhist(img, oedges)
+    cdf = cumsum(histogram[2:end - 1])
+    cdf /= cdf[end]
+    ocdf = cumsum(ohist)
+    ocdf /= ocdf[end]
+    lookup_table = [indmin(abs(ocdf-val)) for val in cdf]
+    map(i -> _hist_match_pixel(i, bins, lookup_table, args...), img)
```
timholy: I suspect you'll get slow performance because of the splatting here. Do we have to have the version that takes `minval/maxval`? What if we don't trim off the edges of `histogram`, but set the edges to 0 before computing `cumsum`?

timholy: slow = absolutely god-awful, dreadful performance

mronian: Actually as we are moving to the (0,1) range for all Images, it is starting to lose sense. I'll remove that.

src/algorithms.jl
```diff
@@ -1531,6 +1534,49 @@ end
 
 histeq(img::AbstractImage, nbins::Integer) = shareproperties(img, histeq(data(img), nbins))
 
+"""
+```
+hist_matched_img = histmatch(img, oimg, nbins)
+hist_matched_img = histmatch(img, oimg, nbins, minval, maxval)
```
timholy: Need to update the docs too. (Not just here, also see the paragraph below.)

src/algorithms.jl
```diff
@@ -1531,6 +1534,47 @@ end
...
+
+function histmatch{T<:Colorant}(img::AbstractArray{T}, oimg::AbstractArray, nbins::Integer = 400)
+    el_gray = graytype(eltype(img))
+    oedges, ohist = imhist(oimg, nbins, zero(el_gray), one(el_gray))
+    _histmatch(img, oedges, ohist)
+end
+
+function _histmatch(img::AbstractArray, oedges::Range, ohist::AbstractArray{Int})
+    bins, histogram = imhist(img, oedges)
+    ohist[1] = 0
```
timholy: Not worth changing it, but for fun:
```jl
ohist[1] = ohist[end] = histogram[1] = histogram[end] = 0
```

timholy:
LGTM. I'll merge this once CI passes (or fails for known reasons).

mronian:
Awesome! Thanks :)

timholy:
Looks like there's an ambiguity warning on 0.4 (see the Travis logs). You could try `Base.Test.detect_ambiguities(Images, Base, Core)` on 0.5.

mronian:
@timholy I understand why there is an ambiguity but I am not sure how to go about resolving it. Could you help me out?

mronian:
I understood how to do it with help from the gitter channel and the docs. Its fixed now :)

timholy:
Thanks!

PR #498, opened 2016-06-12
Updates imcorner and makes it modular, adds new features.

mronian:
1. Support for Other ColorTypes
2. Thresholding
3. Kitchen Rosenfeld Corner Detection
4. Gamma Weighted Covariance Matrices

I had a discussion with @SimonDanisch who suggested using function names as arguments to decide which method to select. @timholy He asked me to take your advice before implementing this in other cases.

### DIFF
--- a/src/corner.jl
+++ b/src/corner.jl
@@ -1,38 +1,103 @@
 """
 ```
-imge = imcorner(img; [method], [border], [blockSize], [k])
+corners = imcorner(img, threshold, percentile; [method])
 ```
-Performs corner detection, using either the Harris method or the Shi-Tomasi
-method. `method` is `'harris'` (default) or `'shi-tomasi'`. `'border'` is the border 
-mode used in computing the gradient (default is `'replicate'`). `'blockSize'` is
-the size of the box filter kernel (default of `3`). `k` is the Harris free parameter 
-(default of `0.04`). It is only used when method is set to `'harris'`.
 
+Performs corner detection using one of the following methods - 
+
+    1. harris
+    2. shi_tomasi
+    3. kitchen_rosenfeld
+
+The parameters of the individual methods are described in their documentation. If
+a threshold is specified, the values of the responses are thresholded to give the 
+corner pixels. The threshold is assumed to be a percentile value unless `percentile`
+is set to false. 
 """
-function imcorner(img::AbstractArray; method::AbstractString="harris", border::AbstractString="replicate", blockSize::Int=3, k::Float64=0.04)
-
-    (grad_x,grad_y) = imgradients(img, "sobel",border);
-
-    cov_xx = grad_x .* grad_x;
-    cov_xy = grad_x .* grad_y;
-    cov_yy = grad_y .* grad_y;
-
-
-    box_filter_kernel = (1/(blockSize*blockSize)) * ones(blockSize,blockSize);
-
-    cov_xx = imfilter(cov_xx,box_filter_kernel);
-    cov_xy = imfilter(cov_xy,box_filter_kernel);
-    cov_yy = imfilter(cov_yy,box_filter_kernel);
-
-    if method == "harris"
-        corner = (cov_xx.*cov_yy - cov_xy.*cov_xy - k*(cov_xx + cov_yy).*(cov_xx + cov_yy));
-    elseif method == "shi-tomasi"
-        cov_yy = 0.5*cov_yy;
-        cov_xx = 0.5*cov_xx;
-        corner = ((cov_xx + cov_yy) - sqrt((cov_xx - cov_yy).*(cov_xx - cov_yy) + cov_xy.*cov_xy));
-    else
-        error("Unknown corner method: $method")
+function imcorner(img::AbstractArray, threshold = 0.99, percentile = true; method::Function = harris, args...)
+    img_gray = convert(Array{Gray}, img)
+    responses = method(img_gray; args...)
+    
+    if percentile == true
+        threshold = StatsBase.percentile(responses[:], threshold * 100)
     end
 
-    return corner
+    corners = map(i -> i < threshold ? zero(Gray{U8}) : one(Gray{U8}), responses)
+    corners
+end
+
+"""
+```
+harris_response = harris(img; [k], [border], [weights])
+```
+
+Performs Harris corner detection. The covariances can be taken using either a mean
+weighted filter or a gamma kernel.
+"""
+function harris(img::AbstractArray; k::Float64 = 0.04, args...)
+    cov_xx, cov_xy, cov_yy = gradcovs(img, args...)
+    corner = map((xx, yy, xy) -> xx * yy - xy ^ 2 - k * (xx + yy) ^ 2, cov_xx, cov_yy, cov_xy)
+    corner
+end
+
+"""
+```
+shi_tomasi_response = shi_tomasi(img; [border], [weights])
+```
+
+Performs Shi Tomasi corner detection. The covariances can be taken using either a mean
+weighted filter or a gamma kernel.
+"""
+function shi_tomasi(img::AbstractArray; border::AbstractString = "replicate", args...)
+    cov_xx, cov_xy, cov_yy = gradcovs(img, border; args...)
+    corner = map((xx, yy, xy) -> ((xx + yy) - (sqrt((xx - yy) ^ 2 + 4 * xy ^ 2))) / 2, cov_xx, cov_yy, cov_xy)
+    corner
+end
+
+"""
+```
+kitchen_rosenfeld_response = kitchen_rosenfeld(img; [border])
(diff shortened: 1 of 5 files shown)


SimonDanisch:
Note that keyword arguments aren't considered when dispatching, meaning that you miss out on static type inference... Work around:
http://docs.julialang.org/en/release-0.4/manual/performance-tips/#separate-kernel-functions

Also, callbacks are not fast on 0.4, but it seems like calls to these functions don't seem performance critical....

@timholy, what do you think about introducing a `RepeatedView` type, handling the borders `replicate`, `circular`, etc ?
This solution does look a bit antique after all the cool array changes: https://github.com/timholy/Images.jl/blob/master/src/algorithms.jl#L1172
Then we can use dispatch to skip a few operations for arrays, that are already wrapped in a View :)

src/corner.jl
```diff
@@ -1,38 +1,111 @@
...
+    2. Shi Tomasi
+    3. Kitchen Rosenfeld
+
+The parameters of the individual methods are described in their documentation. If
+a threshold is specified, the values of the responses are thresholded to give the 
+corner pixels. The threshold is assumed to be a percentile value unless `percentile`
+is set to false. 
 """
-function imcorner(img::AbstractArray; method::AbstractString="harris", border::AbstractString="replicate", blockSize::Int=3, k::Float64=0.04)
+function imcorner(img::AbstractArray, threshold = 0.99, percentile = true; method::Function = harris, args...)
```
kmsquire: You need to update `test/corner.jl` to pass in the `harris` function instead of "harris".

src/corner.jl
```diff
@@ -1,38 +1,111 @@
...
 ```
-Performs corner detection, using either the Harris method or the Shi-Tomasi
-method. `method` is `'harris'` (default) or `'shi-tomasi'`. `'border'` is the border 
-mode used in computing the gradient (default is `'replicate'`). `'blockSize'` is
-the size of the box filter kernel (default of `3`). `k` is the Harris free parameter 
-(default of `0.04`). It is only used when method is set to `'harris'`.
 
+Performs corner detection using one of the following methods - 
+
+    1. Harris
```
timholy: You might change these to be the same as the method you pass to `imcorner` as an argument.

src/corner.jl
```diff
@@ -1,38 +1,111 @@
...
+    
+    numerator = (grad_xx .* grad_y .^ 2 + grad_yy .* grad_x .^ 2 - 2 * grad_xy .* grad_x .* grad_y)
+    denominator = (grad_x .^ 2 + grad_y .^ 2)
+    non_zero = denominator .!= 0.0
+    corner = zeros(size(numerator))
+    corner[non_zero] = numerator[non_zero] ./ denominator[non_zero]
+    corner
 end
+
+function meancovs(img::AbstractArray, border::AbstractString = "replicate"; blockSize::Int = 3)
```
timholy: You could just have a `gradcovs` function and pass in either `img->imfilter(img, kernel)` or `img->imfilter_gaussian(img, [sigma,sigma])`. That would save almost all of the code duplication.

timholy:
I agree some of Images' interface is a little outdated, especially with julia-0.5 and the ability to pass functions as arguments with no performance penalty. I would write new code taking full advantage of the features of julia-0.5. If something depends on julia-0.5, we can even bump REQUIRE (I will do that soon anyway once https://github.com/JuliaIO/ImageMagick.jl/pull/41 becomes mergable...post JuliaCon).

mronian:
@timholy Please review the changes :)

src/corner.jl
```diff
@@ -1,38 +1,103 @@
 """
 ```
-imge = imcorner(img; [method], [border], [blockSize], [k])
+corners = imcorner(img, threshold, percentile; [method])
```
timholy: Since Julia 0.5 assigns a type to each function, the call to `method(img_gray; args...)` below might be inferable. However, type inference doesn't work with keyword arguments.

I don't know whether it really matters here, though, because the methods operate on the whole array, rather than pixel-by-pixel. Consequently the cost you pay for the type instability seems small. The only real problem would be if the return type, `corners`, is not inferred. Can you check?

src/corner.jl
```diff
@@ -1,38 +1,103 @@
...
+    denominator = map((x, y) -> x ^ 2 + y ^ 2, grad_x, grad_y)
+    corner = map((n, d) -> d == 0.0 ? 0.0 : n/d, numerator, denominator)
+    corner
+end
+
+function gradcovs(img::AbstractArray, border::AbstractString = "replicate"; weights::Function = meancovs, args...)
+    (grad_x, grad_y) = imgradients(img, "sobel", border)
 
-    return corner
+    cov_xx = map(*, grad_x, grad_x)
```
timholy: A little more readable as grad_x.*grad_x?

test/corner.jl
```diff
@@ -5,8 +5,15 @@ facts("Corner") do
...
 
 	context("Harris") do
-		Ac = Images.imcorner(A,method="harris")
-		# check corner intensity
+		Ac = Images.imcorner(A, method = harris)
+		# check corners
+		@fact Ac[16,16] --> 1.0
+		@fact Ac[16,26] --> 1.0
+		@fact Ac[26,16] --> 1.0
+		@fact Ac[26,26] --> 1.0
```
timholy: These tests would pass with an alternative implementation of `imcorner`:
```jl
imcorner(A; method=harris) = ones(size(A))
```
Seems much simpler. :smile:

mronian: Haha. I did not change these tests as they were already there. I'll add some better ones :stuck_out_tongue:

src/corner.jl
mronian: Well, it does give a flag on the result for type instability. I also did ```@code_warntype``` on  the individual response functions and it seems all of them give an unstable result because of the ```border``` argument (AbstractString) used for ```imgradients```. Also, ```imfilter``` and ```imfilter_gaussian``` were found to be type unstable earlier (dont remember which PR it was). 

What would be the best way to do this since ```method``` is proving to be type unstable?

test/corner.jl
mronian: Done.

src/corner.jl
timholy: Probably best is not to worry about it here, and fix it as part of the big revamp.

test/corner.jl
```diff
@@ -5,8 +5,21 @@ facts("Corner") do
...
+
+		@fact all(Ac[1:14, :] .== 0.0) --> true
+		@fact all(Ac[:, 1:14] .== 0.0) --> true
+		@fact all(Ac[27:end, :] .== 0.0) --> true
+		@fact all(Ac[:, 27:end] .== 0.0) --> true
+		@fact all(Ac[18:24, 18:24] .== 0.0) --> true
```
timholy: I like this one particularly, since it would defeat my next attempt at a "cheating" implementation: `copy(A)` :smile:. I think you've defeated me now :wink:.

timholy:
I really like this. I also like that we're finally green for most julia-0.5 tests now!

However, there's a failure on 32-bit, and interestingly it seems to be from this PR (see the Appveyor logs). I don't have the faintest clue why 32/64-bit would matter. You don't by any chance have access to a 32-bit machine, do you? Assuming the answer is "no" (I don't, either), one option would be to push some test commits that just `@show Ac[:,27:end]` and see if that gives you any clues. I'm sure this will be frustrating; sorry for that.

mronian:
The build is failing due to ```writemime``` :open_mouth:

mronian:
I commented out the tests for ```writemime.jl```. Turns out on 32 bit, the corner responses are a bit different(27th coloumn too has a higher response). I've updated the tests to reflect that now.

timholy:
Just ping me when all tests pass (or fail for irrelevant reasons, though I think we're back to passing all of our tests now so hopefully that won't happen).

mronian:
@timholy  The ```corner.jl``` tests do pass. The test failures are because of ```writemime``` which seems unrelated to this PR.

Its probably because of this latest commit on master - 3bc93af by @bjarthur.  If you remove the changes in that, then all tests pass (both in Travis and my local machine). I've fixed the test failures on 32 bit and all tests related to this PR are passing.

mronian:
Just wanted to make sure this time. @timholy All green flags are cleared this time :)

timholy:
Woot!

timholy:
I'd like to tag a new version of Images/ImageMagick soon, but do you want me to wait for your last PRs? (I can of course retag.)

mronian:
I don't mind but I am almost done with FAST corners and the CLAHE algorithm (coming up). It would be great to have all the exposure correction functions in the next release :)

timholy:
Sounds very much worth waiting for :smile:.

PR #508, opened 2016-07-02
Adds Integral Image. Required for CENSURE in ImageFeatures

mronian:


mronian:
Corrected Build error.

src/algorithms.jl
```diff
@@ -1955,3 +1955,20 @@ function shepp_logan(M,N; highContrast=true)
...
+```
+
+Returns the integral image of an image. The integral image is calculated by assigning
+to each pixel the sum of all pixels above it and to its left, i.e. the rectangle from 
+(1, 1) to the pixel. 
+"""
+function integral_image(img::AbstractArray)
+    integral_img = convert(Array{UInt}, raw(img))
+    for i in 1:ndims(img)
+        integral_img = cumsum(integral_img, i)
```
timholy: If `img` is a color image, then `raw` will expand the color channel so it's the first dimension, and this will be an error.

I'd prefer to write it something like this:
```jl
integral_img = Array{accum(eltype(img))}(size(img))
sd = coords_spatial(img)
cumsum!(integral_img, img, sd[1])
for i = 2:length(sd)
    cumsum!(integral_img, integral_img, sd[i])
end
integral_img
```

mronian: I could not find docs for the ```accum``` method. What does that do?

timholy: It basically picks an "accumulation type"; as you noticed, if the underlying type is `U8` (or `UInt8`), overflow will occur. This is an internal method; we could document it, of course.

```sh
tim@diva:~/.julia/v0.5/Images/src$ grep accum *
algorithms.jl:            # When we are reducing along dim == 1, we can accumulate to a temporary
algorithms.jl:            s = zero(accum(eltype(T)))
algorithms.jl:accum{T<:Integer}(::Type{T}) = Int
algorithms.jl:accum(::Type{Float32})    = Float32
algorithms.jl:accum{T<:Real}(::Type{T}) = Float64
```
We'd need to generalize it for colors, of course. Probably
```jl
accum{C<:Colorant}(::Type{C}) = base_colorant_type(C){accum(eltype(C))}
```
would suffice.

mronian:
I've made the changes.

timholy:
Thanks!

PR #509, opened 2016-07-02
Fixes Code Warntype Errors, adds pixel wise operations in Histeq

mronian:


src/algorithms.jl
```diff
@@ -1265,7 +1265,7 @@ If minval and maxval are specified then only the intensities in the range
 Y_MIN = 16
 Y_MAX = 235
 
-function _prep_image_for_histeq(img::AbstractArray, dtype)
+function _prep_image_for_histeq{D<:Union{U8, U16}}(img::AbstractArray, dtype::Type{D})
     img_shape = size(img)
     img = convert(Array{base_colorant_type(eltype(img)){dtype}}, img)
```
mronian: I guess these type warnings cannot be resolved right?

src/algorithms.jl
```diff
@@ -1312,21 +1312,24 @@ end
 function histeq{T<:TransparentGray}(img::AbstractArray{T}, args...)
     opaque_img = convert(Array{base_color_type(eltype(img))}, img)
```
mronian: Also, this one?

src/algorithms.jl
timholy: I'll see what I can do.

timholy:
Very nice!

PR #512, opened 2016-07-02
Adds Tests for Histeq

mronian:
Build fails on 0.4 because zero(YCbCr) is not found. Is ColorTypes not available for 0.4?


timholy:
I tagged a new ColorTypes, which hopefully fixes all the problems with this.

mronian:
Updated the tests. All of them pass now except the one with ```isapprox``` on Travis (JuliaGraphics/ColorVectorSpace.jl#45).

timholy:
Great!

PR #514, opened 2016-07-09
Adds Bilinear Interpolation (Required for LBP Feature)

mronian:


src/algorithms.jl
```diff
@@ -1958,3 +1958,42 @@ function shepp_logan(M,N; highContrast=true)
...
+
+The interpolation is done in one direction first and then the values obtained
+are used to do the interpolation in the second direction.
+
+"""
+function bilinear_interpolation{T}(img::AbstractArray{T, 2}, row::Number, col::Number)
+    y_min = Int(floor(row))
+    x_min = Int(floor(col))
+    y_max = Int(ceil(row))
+    x_max = Int(ceil(col))
```
kmsquire: Probably doesn't matter much, but I think it reads a little nicer if you use, e.g., `floor(Int, row)` and `ceil(Int, col)` instead.

src/algorithms.jl
```diff
@@ -1958,3 +1958,42 @@ function shepp_logan(M,N; highContrast=true)
...
+```
+
+Bilinear Interpolation is used to interpolate functions of two variables
+on a rectilinear 2D grid.
+
+The interpolation is done in one direction first and then the values obtained
+are used to do the interpolation in the second direction.
+
+"""
+function bilinear_interpolation{T}(img::AbstractArray{T, 2}, row::Number, col::Number)
```
kmsquire: The names `row` and `col` suggest a position in a matrix, and it seems like they should be integers.  Consider using `y` and `x` instead.

mronian:
@kmsquire I've made the changes :)

timholy:
I'm fine with adding this, but in the not-too-distant future we should make Images use Interpolations.jl.

mronian:
Shall I use Interpolations.jl instead? The current implementation of LBP ( JuliaImages/ImageFeatures.jl#3 ) uses this.

timholy:
I'm fine with this, if it speeds your work. (And as long as you're aware it might change in the future.)

timholy:
(Note it needs rebasing before it's possible to merge.)

timholy:
Test failures on 0.4.

mronian:
Yep. I was finishing up exposure correction before resuming ImageFeatures. Will look into this.

mronian:
@timholy Tests pass now.

timholy:
Awesome. Time for a new release of Images?

mronian:
Just #523 left. We could tag it later too though, not a problem.

src/algorithms.jl
```diff
@@ -2047,3 +2047,42 @@ function integral_image(img::AbstractArray)
...
+function bilinear_interpolation{T}(img::AbstractArray{T, 2}, y::Number, x::Number)
+    y_min = floor(Int, y)
+    x_min = floor(Int, x)
+    y_max = ceil(Int, y)
+    x_max = ceil(Int, x)
+
+    topleft = try checkbounds(img, y_min, x_min); img[y_min, x_min] catch; zero(T) end
+    bottomleft = try checkbounds(img, y_max, x_min); img[y_max, x_min] catch; zero(T) end
+    topright = try checkbounds(img, y_min, x_max); img[y_min, x_max] catch; zero(T) end
+    bottomright = try checkbounds(img, y_max, x_max); img[y_max, x_max] catch; zero(T) end
```
timholy: I did a little playing with `lbp`, and from profiling these 4 lines account for about 80% of the runtime. Try changing these to `checkbounds(Bool, img, ...)`.

(Sorry I didn't notice earlier.)

mronian: I had noticed this earlier. In fact the original version of this PR had the ```Bool``` version, but it lead to test failures on julia-0.4 since it does not have the ```checkbound(Bool, Array...)``` function. I looked around and found a ```checkbounds(Bool, dimlength....)``` version though for 0.4 and will change this to that.

mronian: @timholy I tried changing the lines but then it gives deprecation warnings and suggests using the ```checkbounds(Bool, Array...)``` version instead. Should we use the old API or bump this to 0.5?

timholy: You could use a pattern like this:
```jl
topleft = chkbounds(Bool, img, y_min, x_min)
```
where `chkbounds` is a function you define yourself:
```jl
if VERSION < v"0.5.0-dev"
   function chkbounds(::Type{Bool}, img, x, y)
       try
            checkbounds(img, x, y)
            return true
       catch
            return false
       end
    end
else
    chkbounds(::Type{Bool}, img, x, y) = checkbounds(Bool, img, x, y)
end
```
(If you can't find a better version than the `try/catch` version.)

PR #515, opened 2016-07-15
Fix for sub/slice->view

timholy:
Hopefully this will fix our tests (and Travis).

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -18,6 +18,8 @@ else
     export float32, float64
 end
 
+import Compat.view
+
 # "deprecated imports" are below
 
 using Colors, ColorVectorSpace, FixedPointNumbers, FileIO, StatsBase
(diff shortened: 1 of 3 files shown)


bjarthur:
looks great.

what else needs to be done for julia 0.5?  i'm happy to help.

timholy:
Thanks @bjarthur! I should resurrect the writemime deprecation PR and see if it works, but I'll do that after this merges. Would you perhaps check ImageMagick and see what needs to be done there? I'd like to make sure the tests pass and there aren't warnings, either when the package is built or during tests.

For Julia 0.5, I'm actually rewriting Images :smile:. Not very quickly because I've been consumed with Julia 0.5 contributions, but I just pushed [this](https://github.com/JuliaLang/julia/pull/17228) so perhaps I'll soon be free of my burdens :wink:.

timholy:
Argh, by pulling in `StatsBase` (for the histogram functions) we've become dependent on Rmath, which is broken on Windows. Still, most tests pass, which is an improvement!

timholy:
Oh, and @bjarthur, if you're taking requests...there's a warning thrown from Base.FFTW about `copy(::Tuple)`; it comes up in the Images "fft and ifft" tests. Almost surely leftover cruft from https://github.com/JuliaLang/julia/pull/15675. Since you've now made numerous base julia contributions, would you be willing to tackle that?

PR #516, opened 2016-07-16
A few changes to histogram functionality

timholy:
@mronian, I was looking at the histogram functionality and noticed a few changes that seemed to make things simpler and faster. It's probably better if you review each commit separately; the first one just moves some code around, which makes it hard to see what I actually changed.

~~This is WIP because right now this won't pass the tests in #512 due to some problem in ColorTypes that I don't have time to track down right now, but if I comment out one line in that package then a suitable version of your tests do pass. So no one should merge this until I have time (not today) to fix the other problem.~~ (Fixed in https://github.com/JuliaGraphics/ColorTypes.jl/pull/41, though we still need to tag a new version before tests will pass.)


src/algorithms.jl
```diff
@@ -437,6 +437,9 @@ accum(::Type{Float32})    = Float32
 accum{T<:Real}(::Type{T}) = Float64
 accum{C<:Colorant}(::Type{C}) = base_colorant_type(C){accum(eltype(C))}
 
+graytype{T<:Number}(::Type{T}) = T
+graytype{C<:AbstractGray}(::Type{C}) = C
+graytype{C<:Colorant}(::Type{C}) = Gray{eltype(C)}
```
mronian: This was really needed :)

mronian:
If you could merge this, I will change the tests in #512 to test the new functions.

src/algorithms.jl
```diff
@@ -1552,6 +1431,99 @@ This assumes the input `img` has intensities between 0 and 1.
...
+
+function _histeq_pixel_rescale{T<:Union{Gray,Number}}(pixel::T, cdf, minval, maxval)
+    n = length(cdf)
+    bin_pixel = clamp(ceil(Int, (pixel-minval)*length(cdf)/(maxval-minval)), 1, n)
+    rescaled_pixel = minval + ((cdf[bin_pixel]-cdf[1])*(maxval-minval)/(cdf[end]-cdf[1]))
+    convert(T, rescaled_pixel)
+end
+function _histeq_pixel_rescale{C<:Color}(pixel::C, cdf, minval, maxval)
+    yiq = convert(YIQ, pixel)
+    y = _histeq_pixel_rescale(pixel_yiq.y, cdf, minval, maxval)
```
mronian: In the case of ```Color```, in case minval and maxval are not provided, you will need to take the min and max possible values of the Y channel in YIQ.

src/algorithms.jl
```diff
@@ -1552,6 +1431,99 @@ This assumes the input `img` has intensities between 0 and 1.
...
+    cdf = cumsum(histogram[2:end-1])
+    img_shape = size(img)
+    minval == maxval && return map(identity, img)
+    hist_equalised_img = map(p -> _histeq_pixel_rescale(p, cdf, minval, maxval), img)
+    hist_equalised_img
+end
+
+histeq{T<:Union{Gray,Number}}(img::AbstractArray{T}, nbins) = histeq(img, nbins, minfinite(img), maxfinite(img))
+
+histeq(img::AbstractArray, nbins) = histeq(img, nbins, minfinite(Gray, img), maxfinite(Gray, img))
```
mronian: I guess this attempts to spread the pixels between the range (minval, maxval) of the current Image. In my opinion, histogram equalisation should try to equalise the image histogram over the entire range possible which is why the whole issue of providing a datatype argument had come up (most libraries prove only 8 bit though). 

For example - 

Consider an 8 Bit image. If it is over exposed i.e. all values are between 200 and 255, the equalisation will happen only in this range and the result will again be an over exposed image. Instead if we equalise it over 0-255 (typemin, typemax) then the result will be much better and the exposure will be corrected. The issue of datatypes came because ```Gray{Float64}``` has a extremes which are too huge and so we scaled down to an 8 Bit / 16 Bit Image.

timholy: Ah, that's interesting, very good point. `typemin` and `typemax` are a little problematic:
```jl
julia> typemin(Gray{Float64})
ERROR: MethodError: no method matching typemin(::Type{ColorTypes.Gray{Float64}})
Closest candidates are:
  typemin(::Type{Bool})
  typemin(::Type{Int8})
  typemin(::Type{UInt8})
  ...
 in eval(::Module, ::Any) at ./boot.jl:234
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46
```
and it's not obvious whether the answer should be `Gray(-Inf)` or `Gray(0.0)`. In general I dislike the idea of forcing the output to be `U8` or `U16`, since some image processing will be using floats and you might as well stick with them. What if we default to `zero(eltype(img))` and `one(eltype(img))`?

src/algorithms.jl
timholy: Now there are no callers of this that do not supply `minval` and `maxval`.

timholy:
Test failures are independent of this PR (in julia itself or in other packages).

mronian:
Just referencing this issue here JuliaGraphics/ColorVectorSpace.jl#44 so that we fix this once the Gray{Float64} is handled.

src/algorithms.jl
```diff
@@ -1552,6 +1431,105 @@ This assumes the input `img` has intensities between 0 and 1.
...
+For coloured images, the input is converted to YCbCr type and the Y channel is equalised. This
+is the combined with the Cb and Cr channels and the resulting image converted to the same type
+as the input.
+
+If minval and maxval are specified then intensities are equalized to the range
+(minval, maxval). The default values are 0 and 1.
+
+"""
+function histeq(img::AbstractArray, nbins::Integer, minval::Union{Number,Gray}, maxval::Union{Number,Gray})
+    bins, histogram = imhist(img, nbins, minval, maxval)
```
mronian: @timholy I was doing the CLAHE implementation and realised that in case of color images, the histogram calculated will be of a converted ```Gray``` image but the values considered while scaling would be the Y channel of ```YIQ``` colorspace. We should either abandon the ```YIQ``` conversion and use grayscale itself, or we should use something like my previous implementation of ```histeq``` which would pass the entire Y channel for histogram calculation.

timholy:
```jl
julia> c = rand(RGB{U8})
RGB{U8}(0.533,0.839,0.298)

julia> Gray(c)
Gray{U8}(0.686)

julia> YIQ(c)
YIQ{Float32}(0.68606275f0,-0.0083590355f0,-0.2330596f0)
```
That's why I switched to YIQ, the Y component is identical to `Gray` up to the `Float32`/`U8` distinction. (Do you think that matters?)

mronian: Oh. If that it is identical then its fine. I was worried it was taking two different spaces and using them to calculate the rescaled pixel. Thanks!

timholy: I'm glad you're paying such close attention, that's great!

PR #520, opened 2016-07-23
Fix Test Failures on writemime

mronian:
Fixes test failures for 433cb650516dd4d57807a307c561e6f1b2c6bcff

### DIFF
--- a/test/writemime.jl
+++ b/test/writemime.jl
@@ -4,7 +4,7 @@ facts("show (MIME)") do
         mkdir(workdir)
     end
     context("no compression or expansion") do
-        A = U8[0.01 0.99; 0.25 0.75]
+        A = Gray{U8}[0.01 0.99; 0.25 0.75]
         fn = joinpath(workdir, "writemime.png")
         open(fn, "w") do file
             @compat show(file, MIME("image/png"), grayim(A), minpixels=0, maxpixels=typemax(Int))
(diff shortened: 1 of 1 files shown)


mronian:
There are failures on julia-0.4. @timholy What needs to be added?

bjarthur:
ooops, my bad.  i should've run the tests before pushing that change.

i specified an eltype there to get ImageMagick/test/readremote.jl line 116 to succeed.  on that line, `img` is of type `AbstractImage{Colorant}` and was dispatched to the version of `show` on line 36 of Images/src/writemime.jl which inputs `AbstractArray{Colorant}`, the supertype of `AbstractImage{Colorant}`, because the show method on the then line 12 input `AbstractImage` (without the eltype specified).  line 37 then proceeds to pass that `AbstractImage{Colorant}` into `Image` which screwed things up.

would it be better to change the tests as @mronian has done here, or to define show methods for *both* `AbstractImage{Colorant}` and `AbstractImage` ?

mronian:
I think the second way is better? That way it will pass on 0.4 too.

timholy:
Best to do such things as a pull request too; I can no longer count the number of times I've had something pass locally for me, only to discover that I was depending on having checked out the master branch of some package. If other issues didn't cause problems, that might even be relevant here, see https://github.com/JuliaLang/Compat.jl/pull/256 (which has not yet been tagged).

timholy:
Superseded by #521, but thanks so much @mronian for trying to tackle this.

PR #521, opened 2016-07-23
Some fixes for writemime

timholy:
I'm not certain this will pass CI because it's a simultaneous change to both this package and ImageMagick, but locally the two together pass on both 0.4 and 0.5. CC @mronian, @bjarthur


bjarthur:
ah!  you beat me to it.  i was working on a PR...

timholy:
Sorry, and thanks for plunging in!!

PR #524, opened 2016-07-24
Separates all exposure correction functions and their respective tests to new files

mronian:
Also modifies ```imcomplement``` and ```imstretch```, adds tests.


timholy:
Very good idea.

PR #523, opened 2016-07-24
Adds Contrast Limited Adaptive Histogram Equalisation(CLAHE)

mronian:
Needs #514.

How should we handle colour images? [Refer https://github.com/timholy/Images.jl/pull/516#discussion_r71996858]

PR #528, opened 2016-07-26
Faster histeq

timholy:
Ref https://github.com/timholy/Images.jl/pull/523#issuecomment-235236460, CC @mronian.

### DIFF
--- a/src/exposure.jl
+++ b/src/exposure.jl
@@ -112,7 +112,11 @@ function histeq(img::AbstractArray, nbins::Integer, minval::Union{Number,Gray}, 
     cdf = cumsum(histogram[2:end-1])
     img_shape = size(img)
     minval == maxval && return map(identity, img)
-    hist_equalised_img = map(p -> _histeq_pixel_rescale(p, cdf, minval, maxval), img)
+    # Would like to use `map` here, but see https://github.com/timholy/Images.jl/pull/523#issuecomment-235236460
+    hist_equalised_img = similar(img)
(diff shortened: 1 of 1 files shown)


timholy:
Since I had this already I thought it would be best to just submit it. But a little bit of poking around suggests that `histmatch` may have the same performance problem, so a good exercise for you :smile: will be to check that it's indeed a problem and fix it.

mronian:
@timholy I tried the same with ```histmatch``` and ```adjust_gamma``` but there seems to be no increase in speed. ```ProfileVIew.view(C=true)``` does not have the internal C function calls or ```typemap``` at the top of the stack like in ```histeq```. Still, should we avoid map just to be safe?

mronian:
Turns out not only ```map``` but any type of ```[do this for i in I]``` type of statement causes speed issues. I changed - 
```jl
table = [T((i / typemax(raw_type)) ^ gamma_inv) for i in zero(raw_type):typemax(raw_type)]
```
to - 
```jl
table = zeros(Float64, typemax(raw_type) + 1)
    for i in zero(raw_type):typemax(raw_type)
        table[i + 1] = T((i / typemax(raw_type)) ^ gamma_inv)
    end
```
in ```adjust_gamma``` and the speed up is incredible.

timholy:
Great, glad to hear it! (Or in a sense, sorry to hear it---would be nice if the original worked just as well. But worthwhile improvements for users.)

PR #529, opened 2016-07-26
Faster histmatch, adjust_gamma

mronian:


src/exposure.jl
```diff
@@ -166,16 +166,38 @@ This is the combined with the I and Q channels and the resulting image converted
...
+    for I in eachindex(img)
+        gamma_corrected_img[I] = Gray(table[convert(base_colorant_type(typeof(img[I])){T}, img[I]).val.i + 1])
+    end
+    gamma_corrected_img
 end
 
-adjust_gamma{T<:Number}(img::AbstractArray{T}, gamma::Number, minval::Number, maxval::Number) = map(i -> _gamma_pixel_rescale(i, gamma, minval, maxval), img)
+adjust_gamma{T<:Number}(img::AbstractArray{T}, gamma::Number) = _adjust_gamma(img, gamma, Float64)
+adjust_gamma{T<:Colorant}(img::AbstractArray{T}, gamma::Number) = _adjust_gamma(img, gamma, T)
+
```
mronian: This was required since type needs to be specified for the return image. ```similar(img)``` will give an ```Int``` image if ```T<:Int```.

timholy:
Great, thanks!

PR #531, opened 2016-07-28
Adds Integral Window Sum

mronian:
Makes things easier in ImageFeatures.jl.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1813,7 +1813,9 @@ integral_img = integral_image(img)
 
 Returns the integral image of an image. The integral image is calculated by assigning
 to each pixel the sum of all pixels above it and to its left, i.e. the rectangle from
-(1, 1) to the pixel.
+(1, 1) to the pixel. An integral image is a data structure which helps in efficient 
+calculation of sum of pixels in a rectangular subset of an image. See `boxdiff` for more
+information.
 """
 function integral_image(img::AbstractArray)
     integral_img = Array{accum(eltype(img))}(size(img))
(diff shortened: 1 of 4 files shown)


timholy:
I understand the reason you want this function, but I'm not so sure about the API. The part that makes it `integral` is what you pass into it; the function itself is just computing a couple of differences. Also, any time you specify a region with 4 numbers, it gets hard to remember the order of arguments; if nothing else it would be better to use `UnitRange`s.

On balance, I wonder if this is better used as a private (non-exported) function in ImageFeatures?

mronian:
No problem. I wanted to put this here since it might help in future algorithms as many algorithms use integral images.

timholy:
It's fine to have it here, but I think we'd need a better API, so if you can come up with something...

mronian:
By API, you mean the 4 coordinates right (I'll change it to ```UnitRange``` and one more with ```CartesianIndex``` so both can be used)? Is the function name fine?

mronian:
Would it be ok to tag a new release of ```Images.jl``` once this PR is done? It will help me avoid changing the travis script of ```ImageFeatures.jl```. Is it too early for a new release?

mronian:
@timholy  Is the modified API good ?

timholy:
It addresses the second comment but not the first comment in https://github.com/timholy/Images.jl/pull/531#issuecomment-236000983.

timholy:
And no, it would be fine to tag a new version again. It was too long between releases before.

mronian:
I am a bit confused :sweat_smile: ....would you like the function to calculate the integral image inside it or do you want the name to be changed so it doesn't give the wrong message that it calculates the integral image in itself.

mronian:
Also, just wanted to check with you if you receive notifications when I comment on any issue/PR in ImageFeatures.jl.

timholy:
I'm not sure, but something isn't right. It doesn't make any sense to recalculate the integral each time, so you want to cache the integral.

I think the biggest point is that the function name doesn't at all match what the function is doing: it's actually computing a `diff`, not an integral.

timholy:
> Also, just wanted to check with you if you receive notifications when I comment on any issue/PR in ImageFeatures.jl.

I do. I am trying to *stop* coding for the day and do other stuff, so it will be tomorrow at the earliest. (Sorry, I have a couple of deadlines.)

mronian:
Oh Great! No problem.

PS - I'll look for a better name for that function.

mronian:
Sorry forgot about this. I've changed it to ```diff_of_integral_image```. Does that seem fine?

mronian:
@timholy The last build of AppVeyor seems to have stopped because of an unrelated error.

timholy:
How about just `boxdiff` or `diff_box` or something? The "integral" part is solely a consequence of which argument you pass, not the actual behavior of the function.

src/algorithms.jl
```diff
@@ -1827,6 +1827,27 @@ end
 
 """
 ```
+sum = boxdiff(integral_image, ytop:ybot, xtop:xbot)
+sum = boxdiff(integral_image, CartesianIndex(tl_y, tl_x), CartesianIndex(br_y, br_x))
+sum = boxdiff(integral_image, tl_y, tl_x, br_y, br_x)
+```
+
+Returns the sum of a window in an image, given the integral image and the top left and
```
timholy: It would be better to describe what the function does, then point out its utility with respect to `integral_image`. Something like:
```
Given an `xrange` and `yrange`, compute the sum of elements at the 4 corners of this box-shaped region weighted by 

    +1  -1
    -1  +1
```

mronian:
Updated the docs now :)

timholy:
That's excellent documentation!

mronian:
Thanks!

mronian:
@timholy Could you tag a new release of Images?

timholy:
All set.

mronian:
Thank you!

Issue #477, opened 2016-04-25
dilate is missing dimension


juliohm:
A recent change introduced a bug:

```julia
using Images: dilate
A = zeros(3,3,3)
A[1,1,1] = 1
B = dilate(A)
```
>3x3x3 Array{Float64,3}:
[:, :, 1] =
 1.0  1.0  0.0
 1.0  1.0  0.0
 0.0  0.0  0.0

>[:, :, 2] =
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0

>[:, :, 3] =
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0

The expected behavior is to perform 3D dilation as I am not specifying any region.

bjarthur:
can you isolate the change to a specific commit using `git bisect` ?

juliohm:
@bjarthur I am not following the development recently, I just found the issue because my package that depends on `Images.jl` was not terminating.

juliohm:
@timholy I think this is a serious issue that may introduce bugs in other packages. @bjarthur could you assign a label to this issue like `bug`?

timholy:
For the record, the last time that code changed in a functional way is May 20, 2015. Hard to imagine it's a recent bug.

juliohm:
That is interesting! I started working with 3D examples just recently, that
is why it didn't affect me earlier.

-Júlio

timholy:
Try it with `A = zeros(3,3,2)`. I'll leave it as an exercise for the reader to figure out what that works.

juliohm:
Thank you Tim, I appreciate your help as usual :blush:  I cannot fix it myself anytime soon, the following weeks are hectic, many things happening at the same time...

timholy:
It's not clear it should be fixed (meaning it's a difficult decision). I'll give you a hint: it's Matlab's fault.

juliohm:
It has to do with MATLAB? What exactly?

-Júlio

timholy:
Try `@which dilate(A)` and I think you'll figure it out.

juliohm:
@timholy it expects an `AbstractArray{T,N}` for both `zeros(3,3,3)` and `zeros(3,3,2)` here. I still don't get it.

timholy:
Notice how the default `region` comes from `coords_spatial(A)`? Let's play with it:
```jl
julia> coords_spatial(zeros(3,3,2))
3-element Array{Int64,1}:
[...]

PR #533, opened 2016-07-31
Add support for N-dimensional arrays in imgradients

juliohm:
This pull request is incomplete.

@timholy could you please guide me on fixing the tests on the `Image` types?

The only difference between this N-dimensional implementation compared to the previous 2D code is that the output is flipped: "grady, gradx" instead of "gradx, grady". The first dimension is vertical (rows), the second dimension is horizontal (columns), and so on...

After all tests pass, I will profile the code further and update the docs.

### DIFF
--- a/docs/src/function_reference.md
+++ b/docs/src/function_reference.md
@@ -375,6 +375,12 @@ integral_image
 boxdiff
 ```
 
+# Pyramids
+
+```@docs
+gaussian_pyramid
+```
+
 # Phantoms
 
 ```@docs
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -276,6 +276,7 @@ export # types
     integral_image,
     boxdiff,
     bilinear_interpolation,
+    gaussian_pyramid,
     
     # distances
     hausdorff_distance,
@@ -302,7 +303,7 @@ Contrast/coloration:
 
 Algorithms:
 
-    - Reductions: `maxfinite`, `maxabsfinite`, `minfinite`, `meanfinite`, `sad`, `ssd`, `integral_image`, `boxdiff`
+    - Reductions: `maxfinite`, `maxabsfinite`, `minfinite`, `meanfinite`, `sad`, `ssd`, `integral_image`, `boxdiff`, `gaussian_pyramid`
     - Resizing: `restrict`, `imresize` (not yet exported)
     - Filtering: `imfilter`, `imfilter_fft`, `imfilter_gaussian`, `imfilter_LoG`, `imROF`, `ncc`, `padarray`
     - Filtering kernels: `ando[345]`, `guassian2d`, `imaverage`, `imdog`, `imlaplacian`, `prewitt`, `sobel`
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1906,3 +1906,29 @@ if VERSION < v"0.5.0-dev+4754"
 else
     chkbounds(::Type{Bool}, img, x, y) = checkbounds(Bool, img, x, y)
 end
+
+"""
+```
+pyramid = gaussian_pyramid(img, n_scales, downsample, sigma)
+```
+
+Returns a  gaussian pyramid of scales `n_scales`, each downsampled
+by a factor `downsample` and `sigma` for the gaussian kernel. 
+
+"""
+function gaussian_pyramid{T}(img::AbstractArray{T, 2}, n_scales::Int, downsample::Real, sigma::Real)
+    prev = img
+    pyramid = typeof(img)[]
+    push!(pyramid, img)
+    prev_h, prev_w = size(img)
+    for i in 1:n_scales
+        next_h = ceil(Int, prev_h / downsample)
+        next_w = ceil(Int, prev_w / downsample)
+        img_smoothed = imfilter_gaussian(prev, [sigma, sigma])
+        img_scaled = imresize(img_smoothed, (next_h, next_w))
+        push!(pyramid, img_scaled)
+        prev = img_scaled
+        prev_h, prev_w = size(img_scaled)
+    end
+    pyramid
+end
--- a/test/algorithms.jl
+++ b/test/algorithms.jl
@@ -198,6 +198,26 @@ facts("Algorithms") do
         @fact int_sum --> 1400
         int_sum = boxdiff(int_img, CartesianIndex((4, 4)), CartesianIndex((8, 8)))
         @fact int_sum --> 1400
+
+        img = zeros(40, 40)
+        img[10:30, 10:30] = 1
+        pyramid = gaussian_pyramid(img, 3, 2, 1.0)
+        @fact size(pyramid[1]) == (40, 40) --> true
+        @fact size(pyramid[2]) == (20, 20) --> true
+        @fact size(pyramid[3]) == (10, 10) --> true
+        @fact size(pyramid[4]) == (5, 5) --> true
+        @fact isapprox(pyramid[1][20, 20], 1.0, atol = 0.01) --> true
+        @fact isapprox(pyramid[2][10, 10], 1.0, atol = 0.01) --> true
+        @fact isapprox(pyramid[3][5, 5], 1.0, atol = 0.05) --> true
+        @fact isapprox(pyramid[4][3, 3], 0.9, atol = 0.025) --> true
+        
+        for p in pyramid
+            h, w = size(p)
+            @fact all([isapprox(v, 0, atol = 0.06) for v in p[1, :]]) --> true
+            @fact all([isapprox(v, 0, atol = 0.06) for v in p[:, 1]]) --> true
+            @fact all([isapprox(v, 0, atol = 0.06) for v in p[h, :]]) --> true
+            @fact all([isapprox(v, 0, atol = 0.06) for v in p[:, w]]) --> true
+        end 
     end
 
     context("fft and ifft") do


juliohm:
Someone free that can help with this?

timholy:
Can you check the `spatialorder` property and adjust accordingly?

This will all be greatly simplified when the new Images is ready.

juliohm:
Hi Tim, this is related to the weird MATLAB convention affecting #477 and #539? What is coming up in the new `Images.jl` release that will solve this issue?

timholy:
All 2d images will be vertical-major, and in general the orientation will be determined by the order of dimensions. (In the "old" images, a "lazy transpose" was intended simply if you switch the order of the `spatialorder` property, but that was too fragile.)

juliohm:
I don't know if I understood it fully, but there is anything that we can do meanwhile to get this PR merged?

timholy:
I can't merge this here with test failures.

If you submit it to ImagesFiltering.jl instead, it might go faster since there aren't any tests yet :smile:. See #542.

src/edge.jl
```diff
@@ -109,35 +109,108 @@ function ando5_sep()
     return f', f
 end
 
-# Image gradients in the X and Y direction
+# Auxiliary functions for computing N-dimensional gradients
+function _gradientpad(img::AbstractArray, border::AbstractString="replicate")
+  border ∈ ["replicate", "circular", "reflect", "symmetric"] || error("border option not supported")
```
timholy: function bodies should be indented with 4 spaces for consistency with the rest of the code

src/edge.jl
```diff
@@ -109,35 +109,108 @@ function ando5_sep()
...
+
+  for dir in 1:ndirs
+    # kernel = centered difference + perpendicular smoothing
+    if extent[dir] > 1
+      kern = _directional_kernel(dir, extent, method)
+      G[dir] = imfilter(imgpad, kern, "inner")
+    end
+  end
+
+  G
```
timholy: return `(G...,)` to return as a tuple.

src/edge.jl
```diff
@@ -109,35 +109,108 @@ function ando5_sep()
...
+
+function imgradients(img::AbstractArray; method::AbstractString="ando3", border::AbstractString="replicate")
+  extent = size(img)
+  ndirs = length(extent)
+
+  # pad input image only on appropriate directions
+  imgpad = _gradientpad(img, border)
+
+  # gradient tuple
+  G = fill(zeros(extent), ndirs)
```
timholy: This could simply be `Array(Array{T,N}, ndirs)` where you add type parameters to `img` in the function signature.

timholy: Hmm, a tricky point is that the element type of the gradient may not be the same as the element type of `img`. The standard trick for this is to say something like `Tret = typeof(img[1]*kern[1])`, as long as `kern` is inferrable.

timholy:
This look pretty good overall, so with those couple of fixes I'd merge it into ImagesFilter.jl

src/edge.jl
```diff
@@ -109,35 +109,108 @@ function ando5_sep()
...
+
+        G[k,dir] = sum(kern .* A)
+      end
+    end
+  end
+
+  G
+end
+
+function imgradients(img::AbstractArray; method::AbstractString="ando3", border::AbstractString="replicate")
```
timholy: Probably also want to add another definition:
```jl
@deprecated imgradients(img::AbstractArray, method::AbstractString, border::AbstractString) imgradients(img; method=method, border=border)
```
so that old code continues to work.

timholy:
Actually, with that deprecation and change to tuple-output, you might be able to get this to pass its tests here. That might be the easier path, since ImagesFiltering changes the way borders are handled.

src/edge.jl
juliohm: Hi Tim, what should the type parameters be? Is the following line ok?

```julia
function imgradients{T<:Number,N<:Integer}(img::AbstractArray{T,N}, ...)
```

juliohm:
Hi Tim, I fixed some of the issues you pointed. Can you please give a second look at it? :)

timholy:
You'll need to rebase on current master: see the "This branch has conflicts that must be resolved" badge above? It can't run the tests until you do that.

timholy:
Instructions: http://docs.julialang.org/en/latest/manual/packages/#squashing-and-rebasing

src/edge.jl
timholy: just `imgradients{T,N}(img::AbstractArray{T,N}, ...)` would be fine

juliohm:
Hi Tim, I rebased the commits, how should we go about the return type of imgradients?

timholy:
Great, thanks.

Re the return type, did you see my comment [here](https://github.com/timholy/Images.jl/pull/533#discussion_r74691083)?

juliohm:
Hi Tim, I saw the comment, the variable `kern` is only defined inside of the loop. How can we alternatively infer the type `Tret`?

timholy:
Ah. The best way to do this is to use tuples in the first place, a programming style that several of us call "lispy":
```jl
function imgradients(img::AbstractArray; method::AbstractString="ando3", border::AbstractString="replicate")
    # pad input image only on appropriate directions
    imgpad = _gradientpad(img, border)
    _imgradients((), imgpad, method)
end

# When all N gradients have been calculated, return the result
_imgradients{T,N}(G::NTuple{N}, imgpad::AbstractArray{T,N}, method) = G

# Add the next dimension to G
function _imgradients{T,M,N}(G::NTuple{M}, imgpad::AbstractArray{T,N}, method)
    d = M+1  # the dimension we're working on now
    extent = size(img)
    if extent[d] > 1
        kern = _directional_kernel(dir, extent, method)
        return _imgradients((G..., imfilter(imgpad, kern, "inner")), imgpad, method)
    end
    # return an all-zeros value for the gradient along this dimension, hopefully using an eltype that matches the result of imfilter
    return _imgradients((G..., zeros(typeof(zero(eltype(imgpad))*0.0), extent)), imgpad, method)
end
```

When using tuples, `(t..., x)` is the equivalent of `push!(t, x)` for Arrays. If you don't understand this, I can just change it myself after merging. (Let me know.)

But the test failures must be fixed first. You know you can run the tests locally, right? To give you a head start:
- the `data` failure is pretty easy: changes `edges.data` to `data(edges)`.
- For the rest of them, do you see the special 2d logic involving `spatialorder` in the method you deleted? You will need to mimic that. One advantage of the implementation I posted above here is that `imgradients` is so short, it's easy to write a specialized version for `AbstractMatrix`.

juliohm:
Thank you Tim, I will read the code carefully later today.

juliohm:
Hi Tim, I fixed the issue with `edges.data`.

I am still wondering if we can get rid of this "special" case with `spatialorder`? Most active `Images.jl` contributors favored the removal of this weird MATLAB convention,  should we keep it? If we don't remove it now, it will be tricky to remove it in the future when `Images.jl` becomes the best package for image processing in the world :)

timholy:
I think you're conflating two issues:
- the MATLAB convention, where the third dimension (if it's of size 3) is color. This horrible convention is already gone in "Images-next" (i.e., the packages at https://github.com/JuliaImages like ImagesCore).
- the use of `spatialorder` as an attempt at "lazy permutations." This is also already gone in "Images-next."

When we switch to Images-next, it will be a breaking change (probably tagged 0.6.0) with a very carefully arranged deprecation strategy and lots of documentation. If I understand correctly, you're asking whether you have to deal with the issue of spatial order for this PR. This is why we have tests :smile:. Sadly, I've caused my share of *accidental* breakages over the years (apologies), but I've tried to avoid *deliberately* breaking anything, ever, in the middle of a release cycle (we're in 0.5.x now). So, unless you fix the spatialorder issues here now, this can't be merged until 0.6.0.

To give you some sense of what I mean by a "carefully arranged deprecation strategy," see files named `deprecated.jl` in the new packages:
- https://github.com/JuliaImages/ImagesCore.jl/blob/master/src/deprecated.jl
- https://github.com/JuliaImages/ImagesCore.jl/blob/master/test/deprecated.jl
- https://github.com/JuliaImages/ImagesMeta.jl/blob/master/src/deprecated.jl
- https://github.com/JuliaImages/ImagesMeta.jl/blob/master/test/deprecated.jl
- https://github.com/JuliaImages/ImagesFiltering.jl/blob/teh/finish/src/deprecated.jl
- https://github.com/JuliaImages/ImagesFiltering.jl/blob/teh/finish/test/deprecated.jl

Altogether it's more than 1000 lines of code, including tests, whose sole purpose is to try to ensure that users migrating to Images-next get helpful tips about how they need to change their code. My biggest worry is that it's not enough.

juliohm:
This is amazing, thank you for all this work and tests, they make everyone's lives much easier!

I will work on the spatial order issue for sure, trying to find time...

juliohm:
Hi Tim,

What needs to be changed in order to get this implemented in the new Images.jl instead?

timholy:
I just updated ImagesFiltering so you could make a PR there.

I'd probably put `imgradients` in `specialty.jl`. The default should be something like
```jl
function imgradients(img, kernels=KernelFactors.ando3(), border=Pad{:replicate}())
```
Note these are positional arguments (like the original) and not keyword arguments; Julia doesn't perform inference on keyword arguments, and one of the strengths of the new ImagesFiltering is that everything is type-stable.

timholy:
Also, I'd fully understand if you have difficulty figuring out the structure of ImagesFiltering, since it's in flux and not yet well documented. I'd be fine with you just copy/pasting what you have here into appropriate places, submitting a broken PR, and then I'll clean it up for you (but you'll get "credit" for your commits).

juliohm:
Hi Tim, I saw your comment in ImagesFiltering.jl, thank you for the update. I will wait for the package to become more stable. Also, I cannot migrate to Julia v0.5 now, I need to finish some research first before trying the bleeding edge.

I will try to fix the issue with spatialorder, it is a little annoying, but necessary as you said.

PR #534, opened 2016-08-01
Adds Corner Peaks, Changes return type to Boolean

mronian:
Since we went with a boolean return for ```fastcorners```, it seems better to have the same for ```imcorner``` as well. Also, finding corners using maxima is easier to use.

### DIFF
--- a/src/corner.jl
+++ b/src/corner.jl
@@ -1,5 +1,6 @@
 """
 ```
+corners = imcorner(img; [method])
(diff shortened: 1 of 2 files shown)


timholy:
If there's a good workaround for 0.4, that would be nice. But "soon" Images will require 0.5, so if needed we can bump.

mronian:
Fixed the errors :)

PR #536, opened 2016-08-04
Adds Gaussian Pyramid (Required for ORB)

mronian:


mronian:
There does seem to be a problem with ImageMagick as it is failing on 0.5 here too.

mronian:
@timholy I am at a loss as to why the build is failing on 0.5 (Tried out a lot of things but it does not pass). I even renamed the function to my name just in case ```gaussian_pyramid``` was defined somewhere (It was showing some MapDict error) and still the build failed. Commenting out every line of the function except the signature and the tests also has no effect. All tests pass on my machine without any problems. Could you look into this?

timholy:
I'm starting to hit this too. https://github.com/JuliaLang/julia/issues/17809 and https://github.com/JuliaLang/julia/issues/17821.

Don't update your julia, and don't bother submitting anything :frowning:,  until this is fixed.

mronian:
@timholy  Tests pass now :)

src/algorithms.jl
```diff
@@ -1906,3 +1906,29 @@ if VERSION < v"0.5.0-dev+4754"
 else
     chkbounds(::Type{Bool}, img, x, y) = checkbounds(Bool, img, x, y)
 end
+
+"""
+```
+pyramid = gaussian_pyramid(img, n_scales, downsample, sigma)
```
timholy: We have something that should be even faster: `restrict`, which simultaneously smooths and downsamples. But this is fine (people will expect it).

mronian: I think it does a two-fold downsampling of the image. Should I try to add to that function itself instead of this?

timholy: Do you mean, should we have a `restrict_pyramid` function? I would say we could have that in addition rather than instead of this.

timholy:
Oh, sorry, I thought I'd merged this already.

mronian:
Thats why I was wondering if you wanted me to change something :stuck_out_tongue: :smile:

PR #543, opened 2016-08-13
fix gaussianpyramid

mronian:
Small changes I had locally done while working on ORB but forgot to push to this PR.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1918,8 +1918,9 @@ by a factor `downsample` and `sigma` for the gaussian kernel.
 """
 function gaussian_pyramid{T}(img::AbstractArray{T, 2}, n_scales::Int, downsample::Real, sigma::Real)
     prev = img
-    pyramid = typeof(img)[]
-    push!(pyramid, img)
+    img_smoothed_main = imfilter_gaussian(prev, [sigma, sigma])
+    pyramid = typeof(img_smoothed_main)[]
+    push!(pyramid, img_smoothed_main)
     prev_h, prev_w = size(img)
     for i in 1:n_scales
         next_h = ceil(Int, prev_h / downsample)


src/algorithms.jl
```diff
@@ -1918,8 +1918,9 @@ by a factor `downsample` and `sigma` for the gaussian kernel.
 """
 function gaussian_pyramid{T}(img::AbstractArray{T, 2}, n_scales::Int, downsample::Real, sigma::Real)
     prev = img
-    pyramid = typeof(img)[]
-    push!(pyramid, img)
+    pyramid = Image[]
+    img_smoothed_main = imfilter_gaussian(prev, [sigma, sigma])
```
timholy: If you reversed the order of these two lines, could you say `pyramid = typeof(img_smoothed_main)[]`? I think once you've run one round of filtering, the others should generate the same output type.

`imfilter_gaussian` may not currently be type-stable, but it is in the new ImageFiltering.

mronian:
I tried the version with `typeof(img_smoothed)` but surprisingly, that one gives a type unstable result. The current one with `typeof(img)` gives a type stable result. Both of them have multiple unstable temporaries though because of `imfilter_gaussian`. 

Should I keep this version?

timholy:
From your "surprise," it sounds like your intuition is right. Just in case it's not fully clear, because of this:
```julia
julia> using Images, Colors

julia> a = rand(Gray{U8}, 5, 5);

julia> eltype(imfilter_gaussian(a, [2,2]))
ColorTypes.Gray{Float64}
```
this version can't be type-stable for general inputs.

In the long run, it has a much greater chance of being type-stable if you use `typeof(img_smoothed_main)[]`; what's working against you at the moment is that `imfilter_gaussian` is not inferrable. (This is already fixed in ImagesFiltering, on the `finish` branch.)
I'd say it's probably best to write the code in a way that looks towards the future, since this version will only be type stable for certain kinds of inputs.

mronian:
Oh, I understand now. I'll make the changes and submit :)

mronian:
@timholy Is it ok if we tag a new release?

timholy:
Done.

mronian:
Thanks!

PR #544, opened 2016-08-21
Add support for N-dimensional arrays in imgradients

juliohm:
Hi Tim,

I accidentally deleted the previous pull request during my Git adventures, opening a new one with corrections.

### DIFF
--- a/src/edge.jl
+++ b/src/edge.jl
@@ -252,7 +252,7 @@ function imgradients{T<:Color}(img::AbstractArray{T}; method::AbstractString="an
     rawimg = reinterpret(eltype(eltype(img)), img)
 
     # handle 2D images differently
-    if ndims(img) == 2
+    if sdims(img) == 2
         _imgradients2D(rawimg, method=method, border=border)
     else
         imgradients(rawimg, method=method, border=border)
--- a/src/edge.jl
+++ b/src/edge.jl
@@ -123,17 +123,22 @@ function _directional_kernel(dir::Int, extent::Dims, method::AbstractString)
     ndirs = length(extent)
 
     # smoothing weights
-    weights = (method == "sobel" ? [1.,2.,1.] :
-               method == "ando3" ? [.112737,.274526,.112737] :
+    weights = (method == "sobel"   ? [1.,2.,1.] :
+               method == "prewitt" ? [1.,1.,1.] :
+               method == "ando3"   ? [.112737,.274526,.112737] :
                error("Unknown gradient method: $method"))
 
+    # finite difference scheme
+    finitediff = (method ∈ ["sobel","prewitt","ando3"] ? [-1.,0.,1.] :
+                  error("Unknown gradient method: $method"))
+
     # centered difference
-    idx = ones(Int, ndirs); idx[dir] = 3
-    kern = reshape([-1.,0.,1.], idx...)
+    idx = ones(Int, ndirs); idx[dir] = length(finitediff)
+    kern = reshape(finitediff, idx...)
     # perpendicular smoothing
     for j in [1:dir-1;dir+1:ndirs]
         if extent[j] > 1
-            idx = ones(Int, ndirs); idx[j] = 3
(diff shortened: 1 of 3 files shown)


juliohm:
A new error is showing in Travis, could you please comment on how to fix it?

src/edge.jl
```diff
@@ -109,40 +109,121 @@ function ando5_sep()
...
+
+    result
 end
 
+@deprecate imgradients(img::AbstractArray, method::AbstractString, border::AbstractString) imgradients(img; method=method, border=border)
+
 function imgradients{T<:Color}(img::AbstractArray{T}, method::AbstractString="ando3", border::AbstractString="replicate")
     # Remove Color information
-    imgradients(reinterpret(eltype(eltype(img)), img), method, border)
+    imgradients(reinterpret(eltype(eltype(img)), img), method=method, border=border)
```
timholy: Your Travis error is probably due to this method not being called anymore (note this one isn't using kwargs).

juliohm: Hi Tim, how can I fix this issue?

timholy: Change the signature to be the same as your other methods (i.e., insert a semicolon after the `img` argument).

juliohm:
Hi Tim, we are almost there. The last error is related to RGB channels. What is the expected behavior for imgradients on the channels? Should it be applied in 2D for each channel separately and the result put together in this 3xmxn convention?

Also, we have another issue with the non-separable kernels, I couldn't figure out a way to tweak the code in order to incorporate ando4, and ando5 for example. Should we have an extra imgradients2D supporting these kernels instead?

timholy:
> The last error is related to RGB channels. What is the expected behavior for imgradients on the channels? Should it be applied in 2D for each channel separately and the result put together in this 3xmxn convention?

The tests are basically what define the behavior we expect/enforce (that doesn't mean it can't change in the future, but this would be a breaking change, not a minor update). That seems to be [what's expected](https://github.com/timholy/Images.jl/blob/f629d05c981f1d676d6ab1d4e05543f181c0fafb/test/edge.jl#L227).

> Also, we have another issue with the non-separable kernels, I couldn't figure out a way to tweak the code in order to incorporate ando4, and ando5 for example. Should we have an extra imgradients2D supporting these kernels instead?

Or allow them when calling `imgradients` on 2d images, but not otherwise.

juliohm:
I tried to rewrite the function to something like:

```julia
function imgradients{T<:Color}(img::AbstractArray{T}; method::AbstractString="ando3", border::AbstractString="replicate")
    # Remove Color information
    rawimg = reinterpret(eltype(eltype(img)), img)
    Rx, Ry = imgradients(rawimg[1,:,:], method=method, border=border)
    Gx, Gy = imgradients(rawimg[2,:,:], method=method, border=border)
    Bx, By = imgradients(rawimg[3,:,:], method=method, border=border)

    gradx = cat(1, Rx, Gx, Bx)
    grady = cat(1, Ry, Gy, By)

    gradx, grady
end
```
without success. I just need to know how the filtering looks like when one of the dimensions means a channel.

Regarding the non-separable kernels, I will add a big if-else for 2D images...

timholy:
Couldn't you have just left the old one there and add the semicolon? I think you don't need to do all that, just reshape the kernel so it's of size 1 along the first dimension.

timholy:
When in doubt, don't forget: you can always commit or stash your current work, check out master, and run a quick test to see how the current code works. Then go back to yours and make adjustments.

juliohm:
Yes Tim, that is probably the best way to handle these special cases. I will just revive the old implementation, call it _imgradients2D() and put a big if-else in the beginning of imgradients.

juliohm:
Hi Tim, I am falling back to the old implementation for 2D images, but I am still getting similar errors, could you please take a quick look when you have time? I will squash all these commits after the tests pass, sorry for the noise, getting more and more used to Git control.

juliohm:
This is moved to `ImagesFiltering.jl`.

PR #546, opened 2016-08-27
blob_LoG: don't discard edges in the "sigma" dimension

timholy:
Ref. https://groups.google.com/d/msg/julia-users/NrSr8IndWAw/l5ezvy-sBgAJ.

CC @bjarthur, hope this is OK. I noticed your test used a triangle-shaped sigma, perhaps not unrelated? If you disagree with this, then we should at least document the issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os:
     - osx
 julia:
     - 0.4
+    - 0.5
     - nightly
 notifications:
     email: false
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,6 +2,8 @@ environment:
   matrix:
   - JULIAVERSION: "julialang/bin/winnt/x86/0.4/julia-0.4-latest-win32.exe"
   - JULIAVERSION: "julialang/bin/winnt/x64/0.4/julia-0.4-latest-win64.exe"
+  - JULIAVERSION: "julialang/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
+  - JULIAVERSION: "julialang/bin/winnt/x64/0.5/julia-0.5-latest-win64.exe"
   - JULIAVERSION: "julianightlies/bin/winnt/x86/julia-latest-win32.exe"
   - JULIAVERSION: "julianightlies/bin/winnt/x64/julia-latest-win64.exe"
 
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1056,16 +1056,18 @@ Note that only 2-D images are currently supported due to a limitation of `imfilt
         end
 
         radii = sqrt(2.0)*sigmas
-        maxima = findlocalmaxima(img_LoG, 1:ndims(img_LoG), false)
+        maxima = findlocalmaxima(img_LoG, 1:ndims(img_LoG), (true, falses(N)...))
         [(img_LoG[x...], radii[x[1]], (@ntuple $N d->x[d+1])...) for x in maxima]
     end
 end
 
-@generated function findlocalextrema{T,N}(img::AbstractArray{T,N}, region::Union{Tuple{Int},Vector{Int},UnitRange{Int},Int}, edges::Bool, order::Base.Order.Ordering)
+findlocalextrema{T,N}(img::AbstractArray{T,N}, region, edges::Bool, order) = findlocalextrema(img, region, ntuple(d->edges,N), order)
+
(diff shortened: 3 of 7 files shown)


timholy:
Hmm. @bjarthur, should we be sorting `sigma`? Now that I think about it, `sigma = [0.5,3.0,0.4,4.0,0.5]` seems like a great way to generate lots of local maxima.

bjarthur:
doesn't hurt to add a tuple of edge conditions to `findlocalextrema`.

however, if `blob_LoG` does not find a blob that the user expects it to only because the `sigmas` they specify do not bound it's size, i'd suggest they use a wider range.  if you use the edges of `sigmas`, you will not know that a found blob is actually that size, as it could fit a smaller or larger sigma better.

and yes, `sigmas` should be sorted if it is not already.  the idea is that you want to find blobs of a certain size, and sigmas that are slightly smaller or larger won't fit as well, and so won't be the local maximum.

https://en.wikipedia.org/wiki/Blob_detection#The_Laplacian_of_Gaussian

timholy:
> if you use the edges of sigmas, you will not know that a found blob is actually that size, as it could fit a smaller or larger sigma better.

True. But the first thing I tried was with a single sigma, and the former code gave no output in that case. That's either something that needs to be fixed with documentation or code changes (here I chose the latter, but we could choose the former if you think that's better).

> and yes, sigmas should be sorted if it is not already. 

The original test supplies the sigmas in non-sorted order, I had wondered if that was intentional.

bjarthur:
oi, my bad on the original test not supplying sorted sigmas.  that is certainly not how lindeberg described it.  though perhaps we should not enforce that, if you and possibly others find it useful.

i can see how considering a maxima at the edge of the sigmas might be useful.  certainly it's more computationally efficient if you don't care if it's a true maxima.  on the other hand, some people will care (like me), so we should try to accommodate both.

for reference is [python's implemenation](http://scikit-image.org/docs/dev/api/skimage.feature.html#skimage.feature.blob_log)

timholy:
Thanks for the link to the python implementation. Since they don't allow one to pass in a vector of sigmas, it's always sorted. I suspect we should do the same.

I agree that returning points on the edges of the sigma range is an indication that the measured size will be inaccurate, but OTOH I'd still say that's probably better than returning an empty array in cases where there really is a blob. If nothing else, the caller can check for that case from the return values. One of the things that makes that harder than it could be is taking `sigma` as an input and returning `radii = sqrt(2)*sigma`. How would you feel about changing that? (For compatibility, we'd probably have to go through a period where that information was passed in with a keyword argument and we issued a deprecation warning.) I'm not sure it has to be changed, but on balance I think it would be more consistent to report in the same units as the input.

bjarthur:
sounds good on all counts.  so to recap proposed changes to the original code:

1. add `sort!(sigmas)`.  or should we retain flexibility and just `issorted(sigmas) || warn("sigmas not sorted")`.  or should it be an `error`.  or should we add a `sort_sigmas_for_me::Bool=true` keyword argument?

2. move `radii = sqrt(2)*sigmas` from the code to the documentation, with an intervening period of providing a `return_radii::Bool=false` keyword argument and an accompanying deprecation warning.  sure would be nice if julia phoned home usage coverage stats so we knew if this was necessary.

3. alter `findlocalextrema` to be more flexible on which dimensions it included the edges.

4. return maxima on the edges of sigma.  but, should we give users an option to return maxima on the edges of the image as well?

timholy:
Comments:

1. Is there a use case for unsorted sigma? I'm inclined to sort and then wait for a complaint.

2. Agreed

3. Already done (this PR).

4. Probably a good idea.

bjarthur:
> Is there a use case for unsorted sigma?

i don't think so, but you had suggested earlier to use `sigma = [0.5,3.0,0.4,4.0,0.5]`

timholy:
No, I just meant that as an example of how to get wrong results :smile:

bjarthur:
ahah.  i didn't pick up on that.  okay, so we are in agreement.

timholy:
Sorry, it was indeed ambiguous. Text is often so limiting.

bjarthur:
to accommodate all preferences, i'd suggest adding an edges argument to `blog_LoG` which is just passed verbatim into `findlocalextrema`.  then the user can either ignore or include the edges of the sigmas and image as they see fit, without having to filter the results afterwards.

PR #552, opened 2016-09-09
Fix depwarn on 0.6 due to vectorized functions

yuyichao:
All tests passes locally. The WIP is because there's [one test](https://github.com/timholy/Images.jl/compare/yyc/0.6-dep?expand=1#diff-efda1537dfa19ffae4976ef9a6bf697cL33) that I need to change. The test starts to fail on 0.4 half way when I'm making the changes and I have no idea why it's happening.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,5 +6,5 @@ SIUnits
 Zlib
 Graphics 0.1
 FileIO
-Compat 0.8.4
+Compat 0.9.1
 StatsBase   # for histrange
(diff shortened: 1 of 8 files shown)


yuyichao:
The Travis linux nightly failure looks new and unrelated. @tkelman in case it's known or worth attention.

yuyichao:
Figured out what's wrong with the test and should be fixed now. Not sure about the libgit2 issue (if it appears again).

timholy:
Looks great, thanks a bunch!

PR #560, opened 2016-09-27
Fix types due to changes in UFixed promotion

timholy:
Also fixes a couple of deprecation warnings


timholy:
Hmm, the `@generated` function error on 0.6 is interesting. I plan on rewriting that function anyway for the new Images.

PR #561, opened 2016-10-03
change a few backticks in docstrings

visr:
Because it confused the syntax highlighting in Atom.


timholy:
Thanks! Sorry I failed to notice this for so long.

PR #565, opened 2016-10-17
Fix for removal of float32/float64. Fixes ImageView issue 102

timholy:


src/map.jl
```diff
@@ -707,8 +707,13 @@ for (fn,T) in ((:float32, Float32), (:float64, Float64), (:ufixed8, UFixed8),
         end
         $fn{C<:Colorant}(img::AbstractImage{C}) = shareproperties(img, $fn(data(img)))
     end
+    if VERSION >= v"0.5.0"
```
timholy: Now that 0.5 has been out a while, do we really need to? The 0.5 branch of Images doesn't have long to live anyway, and I've fixed this in a different way for the 0.6 release.

On a related topic, is it OK in the REQUIRE file to switch from `julia 0.5-` to `julia 0.5` without going through a minor bump?

tkelman: Still may require bisecting. What was the change this is compensating for, removal of `float32` deprecations?

Yes, dropping the `-` can be done whenever.

timholy: The logic being that if they were deprecated even before the release of 0.4.0, the specific moment of removal during the julia 0.5 release is irrelevant.

PR #566, opened 2016-10-17
Rework interface to blob_LoG and reimplement findlocalextrema

timholy:
At the moment, julia-0.6 is picky about what can and can't be in a `@generated` function, and `findlocalextrema` seems not to meet these requirements. As a consequence our tests have been failing on `nightly` for some time (e.g., https://travis-ci.org/timholy/Images.jl/builds/168159481).

I've gotten rid of a lot of the `@generated` functions in the `images-next` branch, and this PR (against that branch, not master) removes the last one, `findlocalextrema`. As a bonus, it's about 30-fold faster in my hands (I'm not quite sure why, the old version looked mostly OK, though if I had to guess the `in(d,region) ? x : y` test was to blame).

I also restructured `blob_LoG` according to the discussion in #546, and even made the output its own type. This is obviously a breaking change, but since the 0.6 of release of Images will contain several other (unavoidable) breaking changes I don't think this one is exceptional. Fundamentally it's quite hard to have a good deprecation strategy for situations in which the output changes. We could easily add a `getindex(::BlobLog, i) = error("the output of blob_LoG has changed, see the help and NEWS.md")`, but I'm uncertain whether that would actually be better than the `MethodError`.

CC @bjarthur

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -95,7 +95,7 @@ Key changes (of which many are breaking):
 - Many deprecation warnings were designed to help users of the current
   Images package transition to the new framework.
 
-Other changes:
+Other changes (all of which are breaking):
 
 - The gradient components returned by `imgradients` match the
   dimensions of the input; in `g1, g2, ... = imgradients(img,
@@ -117,6 +117,13 @@ Other changes:
 
 - The old `extrema_filter` discards the edges of the image, whereas
   the new one (based on `mapwindow`) returns an array of the same size as the input.
+
+- The output of `blob_LoG` is now a `Vector{BlobLoG}`, a new exported
+  immutable, rather than the old tuple format.
+
+- `findlocalextrema` now returns a `Vector{CartesianIndex{N}}` rather
+  than a `Vector{NTuple{N,Int}}`. This makes it ready for use in efficient
+  indexing.
 
 # Older versions
 
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -64,6 +64,7 @@ include("distances.jl")
 include("deprecated.jl")
 
 export # types
+    BlobLoG,
     ColorizedArray,
 
     # macros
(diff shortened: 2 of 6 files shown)


timholy:
I had to rebase and CI didn't restart. Will close/reopen.

bjarthur:
30x faster is great!

re. https://github.com/timholy/Images.jl/pull/546, how about hoisting the edges argument to `findlocalextrema` up to the inputs of `blob_LoG` so that the user can specify whether to find blobs on the edge of the image / sigma, or not.  as it stands, maxima occurring at the extreme sigma are counted as blobs, which is not definitely what my use case needs.  and some folks might want to automatically include blobs on the edges of the image, which currently are discarded.  flexibility would be good here.

timholy:
Good idea. See if you like this better.

bjarthur:
lgtm

however, if i wanted to nitpick, i'd add

1. i think a default `edges` input of just `false` would be the more likely use pattern

2. would be nice to have tests for the cases in which the blobs were on the edges

3. wonder whether `findlocalextrema` would be better homed somewhere in `JuliaArrays` to reflect it's possibly more general use.

timholy:
> 1. i think a default edges input of just false would be the more likely use pattern

The problem here is that you can easily get *no* output. I was quite confused by this myself the first time I used this, because it's a very hard problem to diagnose. In contrast, if you return something then the user can notice that all the extrema are on the edges.

> 2. would be nice to have tests for the cases in which the blobs were on the edges

Hey, I was just using the old tests provided by someone else :wink:. But yeah, that would be a good thing to have.

> wonder whether findlocalextrema would be better homed somewhere in JuliaArrays to reflect it's possibly more general use.

I'd say that seems reasonable. Feel free to move it after this merges.

bjarthur:
thanks for beating me to this!

Issue #542, opened 2016-08-13
Countdown to the new Images.jl


timholy:
There have been justifiable questions about "what exactly is happening?" so let me post my TODO list here (including completed items, with an explanation of purpose):
- [X] ImagesCore.jl: handles low-level tasks like color separations, raw <--> ufixed conversions, dimension permutations. The new feature is the use of julia's swanky new array technology to provide views for all of these operations, so that we don't sometimes-copy-and-sometimes-return-a-view. Note this also effectively (partly in conjunction with the next package) solves the "Images is really confusing about image orientation") problem, becuase of the dimension permutation views.
- [X] ImagesAxes.jl: images with "named" dimensions, most especially temporal dimensions (like movies). Based on AxisArrays.jl, which also allows one to specify a `Range` (even one with physical units) for each axis, which thus incorporates the `pixelspacing`. All this information is exposed to the type system (rather than through fields of the properties dict). What that means is that indexing operations become type-predictable: for example, if you slice at a particular spot in time, the new array doesn't have a time axis anymore. In the old Images.jl we had to check for this and set the `timedim` property to 0 at runtime, and because that's slow we were constantly throwing away the dictionary for performance and therefore losing track of orientation. Now everything should Just Work.
- [X] ImagesMeta.jl: for those cases where one still wants to attach metadata to images.
- [x] ImagesFiltering.jl: better `imfilter` implementations. Everything will be type-stable, efficient algorithms will be chosen automatically based on the kernel size, and a platform will be in place for more efficient operations with separable kernels. Also options for performing the computation on the GPU, etc. This package will also likely be the first place to test another experimental direction: the ability (at the user's choice) to dispatch to algorithms in other libraries like OpenCV. (Currently this package is well underway but still needs work, and has been the occasion for me making changes to julia and a half dozen other packages. It won't be runnable by anyone else until julia-0.5-rc3 at the earliest.)
- [x] rename the above packages to ImageCore etc.
- [x] Get necessary dependencies registered and tagged. In addition to the above-mentioned packages, these include:
  - [x] AxisArrays
  - [x] SimpleTraits
  - [x] CatIndices
  - [x] FFTViews
  - [x] ComputationalResources
  - [x] Ranges (https://github.com/JuliaArrays/Ranges.jl) for Ranges that are safe with [Unitful](https://github.com/ajkeller34/Unitful.jl) quantities (cf https://github.com/JuliaLang/julia/pull/18777)
[...]

PR #567, opened 2016-10-18
convert tests from factcheck to testsets

jw3126:


timholy:
This is a pretty massive conflict with #542 and the [images-next branch](https://github.com/timholy/Images.jl/tree/images-next).

timholy:
Is this the product of a script? If so, I'd certainly be happy to have it run on `images-next`. One slightly unusual thing there: I added a string-tag to many FactCheck tests, simply because they give poor backtraces (one of the many advantages of testsets).

Evizero:
are the `-->` more compareable to `==` or `===`? If the latter is the case than the tests might not be equivalent

jw3126:
`-->` is `==` and `--> exactly(...)` is `===`. This is mostly work of a [script](https://github.com/jw3126/FactCheck2TestSets.jl), but also about 1 hour of manual fixup. So I could wait for images-next and do it again, but I can't do it several times. Just ping me when you want the final migration.

timholy:
Fixed by #577 and #584. Thanks for your efforts, @jw3126, I'm sorry I didn't take advantage of your very kind offer!

PR #573, opened 2017-01-12
Fix 0.6 depwarns

yuyichao:
This fixes all of the test errors and most of the depwarns on master. Passing tests on master requires https://github.com/JuliaIO/FileIO.jl/pull/94 and some depwarns are fixed by https://github.com/JuliaArchive/FactCheck.jl/pull/80

The depwarns that are not fixed are:

* Extending `.+`, `.==` etc

    Might need `@compat` support and might also need to decide what's the expected return type with loop fusion

* `@test_approx_eq`

    It seems to behave differently wrt `NaN`?

* `reduce_dims`

    Might want to go to `Compat.jl`

### DIFF
--- a/test/edge.jl
+++ b/test/edge.jl
@@ -1,6 +1,14 @@
-using Images, FactCheck, Base.Test, Colors
+module TestImagesEdge
+using Images, FactCheck, Base.Test, Colors, Compat, Graphics
 
 global checkboard
+
+if VERSION < v"0.5.0"
+    # Overwrite `Base.all` to work around poor inference on 0.4
+    function all(ary)
+        Base.all(convert(Array{Bool}, ary))
+    end
+end
 
 facts("Edge") do
 
@@ -138,9 +146,9 @@ facts("Edge") do
 
             # Test that orientation is perpendicular to gradient
             aorient = orientation(agx, agy)
-            @fact all((cos.(agphase) .* cos.(aorient) .+
-                       sin.(agphase) .* sin.(aorient) .< EPS) |
-                      ((agphase .== 0.0) & (aorient .== 0.0))) --> true
+            @fact all((|).(cos.(agphase) .* cos.(aorient) .+
+                           sin.(agphase) .* sin.(aorient) .< EPS,
+                           (&).(agphase .== 0.0, aorient .== 0.0))) --> true
                       # this part is where both are zero because there is no gradient
 
             ## Checkerboard Image with row major order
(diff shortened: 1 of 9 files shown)


yuyichao:
Sign.... 0.4 support might require doing something about broadcast shape in compat.....

yuyichao:
Also for 0.5....

yuyichao:
.... End up working around it in the tests since I'm too lazy (and not sure how) to fix those properly on old julia versions.

There are a few pretty ugly or evil workaround but luckily they are all in the tests....

Local tests passes on 0.4 and 0.5...

test/map.jl
```diff
@@ -200,10 +208,10 @@ facts("Map") do
         @chk map(mapi, 0) reinterpret(RGB24, 0x00000000)
     end
 
-    context("ScaleAutoMinMax") do
+    @compat context("ScaleAutoMinMax") do
         mapi = ScaleAutoMinMax()
         A = [100,550,1000]
-        @chk map(mapi, A) @compat UFixed8.([0.0,0.5,1.0])
+        @chk map(mapi, A) UFixed8.([0.0,0.5,1.0])
```
yuyichao: Ah, this one was a accidental change. I can change it back if someone feels very strongly about it... Otherwise I'll just leave it like this....

yuyichao:
All test errors on master are from `FileIO` and should be fixed by https://github.com/JuliaIO/FileIO.jl/pull/94

yuyichao:
This does not regress on 0.4 and 0.5 and since the FileIO PR is merged (though not tagged) I'll merge this during the weekend if there's no more comments.

test/edge.jl
```diff
@@ -138,9 +146,9 @@ facts("Edge") do
 
             # Test that orientation is perpendicular to gradient
             aorient = orientation(agx, agy)
-            @fact all((cos.(agphase) .* cos.(aorient) .+
-                       sin.(agphase) .* sin.(aorient) .< EPS) |
-                      ((agphase .== 0.0) & (aorient .== 0.0))) --> true
+            @fact all((|).(cos.(agphase) .* cos.(aorient) .+
```
timholy: I'm all in favor of the improvements in efficiency that the new capabilities bring, but it's Interesting how this is a step backwards in terms of readability. If one were to abandon 0.4 I suppose we could use a generator.

timholy: Not that I'm proposing you should do that. Coming in #542.

test/map.jl
timholy: It's fine.

test/edge.jl
yuyichao: The issues that affect 0.4 are https://github.com/JuliaLang/Compat.jl/pull/306 and https://github.com/JuliaLang/Compat.jl/issues/282. If https://github.com/JuliaLang/Compat.jl/pull/306 is merged and used here, the only necessary change should be replacing `|` with `.|`.

yuyichao: (and also `&` with `.&` .....)

yuyichao:
> I had been assuming you were waiting for `FileIO` to be tagged

Not really, since this shouldn't create regression without it and doesn't really depend on anything there. The two are basically independent fixes....

test/edge.jl
timholy: To clarify, I wasn't complaining about this specific use, just making a more general language observation. I'm fine with how you've done it here.

timholy:
Well, like I said, I approve, so merge when you wish. Thanks for the contribution!

PR #574, opened 2017-01-16
Display a matrix of Images in a table in HTML

shashi:
screenshot from 2017-01-16 16-22-15


src/writemime.jl
```diff
@@ -58,3 +58,36 @@ mapinfo_writemime_(img::AbstractImage) = Images.mapinfo(UFixed8,img)
...
+end
+
+function downsize_for_thumbnail(img, w, h)
+    a,b=size(img)
+    a > 2w && b > 2h ?
+        downsize_for_thumbnail(restrict(img), w, h) : img
+end
+
+function show_element(io::IOContext, img::Image)
+    io2=IOBuffer()
```
shashi: Creating a new IOBuffer here was crucial. There seems to be a serious bug with some bytes being LOST if you set `io2=io` here. Try the example with the change!

timholy: Oh my. Which version of Julia were you testing on? I wonder if it's specific to `Base64EncodePipe`?

timholy:
You probably noticed that this busted 0.4. If you aren't in a hurry, we could commit this once 0.5 becomes the minimum version. My goal is to merge the big Images revamp by the end of the weekend, so that may happen quite soon.

A test would not be amiss.

src/writemime.jl
shashi: This was on 0.5.0, trying this right now on julia master.

timholy:
Now that the new Images has been merged, you don't have to worry about 0.4 support anymore. But needs a rebase. I recognize that a really good test might be hard to write, but if nothing else perhaps just make sure that your demo call doesn't error?

src/writemime.jl
```diff
@@ -58,3 +58,36 @@ mapinfo_writemime_(img::AbstractImage) = Images.mapinfo(UFixed8,img)
 
 mapinfo_writemime_restricted{T<:Colorant}(img::AbstractImage{T}) = ClampMinMax(to_native_color(T), 0.0, 1.0)
 mapinfo_writemime_restricted(img::AbstractImage) = Images.mapinfo(UFixed8, img)
+
+function Base.show{T<:Image}(io::IOContext, m::MIME"text/html", imgs::AbstractMatrix{T})
```
timholy: This will be harder now (in julia 0.5, not julia 0.6). Maybe define
```julia
typealias ColorantArray{T<:Colorant,N} AbstractArray{T,N}
```
and then make this `T<:ColorantArray`.

shashi:
Generalized this for N-dimensional arrays of images...

screenshot from 2017-03-26 20-57-23

Please squash merge this, I put a lot of the changes in the merge commit instead of a descriptive one.

cc @dpsanders

timholy:
Thanks, this is really nice!

PR #575, opened 2017-01-20
Save up memory in hausdorff distance

juliohm:
Those moments in which you revisit code you have written and don't understand why you did what you did. This patch cleans up unnecessary memory allocation and puts a little bit more work in the processor.

I used `BenchmarkTools.jl` to compare implementations, there is a noticeable speedup. Can you please review and merge?

### DIFF
--- a/src/distances.jl
+++ b/src/distances.jl
@@ -20,15 +20,15 @@ function hausdorff_distance(imgA::AbstractArray, imgB::AbstractArray)
   # return if there is no object to match
   (isempty(A) || isempty(B)) && return Inf
 
-  # grid coordinates (ndims by npoints)
-  A = hcat([Float64[ind2sub(size(imgA), a)...] for a in A]...)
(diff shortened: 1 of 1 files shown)


juliohm:
The build is successful in stable. It fails in nightly for unrelated reasons.

BTW, how far is `Images.jl` to supporting Julia v0.6? I am afraid my package won't work in a month or so when the new version of the language is released. We are publishing a paper with our `ImageQuilting.jl` package and it will be embarrassing to have it not functioning by the time the paper is out.

timholy:
Thanks to @yuyichao it passed tests on 0.6 before the type system change. I haven't tried since then, I suspect ColorTypes is currently broken (https://github.com/JuliaGraphics/ColorTypes.jl/issues/65).

timholy:
FYI within a month (and actually in considerably less time) #542 will be merged. So there will be disruptions.

juliohm:
Thank you Tim, I will keep an eye on the changes, hopefully it will be a smooth transition after all the work you have put on docs and tests.

Issue #553, opened 2016-09-09
Best practices for handling multi-channel movies in Images.jl discussion


tlnagy:
I'm not sure that this is the best place to ask this, but I generally take movies that are broken up along channel info and time stored as separate TIFF files in a directory (using MicroManager by the way). Currently I'm just using a bunch of looping `load` calls for everything in `readdir`. Is there a better way of handling this? I would prefer to have the time dimension and channel info in the image metadata if possible.

timholy:
TIFF is in many ways a terrible format. I like [NRRD](http://teem.sourceforge.net/nrrd/), which is (decently?) supported by [NRRD.jl](https://github.com/JuliaIO/NRRD.jl). It supports `mmap`ping a large file and should load metadata properly, but obviously please do report any issues you discover.

timholy:
(NRRD lets you concatenate all your image files into a single file.)

tlnagy:
Is that what your lab uses? I agree TIFF kinda sucks for this use-case. Also, is it possible to tell if a file that is loaded is being `mmap`ped or loaded into memory? There's not too much documentation on it.

timholy:
We have a custom acquisition program that outputs in our own format,  [ImagineFormat](https://github.com/timholy/ImagineFormat.jl). But the binary data are in the same format as NRRD (which is to say a raw dump); it's only the header that's different.

I don't know of a way to tell whether an array is `mmap`ped. IIUC, it's possible that most of your arrays are, as `malloc` uses `mmap` internally for large requests.

tlnagy:
NRRD.jl has some issues, I have some fixes here: https://github.com/JuliaIO/NRRD.jl/pull/3. It appears that Fiji's NRRD writer has some weird behavior because it encodes a XYT image as 

```julia
Gray Images.Image with:
  data: 2048×2044×96 Array{UInt16,3}
  properties:
    spaceunits:  pixel pixel pixel
    colorspace: Gray
    spatialorder:  x y z
    spacedirections:  [1.0,0.0,0.0] [0.0,1.0,0.0] [0.0,0.0,1.0]
    comments:  Created by Nrrd_Writer at Sun Sep 11 13:43:43 PDT 2016
    pixelspacing:  1.0 1.0 1.0
```
[...]

Issue #545, opened 2016-08-23
Convert a data matrix to Image type and display it via view do not work


BoundaryValueProblems:
I'm simply trying to display a data matrix of type float64 using the ImageView.view function.
This particular matrix ranges from 23.0 to 226.0, so it's already within 0 to 255. Hence, I thought that the following would work; say, data below is a float 64 array of size 128 x 128:

```
julia> img=convert(Image, round(UInt8, data))
Gray Images.Image with:
  data: 128x128 Array{UInt8,2}
  properties:
    timedim: 0
    colorspace: Gray
    colordim: 0
    spatialorder:  y x
    pixelspacing:  1.0 1.0
julia> view(img,pixelspacing=[1,1])
```

This pops up an ImageView window, but nothing is displayed inside the window; it's simply empty. I would appreciate it if you could point me out what is wrong here.
Thanks a lot!

timholy:
Intensities always default to the range 0->1. See the Images docs.

BoundaryValueProblems:
Thank you very much! It worked after scaling it within [0,1] range.
It would be great if you could put a link to that Images docs in README.md since there may be many people who want to display a given matrix that is **not** loaded from typical image format files such as .png/.jpg files as a grayscale image using the "view" function. I think the link to the Images docs is the following, am I right?
https://github.com/timholy/Images.jl/blob/master/docs/src/core.md
Thanks!

timholy:
Already described at https://github.com/timholy/Images.jl#loading-your-first-image-how-images-are-represented?

BoundaryValueProblems:
Thanks, Tim!
I missed that portion when I read README.md.
Perhaps, it may be a good idea to emphasize that paragraph.

timholy:
Good to know. I am rewriting Images and its documentation now, so that's timely advice.

BoundaryValueProblems:
Thanks, Tim.
Something like the following example will help many people (I used your earlier posting as an example).
```
# viewing some synthetic matrix data of size 500 x 500 using grayscale
julia> y = linspace(0,2pi,500); x=y';data=sin(x .+ cos(x .* y))
julia> view(convert(Image, map(ScaleAutoMinMax(U8),data)))
```
Thanks!

timholy:
Closed by #577

Issue #539, opened 2016-08-07
Spatial Order Future


mronian:
While creating tests with real images for ORB and BRIEF, I realised that the `Image` type has the spatial order [x y] by default. Since I have been testing with hand made arrays, I have used the order [y x] in most functions (changing that right now in ImageFeatures.jl). To make sure, I went through all my PRs for Images.jl and almost all are fine (since they use `CartesianRange` to iterate) except for the following : 
[ ] `boxdiff`
[ ] `bilinear_interpolation`
[ ] `clahe`

I think `integral_image` is fine(https://github.com/timholy/Images.jl/pull/508/files#diff-77d828dbeb561ecf3e21705f067a485dR1963). @timholy Could you confirm?

Also, with the `JuliaImages`, do you plan on having the same spatial order?

Evizero:
IMHO [y,x] should be the appropriate interpretation for plain arrays (It already is AFAIK, and I hope it will remain as such)

mronian:
The representation for plain arrays is [y x] but the `Image` type changes the spatial order to [x y] due to which all functions have to be written to suit that order. The difference can be seen by - 
```jl
save("res.jpg", img)
save("res2.jpg", data(img))
```

mronian:
This difference in spatial orders is creating a problem while porting tests to the [x y] order. How do I change a normal Array to Image with the switched spatial order?

timholy:
My plan is to ignore this problem :smile:. The reason being that the rewrite of Images is getting pretty close---I estimate about a week of work to go. Solving this problem is perhaps the single biggest motivation for the rewrite, so I don't think it's worth worrying about this any longer within "classic" Images. (If anyone wants to try, go for it, but I won't spend any thought on it myself.)

In the new Images, my plan is that orientation is *always* determined by the order of dimensions---if you want to swap the order without actually computing the permutation, there is [`permuteddimsview`](https://juliaimages.github.io/ImagesCore.jl/latest/views/#Views-1) which implements a lazy permutation. This is a much more robust way than setting the `spatialorder` property of the dictionary.

One thing we do have to decide: should they be "vertical major" (yx) or "horizontal major" (xy)? There are arguments both ways:
- "vertical major" is consistent with the ways that arrays are printed in the REPL. (This dates to a horrible decision to number matrices in row,column order rather than column,row order as would be consistent with Cartesian coordinates. Whoever made that decision back in the 1700s should be chastised mercilessly :smile:.)
[...]

Issue #471, opened 2016-04-14
Doc links broken


timholy:
At the bottom of the README, there are links to `core`, `function reference` and `Overlays`. These work on GitHub (though the markup shows in, e.g., `function reference`), but not on [github.io](http://timholy.github.io/Images.jl/home/). Should we make them all links to the io page?

timholy:
Oh, also the images seem to be broken for me on github.io. That seems more serious. If you know how to fix it quickly, @rsrock, be my guest :smile:, otherwise I will learn how this works. Seems like https://github.com/MichaelHatherly/Documenter.jl is the new upcoming standard? (See deprecation notice at https://github.com/MichaelHatherly/Docile.jl.)

rsrock:
No problem, I'll take care of it (probably on the weekend).  It looked to me like Documenter has the ability to let Travis update the docs for us, so we man not need any manual intervention in the future.  I'm looking forward to that!

timholy:
Closed by #577

PR #577, opened 2017-01-29
The new Images is here

### LINKED ISSUE #73: filtering improvements
This is an "up-for-grabs" issue.

While `imfilter` has been substantially improved, there's room for several other advances.
## 1d optionally-in-place filtering

Mostly inspired by #67: people who filter with separable kernels of small size (e.g., Sobel) would probably benefit from a more efficient algorithm for 1d filtering. Currently a separable 2d filter requires 4 passes through the image:
- Copy & pad
- Filter along dimension 1
- Copy & pad
- Filter along dimension 2

This could be reduced to 2 passes with a sufficiently-clever algorithm. It should be written as a not-in-place algorithm, i.e.,

```
function filter_1d_nonaliasing!(dst, src, kernel::Union(Vector,NTuple), dim::Integer, border::String = "replicate", value = 0)
```

but safe to use with `dst==src`, i.e., the 2 passes through the image would be

```
output = Array(...)
filter_1d_nonaliasing!(output, img, kernel1, dim1)      # filters-and-copies
filter_1d_nonaliasing!(output, output, kernel2, dim2)  # filters in-place
```

For the implementation, the idea is that the result would be maintained in a circular buffer and copied to `dst` only after the filtering has progressed beyond the point where the target pixel contributes to the result. Note that for efficiency (http://julialang.org/blog/2013/09/fast-numeric/, section on "cache-friendly codes") it will be crucial to traverse the array in storage-order, meaning that separate implementations for `dim=1` and `dim>1` will need to be written ([example1](https://github.com/timholy/Images.jl/blob/master/src/algorithms.jl#L931-L933) and [example2](https://github.com/timholy/Images.jl/blob/master/src/algorithms.jl#L1083-L1124) in Images, shorter [example](https://github.com/JuliaLang/julia/blob/master/base/reducedim.jl#L96-L108) in base Julia). For filtering along dimension `dim`, the circular buffer will have to be `dim`-dimensional, equal to the image size for the first `dim-1` dimensions and equal to the filter length along dimension `dim`. It also has to handle boundary conditions without resorting to `padarray`. Finally, it should use Cartesian indexing to permit good performance for arbitrary `AbstractArray`s, since we can't guarantee that `img` will be an `Array` unless we make a copy (which would introduce 1 extra pass through the data).

Given these requirements, this will be a reasonably-complex algorithm, but for small filters (e.g., of length 3) this could presumably lead to an approximate 2-fold performance boost. It could also be used to handle the [small-sigma case](https://github.com/timholy/Images.jl/blob/master/src/algorithms.jl#L931-L933) for `imfilter_gaussian` by allowing in-place FIR filtering.

UPDATE: a big part of the complexity could be eliminated by giving up on the idea of in-place filtering, and using it as

```
output = Array(...)
tmp = similar(output)
filter_1d!(tmp, img, kernel1, dim1, border, value)
filter_1d!(output, tmp, kernel2, dim2, border, value)
```

You wouldn't need the circular buffer. It would be a little harder to make this work with `imfilter_gaussian`, but that's a much smaller obstacle. You'd probably want to check for `dst===src` and throw an error.
## Parallelizing filtering using SharedArrays

Currently `imfilter` is slower than its Matlab equivalent, presumably due to the usage of IPP in Matlab. A special variant of [`_imfilter!`](https://github.com/timholy/Images.jl/blob/master/src/algorithms.jl#L722-L742) could be written for `SharedArray` input & output that parallelizes filtering.

timholy:
As discussed in #542. See the documentation at [JuliaImages](http://juliaimages.github.io/latest/) (the documentation will be fully up to date only after this is merged and tagged).

Fixes #73, #124, #187, #267, #319, #368, #375, #367, #386, #447, #449, #451, #471, #539, #542, #545, #553. That should be enough for now :smile:.

This interacts with many other packages, and I've realized that many people might have a subset of packages checked out on master and another subset following released versions, so it will be too disruptive to wait a week or so before tagging (as I had indicated at one point). Consequently, I'm tagging as soon as this merges.

It would be great if QuartzImageIO could update fairly soon (https://github.com/JuliaIO/QuartzImageIO.jl/issues/28), since Mac users are going to encounter trouble until that happens. I've given rather a lot of warning about this, hoping to have a PR queued already, but since little has changed since October it looks like we'll just have to figure it out after the fact.


.travis.yml
```diff
@@ -3,18 +3,15 @@ os:
...
 julia:
-    - 0.4
     - 0.5
     - nightly
 notifications:
     email: false
-script:
-    - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-    - julia -e 'Pkg.clone(pwd()); Pkg.build("Images")'
-    - julia -e 'Pkg.test("Images", coverage=true)'
```
vchuravy: You seemed to have turn Travis testing off accidentally in a0d155ab8

tkelman: if you comment out or delete the entire `script:` section it defaults to a version that should be equivalent to this standard boilerplate

timholy: No, it's still running, it's just using the default script.

timholy:
Wow, getting it to pass tests on 0.5 was surprisingly easy (most of the hassles were in dependent packages). So this is good to go from a CI perspective. I'll leave it open for a bit, since @vchuravy seems to be reviewing (yay!).

tkelman:
Same (not urgent at all) "would be nice if the tests cleaned up after themselves" issue I've raised elsewhere, though the only file creation I can obviously see is `test/old/writemime.jl` which I'm guessing from the name might not be there forever?

tkelman:
> Mac users are going to encounter trouble

How much trouble are we talking?

vchuravy:
> since @vchuravy seems to be reviewing (yay!).
If you can give me about 24h I can try to do a full review, but it is getting late in Japan so I won't manage it today.

tkelman:
I skimmed the whole diff and it looks sane on a quick inspection, but I do wonder how many downstream packages are going to break from this.

timholy:
Here's my in-progress draft of the announcement I plan for discourse:

Big changes in Images.jl (v0.6.0)

Hi everyone,

Images.jl, a package for processing and visualizing images, has essentially been rewritten from scratch and version 0.6.0 has been tagged. Because this is a major and disruptive release, I suspect a few words are in order.

Images.jl is one of the oldest packages in the Julia ecosystem, and its code-origins predate the existence of both the package manager and Julia 0.1. From the beginning of my own involvement, my main goal was to ensure that it would be useful for both "computer vision" (largely a world of 2d images) and for big-data 3D/4D/5D image analysis. We were partially able to pull that off, but there were some ugly compromises, most notably encoding important information such as the spatial orientation and "meaning" of the axes of an image using an easily-lost or desynchronized/confused dictionary. This may have been the best we could do at the time. However, the Julia language has evolved considerably, and with the release of Julia 0.5 it became apparent that all the pieces were in place for a better approach.

The new Images makes far better use of Julia's type system. The dictionary is still available for those who want it (through the ImageMetadata package, which is "bundled" with Images itself), but most people should be able to just use "plain arrays," more specifically a variety of AbstractArray types defined in various packages. Key advances that made this possible include packages such as AxisArrays (which allows one to give names to axes and encode pixel geometry in a type-system-friendly fashion), MappedArrays (which allows one to extend `reinterpret`-style functionality to arbitrary AbstractArrays), and the existence of the PermutedDimsArray in Base (for lazy-changes of orientation with images too large to handle in memory). Capitalizing on these and other advances in Julia's treatment of AbstractArrays is the main motivation for the rewrite.

From a user's perspective, the first thing you're likely to notice is a whole slew of deprecation warnings and errors, which I know is a major hassle and for which I apologize profusely. (If you have an impending grant deadline, your best bet will probably be to pin the package at an earlier release.) One particular pain-point is for Mac users, who for now likely need to rely on ImageMagick (`using ImageMagick`) to save and load images. I'm hopeful that QuartzImageIO will update to the new version soon. More than a thousand lines of code were devoted to nothing but issuing deprecation warnings (and where necessary, error messages) that attempt to coach you through rewriting your code; I hope these will at least somewhat reduce your pain in making this transition.

While the main emphasis of this rewrite has been on "core" functionality, there are a few goodies to tempt you (e.g., the new `mapwindow` for applying an arbitrary function over moving windows of an image). The most important reason to upgrade, however, is that I suspect you'll find the new version to be considerably more robust and well-behaved than the old. 

To learn more, the two main resources are:

- a new [NEWS.md](https://github.com/timholy/Images.jl/blob/fixed-renaming/NEWS.md) file, which describes the major changes in greater detail
- a completely new set of [documentation](http://juliaimages.github.io/latest/), now hosted at the JuliaImages umbrella organization

Finally, Images.jl has been transferred to JuliaImages. Going forward, I hope the main emphasis will be on added functionality, either added to Images.jl itself or implemented in separate packages (Images is evolving into more of an "umbrella" package). I also hope that users contribute "demos" to the JuliaImages documentation, similar to that of other image-processing frameworks (e.g., http://scikit-image.org/docs/dev/auto_examples/). We'd be excited to receive contributions to source code, documentation, or the test suite!

timholy:
> How much trouble are we talking?

I think they basically have to say `using ImageMagick` if they want to read/write images or use IJulia---the Images tests don't pass unless I bypass QuartzImageIO, despite the fact that the only IO functionality tested in this package is "writemime"-like functionality (the more serious tests of IO residing in ImageMagick, QuartzImageIO, etc.)

timholy:
> If you can give me about 24h I can try to do a full review, but it is getting late in Japan so I won't manage it today.

Willing. If this were early in the queue of merges, I'd say no way because the code has been out there for ages on branches with multiple pleas for testing and review. I'm also carefully juggling my own schedule (with deadlines looming) to get this merged so that I have a week or so to deal with any fallout before other tasks consume me. So I'm not willing to push it back more than a day or so. But this is almost the last stage of a long chain, so I think I can safely finish it off tomorrow.

timholy:
> but I do wonder how many downstream packages are going to break from this.

Lots and lots. It will be as bad as a new julia release.

timholy:
For potential reviewers, it's worth emphasizing that huge chunks of functionality were moved to ImageCore, ImageFiltering, AxisArrays, etc---these aren't wholesale deletions with no replacement.

tkelman:
If they're going to break that badly, then my inclination would be to upper-bound them so the breakage is opt-in, rather than *users* of the downstream packages suddenly having code stop working on a `Pkg.update()`.

timholy:
That would be fine, if you really think you can handle the resolver. As you know, I've always been skeptical that without systematic addition of upper bounds to all packages, cherry-picking a few cases where we notice a problem is likely to cause more trouble than it solves (the resolver will find some way to "escape" constraints that ends up making very different choices than you'd want on your own). But you're very good at this kind of stuff, so if you think you can handle it, in principle it's a great idea.

tkelman:
We'll take it on a case by case basis. The resolver surprises me occasionally but I've mostly got it figured out. I finally have a reproducible test case to isolate and hopefully fix that bad-error-message bug.

Most of Images' dependent packages have had far fewer releases, and more uniform dependencies over time, than Images itself has. Though I do wish pkgeval were actually running to completion lately - ever since the string `.data` change it has frozen on nightly every single day.

dependents (including some test-only) of Images:
- ActiveAppearanceModels has never passed tests on julia 0.5
- AffineTransforms passes tests on this branch (with some warnings, this is your package so I assume you'll take care of it)
- AndorSIF passes tests on this branch
- ColorSchemes passes on Images 0.5.14, fails on this branch - I'll add bounds
- Drawing has never passed tests on julia 0.5
- EEG passes tests on this branch (with some warnings, cc @codles)
- FaceDatasets passes tests on this branch (with some warnings, cc @dfdx)
- GLPlot and GLVisualize are hard for anyone who isn't @SimonDanisch to test
- ImageMagick passes tests on this branch (did it have some existing ColorTypes deprecations on v0.1.8?)
- ImageProjectiveGeometry passes tests on this branch (with a bunch of warnings, cc @peterkovesi)
- ImageQuilting passes on Images 0.5.14, fails on this branch - I'll add bounds
- ImageRegistration has never passed tests on julia 0.5
- ImageView and ImagineFormat pass on Images 0.5.14, fail on this branch - I assume you'll get these, but may be worth bounds anyway
- Immerse passes tests on this branch (with some warnings, assuming you'll take care of it - are you still using/maintaining this?)
- Knet passes tests on this branch
- LaTeX passes tests on this branch
- Mads had some precompilation issues, not sure (cc @montyvesselinov, you may want to test against this branch yourself)
- MatlabCompat has never passed tests on julia 0.5
- Netpbm passes tests on this branch
- NRRD passes tests on this branch
- OpenGL hasn't been touched in years
- ParallelAccelerator passes tests on this branch
- PerceptualColourMaps passes tests on this branch (with some warnings, cc @peterkovesi)
- PGFPlots had some latex install issues, not sure (cc @mykelk @tawheeler, you may want to test against this branch yourself)
- PiecewiseAffineTransforms passes tests on this branch
- Plots doesn't pass on either, dunno what's wrong there
- Pyramids passes tests on this branch (with a bunch of warnings, cc @loganwilliams)
- QuartzImageIO isn't (legally) testable on anything I own, but has been discussed elsewhere
- TestImages passes tests on this branch
- VideoIO passes on Images 0.5.14, fails on this branch - I'll add bounds
- VisualRegressionTests passes tests on this branch (with some warnings, cc @tbreloff)
- VLFeat has never passed tests on julia 0.5 releases
- Weber wasn't testable on Linux (cc @haberdashPI, you may want to test against this branch yourself)

So a handful of failures (you may want to look at those, I didn't keep logs but I could re-run and gist them if you can't reproduce yourself), a fair number of warnings.

timholy:
I should also provide a little context for the effort of updating. In three of our private repos, one with ~5k LOS/~1.3LOT (LOS=lines of `src/`, LOT = lines of `test/`not including tests) only needed 7 lines of updating (to fix both errors and depwarns). Another with another ~5k LOS/~1.5k LOT needed 24, and a third with ~1k LOS/400 LOT needed ~30 lines updated.

rsrock:
Let me see what I can do about QuartzImageIO.  I'll try to check out this branch now and see how it works for me.

rsrock:
Pretty good progress today on QuartzImageIO. I have all of the image loading tests passing, and I managed to pound one of the saving tests into submission.

src/algorithms.jl
```diff
@@ -89,53 +17,85 @@ function _meanfinite{T<:AbstractFloat}(A::AbstractArray, ::Type{T}, region)
...
+    val1 = var(view(imgc, first(inds1), colons...); kwargs...)
+    vals = similar(imgc, typeof(val1), inds1)
+    vals[1] = val1
+    for i in first(inds1)+1:last(inds1)
+        vals[i] = var(view(imgc, i, colons...); kwargs...)
+    end
+    base_colorant_type(C)(vals...)
 end
 
+Base.std{C<:Colorant}(A::AbstractArray{C}; kwargs...) = mapc(sqrt, var(A))
```
vchuravy: should `kwargs...` be passed through to `var`? And maybe also add a note about what the kwargs are used for.

src/algorithms.jl
```diff
@@ -147,11 +107,14 @@ function _log(kind::Symbol)
...
 The base b of the logarithm (a.k.a. entropy unit) is one of the following:
-  `:shannon ` (log base 2, default)
-  `:nat` (log base e)
-  `:hartley` (log base 10)
+
+- `:shannon ` (log base 2, default)
+- `:nat` (log base e)
+- `:hartley` (log base 10)
 """
 function entropy(img::AbstractArray; kind=:shannon)
```
vchuravy: this might be more efficient by splitting up the entropy function into

`entropy{Log<:Function}(log::Log, img) = ...`
and `entropy(img; kind=:shannon) = entropy(_log(kind), img)`

so that the cost of dynamic dispatch is payed only once.

src/algorithms.jl
```diff
@@ -520,581 +379,91 @@ function test_approx_eq_sigma_eps{T<:Real}(A::AbstractArray, B::AbstractArray,
...
-                    A[i] -= a1*A[i-1] + a2*A[i-2] + a3*A[i-3]
-                end
-                copytail!(x, A, coloffset, 1, n1)
-                A_mul_B!(vstart, M, x)
-                A[n1+coloffset] = vstart[1]
-                A[n1-1+coloffset] -= a1*vstart[1]   + a2*vstart[2] + a3*vstart[3]
-                A[n1-2+coloffset] -= a1*A[n1-1+coloffset] + a2*vstart[1] + a3*vstart[2]
-                for i = n1-3+coloffset:-1:1+coloffset
-                    A[i] -= a1*A[i+1] + a2*A[i+2] + a3*A[i+3]
+See also: BlobLoG.
```
vchuravy: I think you can do ```[`BlobLoG`](@ref)``` to generate a reference

src/algorithms.jl
```diff
@@ -1115,103 +484,86 @@ Like `findlocalmaxima`, but returns the coordinates of the smallest elements.
...
-    elseif border == "symmetric"
-        I = [1:M; M:-1:1][1 .+ mod.(I .- 1, 2 * M)]
-    elseif border == "reflect"
-        I = [1:M; M-1:-1:2][1 .+ mod.(I .- 1, 2 * M - 2)]
-    else
-        error("unknown border condition")
-    end
-    I
-end
+# """
```
vchuravy: If this is no longer required the whole section of code should be deleted

timholy:
@rsrock, thanks a million! If there's any way I can help, don't hesitate to ask!

timholy:
Thanks for a very helpful review, @vchuravy. I updated docstrings in ImageCore, ImageMetadata, and ImageFiltering as well to include the cross links. Nice usability enhancement!

> What is your plan with regards to tests?

For now I wanted to make sure that old code worked and issued appropriate deprecation warnings. But obviously we'll need to transfer the old tests to new ones. I was wondering if that should wait until we're getting close to wanting to delete the deprecations, since if we just yank them soon then there's always the risk that we'll break the deprecation. Certainly, any new tests should use Base.Test rather than FactCheck.

timholy:
That felt good.

vchuravy:
> That felt good.

Congratulations! :tada:

tkelman:
Worth adding bounds for ImageView and ImagineFormat?

timholy:
ImageView is already updated. (Part of my mistaken direct-push to METADATA.jl). Don't worry about ImagineFormat, there's already a branch that folks can use. Just mulling over whether I should make more changes first.

tkelman:
It's already updated if people are using the latest version of ImageView, but if they try to use any older versions of it against the latest Images it isn't going to work.

timholy:
Then sure, it would be fine. Thanks again!

tkelman:
Oh, hah, you already did -
https://github.com/JuliaLang/METADATA.jl/tree/metadata-v2/ImageView/versions
https://github.com/JuliaLang/METADATA.jl/tree/metadata-v2/ImagineFormat/versions

the failures were just because there wasn't a new tagged version to test from yet.

rsrock:
Quick question here. Is that how you add an upper bound to already tagged and published versions, by editing the REQUIRES files in METADATA?  I'll have to do the same thing soon for QuartzImageIO.

tkelman:
Yep. I can help you through it if you need any help.

montyvesselinov:
A simple load of a PNG file fails:

```
img = Images.load("any_test.png")

Errors encountered while loading "/Users/monty/test.png".
All errors:
   ErrorException("\"colorspace\" property is ignored, now color is encoded only by the element type.\nSee `colorview` to represent a numeric array as a color array.")
   FileIO.NotInstalledError(:ImageMagick,"")
Fatal error:
ERROR: "colorspace" property is ignored, now color is encoded only by the element type.
See `colorview` to represent a numeric array as a color array.
 in checked_import(::Symbol) at /Users/monty/.julia/v0.5/FileIO/src/loadsave.jl:12
 in #load#17(::Array{Any,1}, ::Function, ::FileIO.File{FileIO.DataFormat{:PNG}}) at /Users/monty/.julia/v0.5/FileIO/src/loadsave.jl:83
 in load(::FileIO.File{FileIO.DataFormat{:PNG}}) at /Users/monty/.julia/v0.5/FileIO/src/loadsave.jl:75
 in #load#13(::Array{Any,1}, ::Function, ::String) at /Users/monty/.julia/v0.5/FileIO/src/loadsave.jl:45
 in load(::String) at /Users/monty/.julia/v0.5/FileIO/src/loadsave.jl:45
```

timholy:
If you're on mac, for now first say `using ImageMagick`. (Thanks to @rsrock, QuartzImageIO is almost fixed, https://github.com/JuliaIO/QuartzImageIO.jl/pull/29.)

rsrock:
That's right.  I'm taking the chance to clean up a few more bits, but hopefully the fixed QuartzImageIO will land today.

montyvesselinov:
```
Pkg.add("ImageMagick")
```

resolved all the issues. Thank you very much!

PR #578, opened 2017-01-30
Remove Zlib from REQUIRE file

ranjanan:
Why is Zlib on the REQUIRE file? I cannot find it in `src/`


timholy:
Retesting with the new framework.

PR #580, opened 2017-02-01
Another implementation of bwdist (feature_transform/euclidean_distance_transform)

timholy:
This revamps #576, which revamps #99. While I was reluctant to wade into this, given that two people (@wkearn and @muhlba91) have both put significant effort into this, I decided I could no longer avoid respecting their heroism. To make sure everyone got due credit, I first rebased #99 on master (assigning credit to @wkearn), then "replayed" #576 on top of that (assigning credit to @muhlba91), and then added my own tweaks.

I almost merged #576 directly (after resolving the merge conflicts), but after profiling it and discovering that indeed `ind2sub` accounted for most of the runtime, I couldn't resist investigating. More importantly, I decided that a return type of `Array{CartesianIndex}` might be more useful to users than `Array{Int}` (where users would have to call `ind2sub` on the output). Since it's difficult to deprecate output-types, I felt it would be better to rework the API before merging.

This splits the former `bwdist` into two functions, `feature_transform` and `distance_transform`. If all you want are the positions, just call `feature_transform`; if you need the distances you call `distance_transform` on the output of `feature_transform`. It also implements support for anisotropic voxels, which might be important for 3d imaging where often the spacing among slices is different than the pixel-spacing within a slice.

The performance is now quite good:
```julia
julia> A = rand(Bool, 500, 500);

julia> using BenchmarkTools

julia> @benchmark feature_transform($A) seconds=1
BenchmarkTools.Trial: 
  samples:          70
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  3.85 mb
  allocs estimate:  11
  minimum time:     13.01 ms (0.00% GC)
  median time:      13.73 ms (0.00% GC)
  mean time:        14.39 ms (0.47% GC)
  maximum time:     21.29 ms (2.09% GC)
```
which is an improvement by a factor of 5 over #576, and a factor of ~40 over #99.

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -65,6 +65,7 @@ include("edge.jl")
 include("showmime.jl")
 include("corner.jl")
 include("distances.jl")
+include("bwdist.jl")
 include("deprecated.jl")
 
 export # types
@@ -157,6 +158,7 @@ export # types
     imgaussiannoise,
     imlineardiffusion,
     imROF,
+    bwdist,
 
     #Exposure
     complement,
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -66,6 +66,7 @@ include("showmime.jl")
 include("corner.jl")
 include("distances.jl")
 include("bwdist.jl")
+using .FeatureTransform
 include("deprecated.jl")
 
 export # types
@@ -158,7 +159,8 @@ export # types
     imgaussiannoise,
     imlineardiffusion,
     imROF,
-    bwdist,
+    feature_transform,
+    distance_transform,
 
     #Exposure
     complement,
@@ -233,7 +235,7 @@ Algorithms:
     - Edge detection: `imedge`, `imgradients`, `thin_edges`, `magnitude`, `phase`, `magnitudephase`, `orientation`, `canny`
     - Corner detection: `imcorner`, `harris`, `shi_tomasi`, `kitchen_rosenfeld`, `meancovs`, `gammacovs`, `fastcorners`
     - Blob detection: `blob_LoG`, `findlocalmaxima`, `findlocalminima`
-    - Morphological operations: `dilate`, `erode`, `closing`, `opening`, `tophat`, `bothat`, `morphogradient`, `morpholaplace`
+    - Morphological operations: `dilate`, `erode`, `closing`, `opening`, `tophat`, `bothat`, `morphogradient`, `morpholaplace`, `feature_transform`, `distance_transform`
     - Connected components: `label_components`, `component_boxes`, `component_lengths`, `component_indices`, `component_subscripts`, `component_centroids`
     - Interpolation: `bilinear_interpolation`
 
--- a/src/bwdist.jl
+++ b/src/bwdist.jl
@@ -1,0 +1,156 @@
+using Base.Cartesian
+
+# A function to permute a tuple of subscripts given a
+# permutation vector
+function permutesubs(subs::(Int...),perm::Vector{Int})
+    n = length(subs)
+    ntuple(n,x->subs[perm[x]])
+end
+
+# A function to both permute the dimensions of an array and
+# those of the linear indices stored in that array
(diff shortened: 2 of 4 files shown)


timholy:
Oh, and by the way: this is a nice example of the power of the CartesianIndex framework (and might make a nice study for anyone wishing to learn how to use it more effectively). If you look at the code and compare against Maurer et al, you'll see it's almost a line-for-line translation of their pseudocode (which involves arbitrary-dimensional nested loops). Read it and weep, other programming languages... :wink:

timholy:
@mulbha91, I'd rather wait to hear from you whether this seems OK with you.

Also, in crediting 65bbaca89f35a6804cd0b72d5aad3cbe10f52052 to you, GitHub seems not to recognize your email or something... I notice that in #576 your commits have also become de-linked from your account. If there's a different email you'd want me to associate this commit with, I can revamp it. Or use the GitHub interface to associate that email with you.

timholy:
Messed up on pinging @muhlba91

muhlba91:
Hi!

Sorry for my late response but I got quite busy the past days...

However, I totally agree that this is a perfect example for `CartesianIndex`. 👍 
(To be honest, I wasn't familiar with those at all since it's also the first time I'm working with Julia - so I was already happy about the progress I made. ^^)

I was looking through your code and tested it and am totally happy with it - amazing how those `CartesianIndex` work - thank you! I'd say let's merge it to master! :)

timholy:
I thought you did great with #576, all the more so for someone new to Julia! Hope to see you around more :smile:.

Before I hit merge, we should get that commit linked to your email. If you visit https://github.com/JuliaImages/Images.jl/pull/580/commits/65bbaca89f35a6804cd0b72d5aad3cbe10f52052, you'll notice that next to your name there's a ? mark. If you click it, you might be taken to a page that lets you register that email. Or alternatively, I can edit the commit authorship to a different email address. I don't think it was just me that caused this, as all your commits in #576 are also now unlinked from your account.

muhlba91:
Thank you! :)

OK, I see. If you don't mind I'd be happy if you could add the authorship with daniel@muehlbachler.org ? ;)

timholy:
That worked. Once this passes CI I'll merge. Thanks again for the contribution!

chadagreene:
Well I see I'm pretty late to this party. 

New Julian here, coming from Matlab, where I've spent years relying heavily on the `bwdist` function. Today I had a dickens of a time figuring out that the `bwdist` functionality requires this sequence of functions, whose names are somewhat vague and unmemorable. Even knowing the names of the `feature_transform` and `distance_transform` functions, the documentation really doesn't spell out how to combine them to calculate the Euclidean distance to true pixels in a binary mask. 

Any chance we could get a `bwdist` function that would combine the two transformation functions? If not, would it be possible to add an example like this to the `distance_transform` function header? 

```julia 
# Create an example binary mask: 
I = zeros(Bool,(10,10));
I[[13:16;24:26]] .= true;
 
# Calculate Euclidean distance to true pixels:
D = distance_transform(feature_transform(I))
10×10 Matrix{Float64}:
 2.23607  2.0  2.23607  2.82843  3.60555  …  5.83095  6.7082   7.61577
 1.41421  1.0  1.41421  2.23607  2.82843     5.38516  6.32456  7.28011
 1.0      0.0  1.0      1.41421  2.23607     5.09902  6.08276  7.07107
 1.0      0.0  0.0      1.0      2.0         5.0      6.0      7.0
 1.0      0.0  0.0      1.0      2.0         5.0      6.0      7.0
 1.0      0.0  0.0      1.0      2.0      …  5.0      6.0      7.0
 1.41421  1.0  1.0      1.41421  2.23607     5.09902  6.08276  7.07107
 2.23607  2.0  2.0      2.23607  2.82843     5.38516  6.32456  7.28011
 3.16228  3.0  3.0      3.16228  3.60555     5.83095  6.7082   7.61577
 4.12311  4.0  4.0      4.12311  4.47214     6.40312  7.2111   8.06226
```

PR #581, opened 2017-02-03
Update test REQUIRE

rsrock:
Might as well wait for the QuartzImageIO tag 

JuliaLang/METADATA.jl#7814


timholy:
Woot! Tagging another release now.

PR #589, opened 2017-02-13
WIP: Move restrict and imresize to ImageTransformations

Evizero:
**NOT MERGE READY**

The branch corresponding to the move of `restrict` and `imresize` to ImageTransformations.jl.
This depends on ImageTransformations.jl being registered in METADATA.

see: https://github.com/JuliaImages/ImageTransformations.jl/pull/4

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,6 +4,7 @@ Colors 0.7.0
 ColorVectorSpace 0.2
 FixedPointNumbers 0.3.0
 ImageCore
+ImageTransformations
 ImageFiltering
 AxisArrays
 ImageAxes
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -47,10 +47,12 @@ using IndirectArrays, MappedArrays
 const is_little_endian = ENDIAN_BOM == 0x04030201
 
 @reexport using ImageCore
+@reexport using ImageTransformations
 @reexport using ImageAxes
 @reexport using ImageMetadata
 @reexport using ImageFiltering
 
+import ImageTransformations: restrict
 using ImageMetadata: ImageMetaAxis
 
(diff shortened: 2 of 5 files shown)


Evizero:
Seems like the refactor of `imresize` broke the tests for the `gaussian_pyramid` here: https://github.com/JuliaImages/Images.jl/blob/master/test/algorithms.jl#L202-L210

timholy:
Turns out to be because, for a 2-fold reduction, the old one started at `imgr[1,1] = imgo[1,1]` and the new one starts at `imgr[1,1] = imgo[2,2]`. I don't think either one is the best choice, see https://github.com/JuliaImages/ImageTransformations.jl/pull/8.

timholy:
BTW, I have a local branch that builds off this and fixes those tests (plus another issue I observed with `gaussian_pyramid`), so you don't need to do more. Thanks for tackling this!

Evizero:
Thanks for looking into this. How shall we proceed from here?

timholy:
I'm finding some other bugs that my changes triggered, so let me squash those first.

As far as your involvement, perhaps review JuliaImages/ImageTransformations.jl#8 and see if you can spot any problems? I won't merge it until I know it plays nicely with my local changes in Images, though; I just found a bug in that work (fix just pushed moments ago).

Once that's merged and tagged, I'll push my branch that address the issues in Images. Since it's based off of your branch here, I suspect this PR will close automatically then. Let's leave this open until it's really solved.

timholy:
OK, I'm happy locally. I decided to submit it as a PR (# ), will restart the CI tests after JuliaImages/ImageTransformations.jl#8 merges and is tagged (so you can review that one anytime now).

PR #597, opened 2017-02-28
added convex hull algorithm

tejus-gupta:
The method convexhull computes the convex hull of a binary image using Graham scan algorithm. The method can return either the boundary of the convex hull or the filled out convex hull. I have also added some tests for convexhull.

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -162,7 +162,7 @@ export # types
     imROF,
     feature_transform,
     distance_transform,
-    convexhull
+    convexhull,
 
     #Exposure
     complement,
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -68,6 +68,7 @@ include("distances.jl")
 include("bwdist.jl")
 using .FeatureTransform
 include("deprecated.jl")
+include("convexhull.jl")
 
 export # types
     BlobLoG,
@@ -161,6 +162,7 @@ export # types
     imROF,
     feature_transform,
     distance_transform,
+    convexhull
 
     #Exposure
     complement,
@@ -235,7 +237,7 @@ Algorithms:
     - Edge detection: `imedge`, `imgradients`, `thin_edges`, `magnitude`, `phase`, `magnitudephase`, `orientation`, `canny`
     - Corner detection: `imcorner`, `harris`, `shi_tomasi`, `kitchen_rosenfeld`, `meancovs`, `gammacovs`, `fastcorners`
     - Blob detection: `blob_LoG`, `findlocalmaxima`, `findlocalminima`
-    - Morphological operations: `dilate`, `erode`, `closing`, `opening`, `tophat`, `bothat`, `morphogradient`, `morpholaplace`, `feature_transform`, `distance_transform`
+    - Morphological operations: `dilate`, `erode`, `closing`, `opening`, `tophat`, `bothat`, `morphogradient`, `morpholaplace`, `feature_transform`, `distance_transform`, `convexhull`
     - Connected components: `label_components`, `component_boxes`, `component_lengths`, `component_indices`, `component_subscripts`, `component_centroids`
     - Interpolation: `bilinear_interpolation`
 
--- a/src/convexhull.jl
+++ b/src/convexhull.jl
@@ -46,6 +46,14 @@ function convexhull{T<:Images.Gray}(img::AbstractArray{T, 2}, returntype="points
         error("Invalid argument returntype")
     end
 
+    for i in CartesianRange(size(img))
+        if img[i]!=0 && img[i]!=1
+            warn("Input image isn't binary!")
+            warn("Function considers pixel intensity<255 as black")
+            break
+        end
+    end
+
     function getboundarypoints{T<:Images.Gray}(img::AbstractArray{T, 2})
 
         points = CartesianIndex{2}[]
--- a/src/convexhull.jl
+++ b/src/convexhull.jl
@@ -13,38 +13,7 @@ The function can return either of these-
 In case the image isn't a binary image, it considers pixel intensity<255 as black.
 """
 
-function drawline{T<:ColorTypes.Gray}(img::AbstractArray{T, 2}, y0::Int, x0::Int, y1::Int, x1::Int, color)
-    dx = abs(x1 - x0)
-    dy = abs(y1 - y0)
- 
-    sx = x0 < x1 ? 1 : -1
-    sy = y0 < y1 ? 1 : -1;
- 
-    err = (dx > dy ? dx : -dy) / 2
-    
-    while true
-        img[y0, x0] = color
-        (x0 != x1 || y0 != y1) || break
-        e2 = err
-        if e2 > -dx
-            err -= dy
-            x0 += sx
-        end
-        if e2 < dy
-            err += dx
-            y0 += sy
-        end
-    end
-
-    img
-end
-
-
-function convexhull{T<:Images.Gray}(img::AbstractArray{T, 2}, returntype="points")
-
-    if returntype!="points" && returntype!="boundary" && returntype!="filled"
-        error("Invalid argument returntype")
-    end
+function convexhull{T<:Images.Gray}(img::AbstractArray{T, 2})
 
     for i in CartesianRange(size(img))
         if img[i]!=0 && img[i]!=1
@@ -111,44 +80,6 @@ function convexhull{T<:Images.Gray}(img::AbstractArray{T, 2}, returntype="points
         end
     end
 
-    function drawboundary(convex_hull, img)
-        chull = zeros(T, size(img))
-        for point in convex_hull
-            chull[point]=1
-        end
-
-        for i in 1:length(convex_hull)-1
-            chull=drawline(chull, convex_hull[i][1], convex_hull[i][2], convex_hull[i+1][1], convex_hull[i+1][2], 1)
-        end
-        chull=drawline(chull, convex_hull[1][1], convex_hull[1][2], convex_hull[end][1], convex_hull[end][2], 1)
-
-        return chull
-    end
-
-    function fillhull(convex_hull, img)
-        chull = zeros(T, size(img))
-        for point in convex_hull
-            chull[point]=1
-        end
-
-        for i in 1:length(convex_hull)-1
-            chull=drawline(chull, convex_hull[i][1], convex_hull[i][2], convex_hull[i+1][1], convex_hull[i+1][2], 1)
-        end
-        chull=drawline(chull, convex_hull[1][1], convex_hull[1][2], convex_hull[end][1], convex_hull[end][2], 1)
-
-        for i in indices(img, 1)
-            ones = find(chull[i,:] .==1)
-            if length(ones) >= 2
-                first = CartesianIndex{2}(i, ones[1])
-                last = CartesianIndex{2}(i, ones[end])
-                chull=drawline(chull, first[1], first[2], last[1], last[2], 1)
-            end
-        end
-
-        return chull
-    end
-
-
     # Used Graham scan algorithm
 
     points = getboundarypoints(img)
@@ -178,12 +109,6 @@ function convexhull{T<:Images.Gray}(img::AbstractArray{T, 2}, returntype="points
         push!(convex_hull, points[i])
     end
 
-    if returntype=="points"
-        return convex_hull
-    elseif returntype=="boundary"
-        return drawboundary(convex_hull, img)
-    elseif returntype=="filled"
-        return fillhull(convex_hull, img)
-    end
+    return convex_hull
 
 end
--- a/src/convexhull.jl
+++ b/src/convexhull.jl
@@ -1,14 +1,8 @@
 """
 ```
 chull = convexhull(img)
-chull = convexhull(img, "points/boundary/filled")
 ```
-
-Computes the convex hull of a binary image.
-The function can return either of these-
-- vector of vertices of the convex hull
-- image with boundary pixels marked
-- image with convex hull filled
+Computes the convex hull of a binary image and returns the vertices of convex hull as a CartesianIndex array.
 
 In case the image isn't a binary image, it considers pixel intensity<255 as black.
 """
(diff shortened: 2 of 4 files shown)


mronian:
Would this be more suited for ImageDraw.jl?

src/convexhull.jl
```diff
@@ -0,0 +1,182 @@
+"""
+```
+chull = convexhull(img)
+chull = convexhull(img, "hull boundary/filled hull")
+```
+
+Computes the convex hull of a binary image.
+"""
+
+function drawline{T<:ColorTypes.Gray}(img::AbstractArray{T, 2}, y0::Int, x0::Int, y1::Int, x1::Int, color)
```
mronian: You could use ```line2d``` instead of this 😄

src/convexhull.jl
```diff
@@ -0,0 +1,182 @@
...
+            err += dx
+            y0 += sy
+        end
+    end
+
+    img
+end
+
+
+function convexhull{T<:Images.Gray}(img::AbstractArray{T, 2}, returntype="hull boundary")
```
mronian: You could pass a function name for the return type required i.e. ```boundary``` and ```filled```. We did this for [```imcorner```](https://github.com/JuliaImages/Images.jl/blob/master/src/corner.jl).

test/algorithms.jl
```diff
@@ -438,6 +438,41 @@ using Base.Test
...
+        @test size(B)==size(A)
+        @test typeof(B)==typeof(A)
+        @test B[25,1]==1
+        @test B[1,10]==1
+        @test B[10,10]==1
+        @test B[10,30]==1
+        @test B[40,30]==1
+        @test B[40,10]==1
+        @test B[50,10]==1
+
```
mronian: You should also add tests which check whether the inside is actually filled for the `filled hull` case and similarly for the other one. This could be done by taking points forming a simple shape like a square/pentagon and then checking if it got filled or not.

src/convexhull.jl
```diff
@@ -0,0 +1,182 @@
...
+        if last_point_found
+            push!(points, last_point)
+        end
+    end
+
+    # Used Graham scan algorithm
+
+    last_point=[-1, -1]
+    for i in 1:h
+        for j in 1:w
```
mronian: Check out [CartesianRange](http://julialang.org/blog/2016/02/iteration) for indexing. Its awesome ✌️

src/convexhull.jl
tejus-gupta: At first I had used line2d. But I found that ImageDraw package isn't a registered package and so can't be installed using package manager. Then I added the function so that convexhull can be used without ImageDraw as a dependency.
Should I remove it and use line2d?

test/algorithms.jl
tejus-gupta: I have added some more tests.

src/convexhull.jl
```diff
@@ -0,0 +1,182 @@
...
+
+    if returntype!="hull boundary" && returntype!="filled hull"
+        error("Invalid argument returntype")
+    end
+
+    points = []
+    h, w = size(img)
+
+    #just store first and last point in each row for computing convex hull
+    for i in 1:h
```
timholy: Julia now allows arrays to start at indices other than 1. How about `for i in indices(img, 1)`? (Similar changes elsewhere.)

src/convexhull.jl
```diff
@@ -0,0 +1,182 @@
...
+            push!(points, first_point)
+        end
+        if last_point_found
+            push!(points, last_point)
+        end
+    end
+
+    # Used Graham scan algorithm
+
+    last_point=[-1, -1]
```
timholy: Here, to be tolerant of images with non-1 indices you could use `last_point = CartesianIndex(map(r->first(r)-1, indices(img))` or something.

src/convexhull.jl
```diff
@@ -0,0 +1,182 @@
...
+
+    function right_oriented(ref, a, b)
+        ((a[2]-ref[2])*(b[1]-ref[1])-(b[2]-ref[2])*(a[1]-ref[1])<0) ? true : false
+    end
+
+    function collinear(a, b, c)
+        ((a[2]-c[2])*(b[1]-c[1])-(b[2]-c[2])*(a[1]-c[1])==0) ? true : false
+    end
+
+    for i in 1:size(points)[1]
```
timholy: `length(points)`? Am I right that It's a vector? (Here and below.)

src/convexhull.jl
```diff
@@ -0,0 +1,182 @@
...
+        if(points[i]==last_point)
+            deleteat!(points, i)
+            break
+        end
+    end
+
+    #angular sort wrt last_point
+    sort!(points, lt=right_oriented)
```
timholy: It's possible this would be more efficient with `lt=(a, b) -> right_oriented(last_point, a, b)`. Not sure, but perhaps worth testing.

src/convexhull.jl
```diff
@@ -0,0 +1,182 @@
...
+    #if polar angle of 2 points wrt last_point is same, just keep farther point
+    i=0
+    while i<=size(points)[1]
+        i=i+1
+        if i+1>size(points)[1]
+            break
+        end
+
+        if(collinear(last_point,points[i], points[i+1]))
+            if (points[i][1]-last_point[1])^2 + (points[i][2]-last_point[2])^2< (points[i+1][1]-last_point[1])^2 + (points[i+1][2]-last_point[2])^2
```
timholy: More readable if you define `dist2(a, b)`. If `a` and `b` are `CartesianIndex`, then
```julia
dist2(a, b) = sum(abs2, (a-b).I))
```
should work (and automatically generalizes to N dimensions).

src/convexhull.jl
```diff
@@ -0,0 +1,182 @@
...
+        end
+    end
+
+    n_points=size(points)[1]
+
+    if n_points<3
+        error("Not enough points to compute convex hull.")
+    end
+
+    convex_hull=[]
```
timholy: Here too, specify the element type.

src/convexhull.jl
```diff
@@ -0,0 +1,182 @@
...
+        while (right_oriented(convex_hull[end],convex_hull[end-1], points[i]))
+            pop!(convex_hull)
+        end
+        push!(convex_hull, points[i])
+    end
+
+
+    chull = zeros(T, h, w)
+    for i = 1:size(convex_hull)[1]
+        chull[convex_hull[i][1], convex_hull[i][2]]=1
```
timholy: If you use `CartesianIndex` this is just `chull[convex_hull[i]] = 1`.

tejus-gupta:
I have used CartesianIndex in my last commit. Presently the function has options to return the convex hull vertices, boundary and filled hull. I have also made the code more readable by adding more functions and restructuring the code.

timholy:
> Presently the function has options to return the convex hull vertices, boundary and filled hull.

I'd rather see the user make those function calls manually, rather than running them depending on the value of a string option. The style of programming you're using here is common among converts from languages like Matlab, where each function means creating a separate file. It's annoying to have too many files, so people try to cram a bazillion options into what should otherwise be simple functions. In a language like Julia, this is simply not a concern, and you should write functions that "do one thing." If it's a different thing, it should be a second function.

It's not just a stylistic thing, it's actively bad:
- more options makes functions more complicated; they need extra arguments and then mechanisms to supply defaults. That's OK when you need them, but here you manifestly don't.
- it makes testing harder; it's easiest to write tests for functions that do simple things.
- it destroys type inference, because the return type depends on the *value* of an argument (here, a string). 

The last point means that any function that calls `convexhull` will, from that point forward, have poor performance.

tejus-gupta:
I have updated the implementation. The function only returns the convex hull vertices as a CartesianIndex array.

tejus-gupta:
I have 2 questions:
1. I will add the functions for draw boundary and fill polygon to ImageDraw.jl. Is that alright?

2. Currently the function interprets all pixel intensity<255 as black and gives a warning if the image is not binary(0/255). Would it be better to only accept boolean inputs?

timholy:
> I have updated the implementation. The function only returns the convex hull vertices as a CartesianIndex array.

Great, thanks. If you now think the name `convexhull` is a misnomer, you could call the core algorithm `convexhull_vertices`.

> I will add the functions for draw boundary and fill polygon to ImageDraw.jl. Is that alright?

That would be perfect! Thanks for tackling this. We can then register and release ImageDraw as a registered package.

> Currently the function interprets all pixel intensity<255 as black and gives a warning if the image is not binary(0/255). Would it be better to only accept boolean inputs?

I think so.

src/convexhull.jl
```diff
@@ -0,0 +1,108 @@
...
+
+    for i in CartesianRange(size(img))
+        if img[i]!=0 && img[i]!=1
+            warn("Input image isn't binary!")
+            warn("Function considers pixel intensity<255 as black")
+            break
+        end
+    end
+
+    function getboundarypoints{T<:Images.Gray}(img::AbstractArray{T, 2})
```
timholy: Maybe try it like this (I picked a different name from yours just so I could try both, I'm fine with `getboundarypoints`):
```julia
function findends(A)
           points = CartesianIndex{2}[]
           for j = indices(A, 2)
               v = view(A, :, j)
               i1 = findfirst(v)
               if i1 != 0
                   i2 = findlast(v)
                   push!(points, CartesianIndex(i1, j))
                   if i1 != i2
                       push!(points, CartesianIndex(i2, j))
                   end
               end
           end
           points
       end
```
For a 1000x1000 random Bool array, I get this:
```julia
julia> @time getboundarypoints(A);
  0.015949 seconds (11.50 k allocations: 9.587 MB, 37.24% gc time)

julia> @time findends(A);
  0.000298 seconds (1.01 k allocations: 111.563 KB)
```

Notice it's better to make the inner loop be on columns, since those are stored adjacently in memory; when you iterate over rows, you pay a much larger price in terms of cache efficiency. See http://julialang.org/blog/2013/09/fast-numeric.

src/convexhull.jl
```diff
@@ -0,0 +1,108 @@
+"""
+```
+chull = convexhull(img)
+```
+Computes the convex hull of a binary image and returns the vertices of convex hull as a CartesianIndex array.
+
+In case the image isn't a binary image, it considers pixel intensity<255 as black.
+"""
+
+function convexhull{T<:Images.Gray}(img::AbstractArray{T, 2})
```
timholy: FYI `Gray` is defined in ColorTypes, not Images, and also exported by both Images and Colors. In Images itself you don't have to scope it this way. (I know you're going to change this to `Bool` anyway, but I just thought I'd mention it.)

src/convexhull.jl
```diff
@@ -0,0 +1,108 @@
...
+            elseif length(ones)!=0
+                push!(points, CartesianIndex{2}(i, minimum(ones)))
+                push!(points, CartesianIndex{2}(i, maximum(ones)))
+            end
+        end
+        return points
+    end
+
+    function right_oriented(ref, a, b)
+        ((a[2]-ref[2])*(b[1]-ref[1])-(b[2]-ref[2])*(a[1]-ref[1])<0) ? true : false
```
timholy: You could get rid of the `? true : false` and just return the expression before the `?`

src/convexhull.jl
```diff
@@ -0,0 +1,108 @@
...
+    end
+
+    function collinear(a, b, c)
+        ((a[2]-c[2])*(b[1]-c[1])-(b[2]-c[2])*(a[1]-c[1])==0) ? true : false
```
timholy: likewise here

test/algorithms.jl
```diff
@@ -438,6 +438,44 @@ using Base.Test
...
+        A= convert(Array{Images.Gray}, A)
+        A[25,1]=1
+        A[1,10]=1
+        A[10,10]=1
+        A[10,30]=1
+        A[40,30]=1
+        A[40,10]=1
+        A[50,10]=1
+        B = convexhull(A)
+        C = CartesianIndex{}[]
```
timholy: Better to say `CartesianIndex{2}[]` but then you can drop the `{}` on the statements below:
```julia
julia> using Base.Test

julia> @inferred(CartesianIndex(1,2))
CartesianIndex{2}((1,2))
```
Using `@inferred` this way (from `Base.Test`) is an easy way to test whether you're writing inferable (type-stable) code.

tejus-gupta:
ana
I decomposed convexhull into functions to check if there are any unusual large memory allocations.
aa() -> find last_point and delete from points
bb() -> sort points
cc() -> for points with same angle, keep only farther one
dd() -> graham scan algorithm

Is the time/memory allocation for sorting unusual?

test/algorithms.jl
tejus-gupta: I have added a test to check if the return type of convexhull can be inferred.

src/Images.jl
```diff
@@ -161,6 +162,7 @@ export # types
     imROF,
     feature_transform,
     distance_transform,
+    convexhull
```
timholy: Hah, turns out this breaks lots of tests (click on the red `x` and read the log on Travis or AppVeyor). You need to add a comma at the end, otherwise so the rest of the exports get omitted. Tricky.

While I sometimes forget, usually before submitting a PR I run the tests locally. Some platforms (particularly, Travis for OSX) get really backed up, and it can be annoying to wait a couple of hours to find out if your changes will pass. That's not a sure-fire test, but it catches many "opportunities" for making mistakes.

tejus-gupta: I have corrected the error. For the above case, the function wasn't removing redundant points when points are collinear. I have corrected that issue.

timholy:
Awesome. Can you add a smaller version of that example to your tests? Any time there's a bug caught it's a good sign of wanting increased test coverage, to make sure it doesn't regress somehow.

Then, I think this is ready to merge.

tejus-gupta:
I have updated a test case to check that.

timholy:
Failures on 0.6 are nothing to worry about. I don't think we can pass tests there until StaticArrays is working on 0.6.

tejus-gupta:
I don’t quite understand your point.

timholy:
If you click on the test status icons, you'll notice that the tests passed (are green) on 0.5 and failed (are red) on 0.6. I'm saying that's nothing you have to worry about; the failures on 0.6 are due to other factors unrelated to your work here.

timholy:
Thanks for a very nice contribution, @tejus-gupta!

tejus-gupta:
@timholy @mronian thanks for code reviews :)

Issue #600, opened 2017-03-02
No implementation of Hough Line Transform


annimesh2809:
I did not find any implementation of hough line transform in julia images. Can I add an implementation of hough line transform similar to that of matlab to julia?

timholy:
That would be great. The one thing you need to be careful of is the license. This package (and all the packages in JuliaImages) are MIT licensed; if you even *read* code that has a restrictive license, you're not allowed to reimplement it using a permissive license. So make sure you check the license status of any code before you start looking at it.

annimesh2809:
Okay thanks, I will be careful about it.

mronian:
@annimesh2809 There is an OpenCV implementation as well. You could use that as a starting point 😄 It would be great to have [other hough transforms](https://en.wikipedia.org/wiki/Hough_transform) (circles etc) as well in ImageFeatures.jl.

timholy:
Closed by https://github.com/JuliaImages/ImageFeatures.jl/pull/28.

PR #601, opened 2017-03-03
Added Hough Line Transform

annimesh2809:
Added a new feature detection : Hough line transform as mentioned in #600 .

The method returns an array of tuples containing (r,theta) as parametric values of lines present in the image.


src/houghtranform.jl
```diff
@@ -0,0 +1,107 @@
...
+            rho::Float64, theta::Float64,
+            min_theta::Float64, max_theta::Float64,
+            threshold::Int64, linesMax::Int64)
+   
+    theta > 0 || error("theta threshold must be positive")
+    rho > 0 || error("radius threshold must be positive")
+    min_theta < max_theta || error("max_theta must be greater than min_theta")
+
+    if ! is_binary_image(img)
+        img = canny(img)
```
timholy: This breaks orthogonality; what if someone wants a different algorithm than `canny`? Will be solved if we simply require a `Bool` eltype.

Hmm, I hadn't noticed that `canny` returns an `Array{Float64}`. We should probably change it to return an `Array{Bool}`.

src/houghtranform.jl
```diff
@@ -0,0 +1,107 @@
...
+            img::AbstractArray{T,2},
+            rho::Float64, theta::Float64,
+            min_theta::Float64, max_theta::Float64,
+            threshold::Int64, linesMax::Int64)
+   
+    theta > 0 || error("theta threshold must be positive")
+    rho > 0 || error("radius threshold must be positive")
+    min_theta < max_theta || error("max_theta must be greater than min_theta")
+
+    if ! is_binary_image(img)
```
timholy: Better to just require an `T<:Union{Bool,Gray{Bool}}` eltype.

src/houghtranform.jl
```diff
@@ -0,0 +1,107 @@
...
+        img = canny(img)
+    end
+
+    width::Int64 = size(img)[2]
```
timholy: No need to declare types like this.

src/houghtranform.jl
```diff
@@ -0,0 +1,107 @@
...
+    min_theta   = minimum theta of lines
+    max_theta   = maximum theta of lines
+    threshold   = no of points to pass through line for considering it valid
+    linesMax    = maximum no of lines to return
+
+"""
+
+function hough_transform_standard{T}(
+            img::AbstractArray{T,2},
+            rho::Float64, theta::Float64,
```
timholy: Too restrictive on the types? Particularly for people who have 32-bit machines, `Int64` is not native (better to use `Int`). `Number` and `Integer` might even be best.

src/houghtranform.jl
```diff
@@ -0,0 +1,107 @@
...
+    min_theta < max_theta || error("max_theta must be greater than min_theta")
+
+    if ! is_binary_image(img)
+        img = canny(img)
+    end
+
+    width::Int64 = size(img)[2]
+    height::Int64 = size(img)[1]
+    irho::Float64 = 1 / rho;
+    numangle::Int64 = round((max_theta - min_theta)/theta)
```
timholy: `round(Int, x)` returns an `Int`. You can use this in several places below, too.

src/houghtranform.jl
```diff
@@ -0,0 +1,107 @@
...
+    numrho::Int64 = round(((width + height)*2 + 1)/rho)
+
+    accumulator_matrix = Array{Int64}(numangle + 2, numrho + 2)
+    fill!(accumulator_matrix, 0)
+
+    #Pre-Computed sines and cosines in tables
+    tabsin = Array{Float64}(numangle)
```
timholy: Even better, `Vector{Float64}` (here and below).

src/houghtranform.jl
```diff
@@ -0,0 +1,107 @@
...
+    tabcos = Array{Float64}(numangle)
+    for i in 1:numangle
+        tabsin[i] = (sin(min_theta + (i-1)*theta) * irho)
+        tabcos[i] = (cos(min_theta + (i-1)*theta) * irho)
+    end
+
+
+    #Hough Transform implementation
+    for pix in CartesianRange(size(img))
+        if img[pix] == 1.0
```
timholy: If you're using `Bool` eltypes, this becomes just `if img[pix]`

src/houghtranform.jl
```diff
@@ -0,0 +1,107 @@
...
+            for i in 1:numangle
+                dist::Int64 = round((pix[1] * tabsin[i] + pix[2] * tabcos[i]))
+                dist += round((numrho -1)/2)
```
timholy: This constant could be hoisted out of the loop.

src/houghtranform.jl
```diff
@@ -0,0 +1,107 @@
...
+                accumulator_matrix[i + 1, dist + 1] += 1
+            end
+        end
+    end
+
+    #Finding local maximum lines
+    validLines = Array{CartesianIndex}(0)
+    for val in CartesianRange(size(accumulator_matrix))
```
timholy: Perhaps we should add a function input to `findlocalmaxima` and just use that?

src/houghtranform.jl
```diff
@@ -0,0 +1,107 @@
...
+            accumulator_matrix[val] > accumulator_matrix[val[1] - 1,val[2]] &&
+            accumulator_matrix[val] >= accumulator_matrix[val[1] + 1,val[2]]
+            push!(validLines,val)
+        end
+    end
+
+    #Sorting by value in accumulator_matrix
+    sort!(validLines, by = (x)->accumulator_matrix[x], rev = true)
+
+    linesMax = min(linesMax, size(validLines)[1])
```
timholy: can just use `length` instead of `size(validLines)[1]` (they're equivalent if `validLines` is a vector).

annimesh2809:
Is there any function available for drawing lines using the normal parameters generated by hough transform? I found the package Cairo.jl but is there something simpler than that.

There is an ImageDraw.jl package and a line function implemented in it. Should I write a function to draw line from normal parameters which uses this existing function?

annimesh2809:
@timholy Yes, I also think it would be better to have it in ImageFeatures.jl instead of having it directly in Images. Should I create a new pull request in that repo?  Also what updates are required in the ImageFeatures.jl package related to the new images framework, maybe I can help :)

timholy:
@annimesh2809, I suspect ImageDraw would indeed be the better way to go...fewer dependencies.

Great, I think ImageFeatures is the best place. With regards to updating ImageFeatures, if you install it and type `using ImageFeatures` you'll already notice a couple of deprecation warnings that should be fixed. The next step would be to run the tests and see if there are more depwarns or breakages. Probably just getting it updated should be a PR on its own, then new stuff can be added.

mronian:
@annimesh2809 It would be great if you work on updating ImageFeatures to the new Images.jl. I have been wanting to do that for a long time but haven't been able to get time. I'd be happy to help in case you need any 😄

annimesh2809:
@mronian I would be really happy to update ImageFeatures.jl. But I did not understand what you meant by update ImgaeFeatures to the new Images.jl

timholy:
The deprecation warnings and test errors should basically tell you what to do.

timholy:
I realized that it may not be obvious that to you "the new Images" *is* what you're running now. There was a big revamp described in #542 and merged in #577. ImageFeatures was written before that work was even started.

mronian:
These links might help -
- https://github.com/JuliaImages/Images.jl/blob/master/NEWS.md
- http://juliaimages.github.io/latest/
- #577

annimesh2809:
I have started working on updating ImageFeatures.jl to work with the new images. But there are some failing test cases without changing any code:

img1
img2

Should I comment out those test cases and create a log of them so that we can correct those test cases later when all the deprecated warnings have been removed?
Also should I create a new issue on ImageFeatures.jl so we can discuss the progress of this update?

mronian:
I guess some changes to Images.jl have modified some values of the descriptor. This should be solved if take an approximate equals instead of equals. Lets close this thread and move it to ImageFeatures.jl.

mronian:
@annimesh2809 Lets continue this discussion at JuliaImages/ImageFeatures.jl#24

mronian:
Whoops. I did not realise this was a PR for hough lines. I thought it was just an issue. I'll reopen it.

annimesh2809:
@mronian Actually we should close this PR because hough lines will be added to ImageFeatures.jl as discussed above once we remove all the errors from that package.

PR #605, opened 2017-03-08
Allow nightly failures, add codecov

rsrock:
Let's get those badges green again!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ os:
 julia:
     - 0.5
     - nightly
+matrix:
+  allow_failures:
+    - julia: nightly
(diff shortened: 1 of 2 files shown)


rsrock:
Looks like this works.  Note that this also switches over to Codecov.io. Any objections to merging?

timholy:
This is great! Merge at will.

timholy:
Once StaticArrays updates for 0.6, hopefully it won't be that much longer before we can support 0.6. I've gotten everything working up through ImageCore, ImageAxes, and ImageMetadata. It's really ImageFiltering that is the problem due to its dependence on StaticArrays.

PR #606, opened 2017-03-15
Changed return type of canny

annimesh2809:
As canny will return a binary image, an `Array{Bool}` would be more memory efficient than `Array{Float64}`.

### DIFF
--- a/src/edge.jl
+++ b/src/edge.jl
@@ -402,7 +402,7 @@ function canny{T<:NumberLike}(img_gray::AbstractMatrix{T}, sigma::Number = 1.4, 
         lowerThreshold = StatsBase.percentile(img_nonMaxSup[:], lowerThreshold * 100)
     end
     img_thresholded = hysteresis_thresholding(img_nonMaxSup, upperThreshold, lowerThreshold)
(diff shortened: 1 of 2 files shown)


src/edge.jl
```diff
@@ -402,8 +402,7 @@ function canny{T<:NumberLike}(img_gray::AbstractMatrix{T}, sigma::Number = 1.4,
         lowerThreshold = StatsBase.percentile(img_nonMaxSup[:], lowerThreshold * 100)
     end
     img_thresholded = hysteresis_thresholding(img_nonMaxSup, upperThreshold, lowerThreshold)
-    S = eltype(img_thresholded)
-    edges = map(i -> i < 0.9 ? zero(S) : one(S), img_thresholded)
+    edges = map(i -> i < 0.9 ? zero(Bool) : one(Bool), img_thresholded)
```
timholy: This could just be
```julia
edges = map(i -> i >= 0.9, img_thresholded)
```

annimesh2809: :sweat_smile: Yes, that will be better

Evizero:
If this change doesn't break any tests, it may be a good idea to add a test verifying the new behaviour

timholy:
Thanks, @annimesh2809!

PR #495, opened 2016-06-11
Adds Canny Edge Detection

mronian:
Based on @siddhantjain's version in #441. Adds thresholding of weak edges using BFS.

PR #607, opened 2017-03-18
changed default percentile option for canny

tejus-gupta:
I have changed the default value of percentile parameter to false. The algorithm is usually used with absolute thresholds and other popular libraries also don't use percentile as default option.

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -60,8 +60,8 @@ using Base.Cartesian  # TODO: delete this
 """
 Percentile - Used to pass arguments to functions as a percentile rather than an absolute value
 
-e.g.,   canny(img, 1.4, (0.8,0.2)) uses absolute threshold
-        canny(img, 1.4, (Percentile(0.8), Percentile(0.2))) uses quantiles of the edge magnitude image as threshold
+e.g.,   canny(img, 1.4, (80, 20)) uses absolute threshold
+        canny(img, 1.4, (Percentile(80), Percentile(20))) uses percentiles of the edge magnitude image as threshold
 
 """
 immutable Percentile{T} <: Real p::T end
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -75,6 +75,7 @@ include("convexhull.jl")
 export # types
     BlobLoG,
     ColorizedArray,
+    Percentile,
 
     # macros
     @test_approx_eq_sigma_eps,
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -57,6 +57,15 @@ using ImageMetadata: ImageMetaAxis
 
 using Base.Cartesian  # TODO: delete this
 
+"""
+Percentile - Used to pass arguments to functions as a percentile rather than an absolute value
+
+e.g.,   canny(img, 1.4, (0.8,0.2)) uses absolute threshold
+        canny(img, 1.4, (Percentile(0.8), Percentile(0.2))) uses quantiles of the edge magnitude image as threshold
+
+"""
+immutable Percentile{T} <: Real p::T end
+
 include("map-deprecated.jl")
 include("overlays-deprecated.jl")
 include("labeledarrays.jl")
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -76,3 +76,12 @@ Base.@deprecate_binding LabeledArray ColorizedArray
 @deprecate ColorizedArray{T,N}(intensity::AbstractArray{T,N}, label::AbstractArray, colors::Vector{RGB}) ColorizedArray(intensity, IndirectArray(label, colors))
 
 @deprecate imcomplement(img::AbstractArray) complement.(img)
+
+function canny{T<:NumberLike}(img_gray::AbstractMatrix{T}, sigma::Number = 1.4, upperThreshold::Number = 0.90, lowerThreshold::Number = 0.10; percentile::Bool = true)
+    depwarn("canny(img, sigma, upperThreshold, lowerThreshold; percentile=true) is deprecated.\n Please use canny(img, (upperThreshold, lowerThreshold), sigma) or canny(img, (Percentile(upperThreshold), Percentile(lowerThreshold)), sigma)",:canny)
+    if percentile==true
+        canny(img_gray, (Percentile(upperThreshold), Percentile(lowerThreshold)), sigma)
+    else
+        canny(img_gray, (upperThreshold, lowerThreshold), sigma)
+    end
+end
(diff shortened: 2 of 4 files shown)


Evizero:
Just a quick opinion on the general nature of this PR.

A colleague of mine got recently burned by a default-value change in a completely unrelated python library. That change took him 3 days to figure out, since suddenly his code was "bugged" (and his git-commit discipline is lacking).

I think it is worth considering that when such a change is merged, then it should be prominently advertised as borderline breaking. I'd say an entry in `NEWS.md` at the very least

timholy:
Sorry this took me a while to get to.

I don't use `canny`, so we should CC @mronian, @annimesh2809, and @kmsquire to see what they think. If we do change this, @Evizero makes a good point. I think the only truly safe option is to change the API in a way that we can issue a deprecation warning. To that effect, perhaps we should copy/paste the current signature into `deprecated.jl` and have it emit a deprecation warning and then call the new API (defined below). We then change the main function so that:

- in the new version, we instead pass a `(lower, upper)` tuple for the thresholds
- we define `immutable Percentile{T} <: Real p::T end`, and if folks want to pass arguments to functions as a percentile rather than an absolute value, require that they pass in `(Percentile(lower), Percentile(upper))`. That will give those numbers unambiguous meaning.

mronian:
If majority of the libraries use absolute thresholds, I am fine with modifying this to what @timholy suggested :smile:

tejus-gupta:
Where should I declare Percentile? The same parameter style can be used in other functions where user may want absolute or percentile parameters. Currently, I have declared it in edge.jl.

tejus-gupta:
If I copy/paste the earlier signature to depreciated.jl, canny(img_gray) will have two valid functions. Should depreciation warning be emitted or the newer version be called straightaway.

tejus-gupta:
If I copy/paste the earlier signature to depreciated.jl, canny(img_gray) will have two valid functions. Should depreciation warning be emitted or the newer version be called straightaway?

tejus-gupta:
If I copy/paste the earlier signature to depreciated.jl, canny(img_gray) will have two valid functions. Should depreciation warning be emitted or the newer version be called straightaway?

timholy:
> Where should I declare Percentile?

Maybe in Images.jl itself? I agree that it's more broadly useful.

> Should depreciation warning be emitted or the newer version be called straightaway?

The goal is to have the deprecation called if the user uses any syntax that would have been supported by the "old" function. Any syntax incompatible with that definition can call the new function straightaway, without issuing the deprecation warning. (This is the main purpose of me changing the types of the inputs, so that we can use the type system to make this happen.)

src/edge.jl
```diff
@@ -385,21 +385,23 @@ Performs Canny Edge Detection on the input image.
 Parameters :
 
   sigma :           Specifies the standard deviation of the gaussian filter
-  upperThreshold :  Upper bound for hysteresis thresholding
-  lowerThreshold :  Lower bound for hysteresis thresholding
+  threshold:        Tuple of upper bound and lower bound for hysteresis thresholding
```
timholy: Might be clearer as "`(upper, lower)` bounds for hysteresis thresholding"

src/Images.jl
```diff
@@ -57,6 +57,15 @@ using ImageMetadata: ImageMetaAxis
 
 using Base.Cartesian  # TODO: delete this
 
+"""
+Percentile - Used to pass arguments to functions as a percentile rather than an absolute value
+
+e.g.,   canny(img, 1.4, (0.8,0.2)) uses absolute threshold
+        canny(img, 1.4, (Percentile(0.8), Percentile(0.2))) uses quantiles of the edge magnitude image as threshold
```
timholy: We should make it clear whether `Percentile(0.8)` corresponds to something that happens 80% of the time or 0.8% of the time. If we use 0.8 to mean 4/5, then perhaps we should use a different name than Percentile?

src/edge.jl
```diff
@@ -385,21 +385,23 @@ Performs Canny Edge Detection on the input image.
...
+function canny{T<:NumberLike, N<:Union{NumberLike,Percentile{NumberLike}}}(img_gray::AbstractMatrix{T}, threshold::Tuple{N,N}, sigma::Number = 1.4)
     img_grayf = imfilter(img_gray, KernelFactors.IIRGaussian((sigma,sigma)), NA())
     img_grad_y, img_grad_x = imgradients(img_grayf, KernelFactors.sobel)
     img_mag, img_phase = magnitude_phase(img_grad_x, img_grad_y)
     img_nonMaxSup = thin_edges_nonmaxsup(img_mag, img_phase)
-    if percentile
-        upperThreshold = StatsBase.percentile(img_nonMaxSup[:], upperThreshold * 100)
-        lowerThreshold = StatsBase.percentile(img_nonMaxSup[:], lowerThreshold * 100)
+    if N<:Percentile{}
+        upperThreshold = StatsBase.percentile(img_nonMaxSup[:], threshold[1].p)
```
timholy: I think you can compute multiple values at once? So you could pass both components of `threshold`?

tejus-gupta:
`astype :          Specifies return type of result`

I think this needed to be deleted after #606.

tejus-gupta:
`astype : Specifies return type of result`

I think this needed to be deleted after #495 .

src/edge.jl
tejus-gupta:
```
if N<:Percentile{}
     threshold=map(x->StatsBase.percentile(img_nonMaxSup[:], x.p), threshold)
end
img_thresholded = hysteresis_thresholding(img_nonMaxSup, threshold...)
```

Like this?

timholy: I think `percentile` can take a vector of thresholds. The slow part of `percentile` is sorting the data array; there is no reason to do that twice.

src/deprecated.jl
```diff
@@ -76,3 +76,12 @@ Base.@deprecate_binding LabeledArray ColorizedArray
 @deprecate ColorizedArray{T,N}(intensity::AbstractArray{T,N}, label::AbstractArray, colors::Vector{RGB}) ColorizedArray(intensity, IndirectArray(label, colors))
 
 @deprecate imcomplement(img::AbstractArray) complement.(img)
+
+function canny{T<:NumberLike}(img_gray::AbstractMatrix{T}, sigma::Number = 1.4, upperThreshold::Number = 0.90, lowerThreshold::Number = 0.10; percentile::Bool = true)
+    depwarn("canny(img, sigma, upperThreshold, lowerThreshold; percentile=true) is deprecated.\n Please use canny(img, (upperThreshold, lowerThreshold), sigma) or canny(img, (Percentile(upperThreshold), Percentile(lowerThreshold)), sigma)",:canny)
+    if percentile==true
+        canny(img_gray, (Percentile(upperThreshold), Percentile(lowerThreshold)), sigma)
```
timholy: upper threshold of 0.9% corresponds to 0.009. Don't you have to multiply by 100 here for backwards compatibility?

timholy:
Looks good. I noticed a couple of "formatting" issues that would have been hard to describe, so see #616.

Note that the Markdown text is not formatted with your linebreaks. This might need fixing elsewhere too:

image

The formatting would be improved by using bullets; it would also be nice to add backticks to make sure the parameter names are highlighted. See an example in the second commit in #616.

timholy:
These commits got squashed and incorporated into #616. Thanks, @tejus-gupta!

PR #608, opened 2017-03-19
Fix `n_scales` condition in `gaussian_pyramid`

Gnimuc:
According to wikipedia:
* `n_scales = 0`: level-0(original image)
* `n_scales = 1` : level-1(1/2 resolution)
* …


timholy:
Thanks, @Gnimuc!

PR #614, opened 2017-03-30
Get Images working on 0.6

timholy:


timholy:
Whew. This has been a long time coming.

QuartzImageIO needs some kind of fix on 0.6, but otherwise this is passing all tests. I went back to requiring that Images pass its tests on nightly, so we'll get overall failures. But given that most stuff should be working now, I'd rather know the worst.

rsrock:
I'll take a look at QuartzImageIO soon.  The last time I checked, something in Images was gumming up the works on the QuartzImageIO tests, so I decided to wait for the dust to settle a bit.  Looks like that time is now.

timholy:
Yeah, there definitely wasn't any point back when Images wouldn't even build on 0.6. Thanks in advance!

PR #616, opened 2017-04-02
Canny revamp (small tweaks to #607)

timholy:
The second commit here makes small tweaks to #607 in the docstrings and depwarns.

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -56,6 +56,15 @@ using ImageMetadata: ImageMetaAxis
 
 using Base.Cartesian  # TODO: delete this
 
+"""
+Percentile - Used to pass arguments to functions as a percentile rather than an absolute value
+
+e.g.,   canny(img, 1.4, (80, 20)) uses absolute threshold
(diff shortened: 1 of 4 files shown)


timholy:
I should explain one change more clearly: notice that I used interpolation `$` in the depwarn for `Percentile`. Since our former thresholds were quantiles rather than percentiles, I was worried that people would forget to multiply their thresholds by 100. The new version should be safer.

tejus-gupta:
@timholy Should canny support one parameter as percentile and other as absolute?

1. If both parameters have to be percentile or absolute, Percentile(80, 20) looks better. In the current version, since the user has to use Percentile() with both the threshold it is repetitive and seems to imply that there is an option to use one threshold as percentile and other as absolute.

2. If we allow one parameter to be percentile and other to be absolute, then the current API looks better.

Even if we don't change anything, we should add a note in the documentation.

timholy:
I suspect it makes most sense to have both be the same. In julia 0.6 we'll be able to use `Percentile.((80, 20))`.

PR #617, opened 2017-04-02
Expunge `percentile` and `method` as (keyword) arguments from Images API

timholy:
After #607 (merged as #616), `canny` disambiguated its threshold using the new `Percentile` type. Before making a new release, for consistency we should get rid of all functions that use an argument corresponding to a `percentile::Bool`. The only other one I'm aware of is `imcorner`.

As a matter of principle, when adding deprecation warnings it's probably a good idea to deprecate anything else related that seems to need changing; this is a kindness to users, to prevent too much API churn. Therefore, I decided to make the `method` keyword argument a positional one. Julia can't dispatch on keyword arguments, so the result wouldn't be type-stable. That's not always important, but it also no longer seems very Julian to me to do this by a keyword argument. The new API supports the following syntaxes:
```julia
imcorner(img)
imcorner(harris, img)
imcorner(img) do A
    harris(A, k=0.1)
end
```
and quite a few others. In particular the `do`-block syntax is a nice way to modify the keywords of the detector-method without having to pass them as arguments to `imcorner` itself (which makes it clearer "who" those keywords are intended for).

My main concern is that other functions in Images (e.g. `imfilter`) take a function as a later argument. I'd be interested in seeing whether people think this is a problem.

@tejus-gupta and @annimesh2809, as our GSoC candidates you might want to look over the first commit as a model of how to handle "difficult" deprecations. I admit this is quite a pain in the neck, but given that we don't really know how many people are using this package and these features (it could be no one, or it could be thousands of people), we have to be pretty careful. As you look that commit over, features worth taking note of:
- I added descriptions of the changes to `NEWS.md`
- The functions added to `src/deprecations.jl` are quite long, but the purpose was to suggest the syntax closest to the one that user is actually calling the function with. So `imcorner(img; method=shi_tomasi)` will suggest `imcorner(shi_tomasi, img)` and `imcorner(img; method=shi_tomasi, border="reflect")` will suggest the `do`-block syntax. I also took care to make sure that the keyword arguments would print as a user might type them (they don't by default).
- I added tests of all the new syntaxes. Because we weren't testing all the old syntaxes, I also added tests for those to our "deprecated" tests.

Naturally, while you (or anyone else) look these commits over you notice problems, by all means suggest improvements! It can go both ways :smile:.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,21 @@
+# v0.10
+
+New features:
+
+- added `Percentile` to disambiguate the interpretation of
+  thresholds. A raw number `x` will now be interpreted as an absolute
+  threshold, whereas `Percentile(x)` (with `0 <= x <= 100`) will
+  choose an absolute threshold based on the distribution of values in
+  the input array.
+
+API changes:
+
+- `canny` passes the threshold as a 2-tuple and uses `Percentile`
+  rather than a `percentile` keyword. It now issues a deprecation
+  warning when used with default arguments.
+
+- `imcorner` now uses `Percentile`. The old syntax issues a deprecation warning.
+
 # v0.9
 
 Breaking changes:
--- a/src/corner.jl
+++ b/src/corner.jl
@@ -17,21 +17,22 @@ The threshold is assumed to be a percentile value unless `percentile` is set to 
 """
 function imcorner(img::AbstractArray; method::Function = harris, args...)
     responses = method(img; args...)
-    corners = falses(size(img))
-    maxima = map(CartesianIndex{2}, findlocalmaxima(responses))
+    corners = similar(img, Bool)
+    fill!(corners, false)
+    maxima = map(CartesianIndex, findlocalmaxima(responses))
     for m in maxima corners[m] = true end
     corners
 end
 
-function imcorner(img::AbstractArray, threshold, percentile; method::Function = harris, args...)
+function imcorner(img::AbstractArray, threshold; method::Function = harris, args...)
     responses = method(img; args...)
+    map(i -> i > threshold, responses)
+end
 
-    if percentile == true
-        threshold = StatsBase.percentile(vec(responses), threshold * 100)
-    end
-
-    corners = map(i -> i > threshold, responses)
-    corners
+function imcorner(img::AbstractArray, thresholdp::Percentile; method::Function = harris, args...)
+    responses = method(img; args...)
+    threshold = StatsBase.percentile(vec(responses), thresholdp.p)
+    map(i -> i > threshold, responses)
 end
 
 """
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -96,3 +96,9 @@ function imcorner(img::AbstractArray, threshold, percentile;
         imcorner(img, threshold; method=method, args...)
     end
 end
+
+function imedge(img::AbstractArray, method::AbstractString, border::AbstractString="replicate")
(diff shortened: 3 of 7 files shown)


annimesh2809:
Quite a nice way to handle "deprecations". :+1:
As you mentioned that some functions like `imfilter` take a function as a later argument, so users might face problems in using this kind of mixed API. I think it would be better to be consistent with the function declarations so that cognizant users do not have to refer to the documentation if they want to use different methods.
Also I could not understand why we have the `/test/old/` directory containing the old tests?

timholy:
> As you mentioned that some functions like imfilter take a function as a later argument, so users might face problems in using this kind of mixed API. I think it would be better to be consistent with the function declarations so that cognizant users do not have to refer to the documentation if they want to use different methods.

Yeah, I'm a little worried about that too. The reason I thought we might be able to get away with it is that `Kernel.gaussian` isn't a function that operates on the image; it's a function that creates the filter data. (It can have different representations, e.g., `Kernel.gaussian` creates a full-dimensional kernel, whereas `KernelFactors.gaussian` uses a factored (separable) representation.) That places it in a different category than `harris`, which operates directly on the image. Nevertheless, it could still be confusing; we could keep the `method` argument as a keyword argument (like it is now) or as a later positional argument, though we'd have to give up the `do`-block syntax. That may not be a huge loss because `imcorner(img, A->harris(A, k=0.1))` is pretty readable. The main advantage of the `do`-block syntax comes when you want to define a longer method, and it's therefore awkward to specify as an anonymous function. (But one could create it as a "standalone" function and pass the parameters using an anonymous function.) The thing I most dislike about the current API is that we pass keyword arguments to `imcorner` that are really intended for the `method`; I think it's clearer to attach the arguments to the function that will use them.

I think if we do change it to be the first argument, we should first try to come up with some coherent policy on argument order. That's not a small job, it basically requires a full API review. I essentially did that as part of creating Images 0.6, but it might need someone else to spend quite a lot of time with http://juliaimages.github.io/latest/function_reference.html so that it's more a reflection of a design team than one single person making stuff up :smile:.

> Also I could not understand why we have the /test/old/ directory containing the old tests?

Great question. The `old` directory contains tests that, by and large, are what they were prior to Images 0.6 (the "new Images"). The goal is to make sure that the old syntax works, even if it emits deprecation warnings. Note that this is not how julia itself is developed (we don't tend to *formally* test the deprecated syntax, though people surely do so locally when developing a deprecation), but I found it useful to make sure that my deprecations were working as intended and that I wasn't breaking stuff. There were so many things deprecated in 0.6 that I was worried that informal tests while developing a specific feature just wouldn't cut it.

Eventually (maybe by mid-summer?) we'll delete `test/old`, or at least the files that are there right now, and perhaps the deprecation warnings themselves. Any new features should have a test in `test/`; I only add to `test/old` when creating new deprecations.

bjarthur:
re. an API review and policy on argument order, we should also consider a policy on function naming.  specifically, there is inconsistent use of "im" as a prefix.  `imadustintensity` for example, but then `adjust_gamma`.  not sure whether it is best to always use it, or always not, but it should always be one way or the other i think.

timholy:
Because I'm interested in fixing #633 (which just needs a tag to close), I've updated this so that it just gets rid of `percentile` as an argument to `imcorner`. We can do a full API review later.

timholy:
OSX failure is just https://github.com/JuliaIO/QuartzImageIO.jl/issues/37. Since this is just a stripped-down version of what I posted a while ago, I'm merging.

PR #627, opened 2017-04-13
iterate over last dimension when grouping N-D tables of images

### LINKED ISSUE #625: Incorrect grouping in multidimensional table of images
I love the table display introduced in https://github.com/JuliaImages/Images.jl/pull/574 :smile: 

But I think it might be grouping arrays of images with more than 2 dimensions in an odd way. Specifically, if I do the following:

```julia
using Colors
[fill(Gray{Float32}(z), 2, 2) for x in 1:2, y in 1:2, z in 0:1]
```

(the `fill(..., 2, 2)` is necessary to work around #623 )

I see:

```
2×2×2 Array{Array{ColorTypes.Gray{Float32},2},3}:
[:, :, 1] =
 ColorTypes.Gray{Float32}[Gray{Float32}(0.0) Gray{Float32}(0.0); Gray{Float32}(0.0) Gray{Float32}(0.0)]  …  ColorTypes.Gray{Float32}[Gray{Float32}(0.0) Gray{Float32}(0.0); Gray{Float32}(0.0) Gray{Float32}(0.0)]
 ColorTypes.Gray{Float32}[Gray{Float32}(0.0) Gray{Float32}(0.0); Gray{Float32}(0.0) Gray{Float32}(0.0)]     ColorTypes.Gray{Float32}[Gray{Float32}(0.0) Gray{Float32}(0.0); Gray{Float32}(0.0) Gray{Float32}(0.0)]

[:, :, 2] =
 ColorTypes.Gray{Float32}[Gray{Float32}(1.0) Gray{Float32}(1.0); Gray{Float32}(1.0) Gray{Float32}(1.0)]  …  ColorTypes.Gray{Float32}[Gray{Float32}(1.0) Gray{Float32}(1.0); Gray{Float32}(1.0) Gray{Float32}(1.0)]
 ColorTypes.Gray{Float32}[Gray{Float32}(1.0) Gray{Float32}(1.0); Gray{Float32}(1.0) Gray{Float32}(1.0)]     ColorTypes.Gray{Float32}[Gray{Float32}(1.0) Gray{Float32}(1.0); Gray{Float32}(1.0) Gray{Float32}(1.0)]
```

The above is hard to read, but the important point is that Julia arrays display in chunks by `[:, :, 1]`, `[:, :, 2]`, etc... which you can confirm by observing that the Gray values in the second chunk are all 1. 

When I load the Images.jl display code, however, I see:

screenshot from 2017-04-13 16 16 08

which groups the elements differently. 

I think the fix for this is just to switch `i` and `colons...` in https://github.com/JuliaImages/Images.jl/blob/master/src/showmime.jl#L65 and https://github.com/JuliaImages/Images.jl/blob/master/src/showmime.jl#L53 but I'd like to make sure this isn't the intended behavior before I try to change that.

rdeits:
Fixes #625 (assuming that that is, in fact, an error). 

Before:

screenshot from 2017-04-13 16 58 35

After:

screenshot from 2017-04-13 16 57 38

My reasoning is that this is more consistent with the way arrays are printed as text in Julia:

```julia
[fill(z * z2) for x in 1:2, y in 1:2, z in [0.5, 1], z2 in [0.5, 1]]
```

gives:

```
2×2×2×2 Array{Array{Float64,0},4}:
[:, :, 1, 1] =
 0.25  0.25
 0.25  0.25

[:, :, 2, 1] =
 0.5  0.5
 0.5  0.5

[:, :, 1, 2] =
 0.5  0.5
 0.5  0.5

[:, :, 2, 2] =
 1.0  1.0
 1.0  1.0
```

### DIFF
--- a/src/showmime.jl
+++ b/src/showmime.jl
@@ -39,7 +39,7 @@ map8(x::Number) = N0f8(x)
 
 function _show_odd{T<:ColorantMatrix}(io::IO, m::MIME"text/html", imgs::AbstractArray{T, 1})
     # display a vector of images in a row
-    for j = 1:length(imgs)
+    for j in eachindex(imgs)
         write(io, "<td style='text-align:center;vertical-align:middle; margin: 0.5em;border:1px #90999f solid;border-collapse:collapse'>")
         show_element(IOContext(io, thumbnail=true), imgs[j])
         write(io, "</td>")
@@ -48,7 +48,7 @@ end
 
 function _show_odd{T<:ColorantMatrix, N}(io::IO, m::MIME"text/html", imgs::AbstractArray{T, N})
     colons = ([Colon() for i=1:(N-1)]...)
-    for i = 1:size(imgs, N)
+    for i in indices(imgs, N)
         write(io, "<td style='text-align:center;vertical-align:middle; margin: 0.5em;border:1px #90999f solid;border-collapse:collapse'>")
         _show_even(io, m, view(imgs, colons..., i)) # show even
         write(io, "</td>")
@@ -60,7 +60,7 @@ function _show_even{T<:ColorantMatrix, N}(io::IO, m::MIME"text/html", imgs::Abst
     centering = center ? " style='margin: auto'" : ""
     write(io, "<table$centering>")
     write(io, "<tbody>")
-    for i = 1:size(imgs, N)
+    for i in indices(imgs, N)
         write(io, "<tr>")
         _show_odd(io, m, view(imgs, colons..., i)) # show odd
         write(io, "</tr>")
--- a/src/showmime.jl
+++ b/src/showmime.jl
@@ -50,7 +50,7 @@ function _show_odd{T<:ColorantMatrix, N}(io::IO, m::MIME"text/html", imgs::Abstr
     colons = ([Colon() for i=1:(N-1)]...)
     for i = 1:size(imgs, 1)
         write(io, "<td style='text-align:center;vertical-align:middle; margin: 0.5em;border:1px #90999f solid;border-collapse:collapse'>")
-        _show_even(io, m, view(imgs, i, colons...)) # show even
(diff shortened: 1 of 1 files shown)


rdeits:
Derp, this doesn't handle dimensions with length > 2 correctly. Please don't merge yet.

rdeits:
Ok, fixed.

src/showmime.jl
```diff
@@ -48,9 +48,9 @@ end
 
 function _show_odd{T<:ColorantArray, N}(io::IO, m::MIME"text/html", imgs::AbstractArray{T, N})
```
timholy: Presumably this needs to be updated to `ColorantMatrix` now? (Sorry if I merged the PRs in the wrong order!)

rdeits: No problem. I've rebased this branch, so it should be fixed now.

rdeits:
Actually, there's one more issue to resolve: this change results in the table of images being displayed in a row-major ordering, which is inconsistent with the rest of Julia. Let me see if I can fix that...

src/showmime.jl
```diff
@@ -48,9 +48,9 @@ end
 
 function _show_odd{T<:ColorantMatrix, N}(io::IO, m::MIME"text/html", imgs::AbstractArray{T, N})
     colons = ([Colon() for i=1:(N-1)]...)
-    for i = 1:size(imgs, 1)
+    for i = 1:size(imgs, N)
```
Evizero: is it worth changing this to `for i in indices(imgs, N)` ?

rdeits:
Oh, never mind; that's clearly the intended behavior, since a vector of images is displayed as a row.

src/showmime.jl
rdeits: Good point. I'll do that.

rdeits: ok, done

timholy:
Thanks!

PR #628, opened 2017-04-13
fix ambiguity with restrict() from ImageTransformations

rdeits:
After this change, the tests are closer to passing, but I'm seeing a strange error: 

```
Restriction: Error During Test
  Got an exception of type ErrorException outside of a @test
  return type AxisArrays.AxisArray{ColorTypes.RGB{Float64},2,Array{ColorTypes.RGB{Float64},2},Tuple{AxisArrays.Axis{:y,FloatRange{Float64}},AxisArrays.Axis{:x,Base.OneTo{Int64}}}} does not match inferred return type Union{AxisArrays.AxisArray{ColorTypes.RGB{Float64},2,Array{ColorTypes.RGB{Float64},2},Tuple{AxisArrays.Axis{:y,FloatRange{Float64}},AxisArrays.Axis{:x,Base.OneTo{Int64}}}},AxisArrays.AxisArray{ColorTypes.RGB{Float64},2,FFTViews.FFTView{ColorTypes.RGB{Float64},2,Array{ColorTypes.RGB{Float64},2}},Tuple{AxisArrays.Axis{:y,FloatRange{Float64}},AxisArrays.Axis{:x,Base.OneTo{Int64}}}},AxisArrays.AxisArray{ColorTypes.RGB{Float64},2,OffsetArrays.OffsetArray{ColorTypes.RGB{Float64},2,Array{ColorTypes.RGB{Float64},2}},Tuple{AxisArrays.Axis{:y,FloatRange{Float64}},AxisArrays.Axis{:x,Base.OneTo{Int64}}}}}
   in macro expansion; at /home/rdeits/.julia/v0.5/Images/test/algorithms.jl:263 [inlined]
```

but I can't reproduce that error at the REPL. I'm curious if it will happen on Travis.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -494,9 +494,9 @@ findlocalminima(img::AbstractArray, region=coords_spatial(img), edges=true) =
 restrict(img::AxisArray, ::Tuple{}) = img
 restrict(img::ImageMeta, ::Tuple{}) = img
 
-const RegionType = Union{Dims,Vector{Int}}
+restrict(img::ImageMeta, region::Vector{Int}) = restrict(img, (region...))
 
-function restrict(img::ImageMeta, region::RegionType=coords_spatial(img))
+function restrict(img::ImageMeta, region::Dims=coords_spatial(img))
     shareproperties(img, restrict(data(img), region))
 end
 
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -500,7 +500,9 @@ function restrict(img::ImageMeta, region::RegionType=coords_spatial(img))
     shareproperties(img, restrict(data(img), region))
 end
 
-function restrict{T,N}(img::AxisArray{T,N}, region::RegionType=coords_spatial(img))
+restrict(img::AxisArray, region::Vector{Int}) = restrict(img, (region...))
+
+function restrict{T,N}(img::AxisArray{T,N}, region::Dims=coords_spatial(img))
     inregion = falses(ndims(img))
     inregion[[region...]] = true
     inregiont = (inregion...,)::NTuple{N,Bool}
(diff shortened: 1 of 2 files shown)


rdeits:
The error goes away if I change:

```julia
function restrict{Ax}(img::Union{AxisArray,ImageMetaAxis}, ::Type{Ax})
    A = restrict(img.data, axisdim(img, Ax))
    AxisArray(A, replace_axis(modax(img[Ax]), axes(img)))
end
```

to

```julia
function restrict{Ax}(img::AxisArray, ::Type{Ax})
    A = restrict(img.data, axisdim(img, Ax))
    AxisArray(A, replace_axis(modax(img[Ax]), axes(img)))
end

function restrict{Ax}(img::ImageMetaAxis, ::Type{Ax})
    A = restrict(img.data, axisdim(img, Ax))
    AxisArray(A, replace_axis(modax(img[Ax]), axes(img)))
end
```

i.e. if I split up the `Union{}` manually. This seems like a Julia bug.

rdeits:
The following seems to be a roughly minimal example that causes the failure: 

```julia
using Images
using Base.Test

@testset "bad" begin
        imgcol = colorview(RGB, rand(3,5,6))
        imgcolax = AxisArray(imgcol, :y, :x)

        # Removing these lines fixes the issue
        imgr = restrict(imgcolax, (1,2))
        @test pixelspacing(imgr) == (2,2)
        @test pixelspacing(imgcolax) == (1,1)  # issue #347

        # This is where the error occurs
        @inferred(restrict(imgcolax, Axis{:y}))

        # Removing these lines also fixes the issue
        img1 = colorview(RGB, fill(0.9, 3, 5, 5))
        img2 = colorview(RGB, fill(N0f8(0.9), 3, 5, 5))
        @test isapprox(channelview(restrict(img1)), channelview(restrict(img2)), rtol=0.01)
end
```

timholy:
Yes, shenanigans are in order. I'd add "change the order in which methods compile" to the list.

To the extent that we can fix it, I think this may be a 2-part fix. Part 1 is in https://github.com/JuliaImages/ImageTransformations.jl/pull/21.

timholy:
Just adding a note that the thing holding up a fix for this is inability to merge https://github.com/JuliaLang/METADATA.jl/pull/8838.

timholy:
Or more specifically, https://github.com/JuliaLang/METADATA.jl/pull/8864, which is queued up behind that one and which suffers from the same problem.

timholy:
Fixed in #631 (relied heavily on changes to ImageCore).

Issue #623, opened 2017-04-13
BoundsError when displaying a matrix of 1-dimensional images


rdeits:
To reproduce (in IJulia):

```julia
using Images
[[RGB{Float64}(r, g, b)] for r in (0.5, 1), g in (0.5, 1), b in (0.5, 1)]
```

gives:

```
BoundsError: attempt to access (1,)
  at index [2]

 in downsize_for_thumbnail(::Array{ColorTypes.RGB{Float64},1}, ::Int64, ::Int64) at /home/rdeits/.julia/v0.5/Images/src/showmime.jl:87
 in show_element(::IOContext{Base.AbstractIOBuffer{Array{UInt8,1}}}, ::Array{ColorTypes.RGB{Float64},1}) at /home/rdeits/.julia/v0.5/Images/src/showmime.jl:95
 in _show_odd(::IOContext{Base.AbstractIOBuffer{Array{UInt8,1}}}, ::MIME{Symbol("text/html")}, ::SubArray{Array{ColorTypes.RGB{Float64},1},1,Array{Array{ColorTypes.RGB{Float64},1},3},Tuple{Int64,Int64,Colon},true}) at /home/rdeits/.julia/v0.5/Images/src/showmime.jl:44
 in _show_even(::IOContext{Base.AbstractIOBuffer{Array{UInt8,1}}}, ::MIME{Symbol("text/html")}, ::SubArray{Array{ColorTypes.RGB{Float64},1},2,Array{Array{ColorTypes.RGB{Float64},1},3},Tuple{Int64,Colon,Colon},true}, ::Bool) at /home/rdeits/.julia/v0.5/Images/src/showmime.jl:65
 in _show_even(::IOContext{Base.AbstractIOBuffer{Array{UInt8,1}}}, ::MIME{Symbol("text/html")}, ::SubArray{Array{ColorTypes.RGB{Float64},1},2,Array{Array{ColorTypes.RGB{Float64},1},3},Tuple{Int64,Colon,Colon},true}) at /home/rdeits/.julia/v0.5/Images/src/showmime.jl:59
 in _show_odd(::IOContext{Base.AbstractIOBuffer{Array{UInt8,1}}}, ::MIME{Symbol("text/html")}, ::Array{Array{ColorTypes.RGB{Float64},1},3}) at /home/rdeits/.julia/v0.5/Images/src/showmime.jl:53
[...]

PR #632, opened 2017-04-19
more specific test for #623

rdeits:
Testing `show()` is somewhat useless, since the whole point of #623 was to prevent any Images.jl code from being called in the first place. This just provides a direct test against the old behavior sneaking back in.


timholy:
Thanks!

PR #639, opened 2017-06-04
Thresholding

tejus-gupta:
Functions for computing threshold using Otsu's method and adaptive thresholding.


src/algorithms.jl
```diff
@@ -908,3 +908,83 @@ function pyramid_scale(img, downsample)
...
+
+Computes treshold for grayscale image using Otsu's method.  
+  
+Parameters:  
+-    img         = Grayscale input image    
+-    bins        = Number of bins used to compute the histogram. Needed for floating-point images.    
+    
+"""
+function otsu_threshold{T<:FixedPointNumbers.Normed}(img::AbstractArray{Gray{T}, 2})
+    bins::Int = 2^T.parameters[2]
```
timholy: Do you really want fewer bins for `N2f6` than `N0f8`? Wouldn't `2^(8*sizeof(T))` be safer? You may also want to put an upper limit on this; a user who supplies a `N16f16` image might be surprised by the memory requirement.

src/algorithms.jl
```diff
@@ -908,3 +908,83 @@ function pyramid_scale(img, downsample)
...
+```
+
+Computes treshold for grayscale image using Otsu's method.  
+  
+Parameters:  
+-    img         = Grayscale input image    
+-    bins        = Number of bins used to compute the histogram. Needed for floating-point images.    
+    
+"""
+function otsu_threshold{T<:FixedPointNumbers.Normed}(img::AbstractArray{Gray{T}, 2})
```
timholy: We also often accept pure numbers as if they are grayscale.

src/algorithms.jl
```diff
@@ -908,3 +908,83 @@ function pyramid_scale(img, downsample)
...
+    bins::Int = 2^T.parameters[2]
+    histogram = zeros(Float64, bins)
+    for v in img
+        histogram[gray(v).i + 1] += 1
```
timholy: More robust (and perhaps clearer) to use `reinterpret(gray(v))`.

src/algorithms.jl
timholy: Should we allow the user to override the default number of bins, and just make `bin` an argument with a default value?

src/algorithms.jl
```diff
@@ -908,3 +908,83 @@ function pyramid_scale(img, downsample)
...
+
+        σb = (μT*ω0-μt)^2/(ω0*ω1)
+
+        if(σb > max_σb)
+            max_σb = σb
+            thres = t
+        end
+    end
+
+    return T((thres-1)/bins)
```
timholy: I think this won't give the right answer for `N2f6`. Maybe `reinterpret(T, thres-1)`?

src/algorithms.jl
```diff
@@ -908,3 +908,83 @@ function pyramid_scale(img, downsample)
...
+end
+
+function otsu_threshold{T<:AbstractFloat}(img::AbstractArray{Gray{T}, 2}, bins::Integer)
+    
+    histogram = zeros(Float64, bins)
+    for v in img
+        index::Int = round(gray(v)*bins + 1)
+        if index == bins + 1
+            index = bins
+        end
```
timholy: What if there are values outside the range of 0-1? This currently should throw an error, which might be best. Alternatively, would it be worth checking `extrema`? The cost, of course, is two passes through the image which I can understand you might want to avoid.

src/algorithms.jl
```diff
@@ -908,3 +908,83 @@ function pyramid_scale(img, downsample)
...
+        histogram[index] += 1
+    end
+    histogram=histogram./sum(histogram)
+
+    ω0 = 0
```
timholy: Seems like this part is common to both methods; worth splitting out into a separate function and having it called by both?

src/algorithms.jl
tejus-gupta: When I am given grayscale image with N2f6 or pure number image levels, I am not sure which range I should use to compute the bins. 

Do I still assume that 0 maps to black and 1 maps to white(then the bins will still be divided uniformly between [0, 1] and will give an error if values are outside this range) or do I divide the bins uniformly between [typemin, typemax] e.g, [0, 4] for N2f6. I think the best option would be divide the bins between minimum and maximum image levels.

timholy: For *display*, 0 maps to black and 1 maps to white, as always. But it's possible for an image to have values that are "out of gamut," meaning "not displayable". ImageView uses `clamp01nan` so it doesn't error. But you have to decide: is this a display-oriented function or an algorithmic function? I assume the latter, in which case it's better to have it work on arbitrary values.
```julia
julia> typemax(N2f6)
4.05N2f6
```

timholy:
I've noted a couple of changes here, but are you looking for review? The most obvious thing is that this needs tests.

tejus-gupta:
I have added tests and this is ready for review. Currently adaptive_threshold uses mapwindow. I am writing an efficient median filter and will submit a separate pull request. [This paper](https://nomis80.org/ctmf.pdf) looks promising.

timholy:
It's not passing tests, though, see the Travis failures. I think you need to add TestImages to `test/REQUIRE`.

src/algorithms.jl
```diff
@@ -908,3 +908,124 @@ function pyramid_scale(img, downsample)
...
+-    sigma       = standard deviation for kernel if "gaussian" method is used
+    
+"""
+function adaptive_threshold{T<:Gray}(img::AbstractArray{T, 2}, block_size::Int, offset::Real, method = "mean"; sigma::Float64 = (block_size - 1) / 6.0)
+
+    block_size>0 || error("block_size must greater than zero and odd\n")
+    isodd(block_size) || error("block_size must be odd\n")
+    
+    if method == "mean"
+        kernel = centered(fill(1/block_size^2, (block_size, block_size)))
```
timholy: For performance it's probably best to use a separable kernel,
```julia
kernel1 = fill(1/block_size, block_size)
kernel = kernelfactors(map(centered, ntuple(d->kernel1, ndims(img))))
```

src/algorithms.jl
```diff
@@ -908,3 +908,124 @@ function pyramid_scale(img, downsample)
...
+  
+Parameters:  
+-    img         = Grayscale input image    
+-    block_size  = Size(must of odd) of pixel neighbourhood used for calculating threshold.  
+-    offset      = Constant subtracted from mean/median/gaussian weighted mean of local neighbourhood to compute threshold
+-    method      = Method used to compute threshold (Mean / Median / Gaussian)
+-    sigma       = standard deviation for kernel if "gaussian" method is used
+    
+"""
+function adaptive_threshold{T<:Gray}(img::AbstractArray{T, 2}, block_size::Int, offset::Real, method = "mean"; sigma::Float64 = (block_size - 1) / 6.0)
```
timholy: Since this doesn't affect output type, I think it's fine to use a string for the method here. In places where it does, I'm thinking of switching to passing in the function itself.

src/algorithms.jl
```diff
@@ -908,3 +908,124 @@ function pyramid_scale(img, downsample)
...
+    block_size>0 || error("block_size must greater than zero and odd\n")
+    isodd(block_size) || error("block_size must be odd\n")
+    
+    if method == "mean"
+        kernel = centered(fill(1/block_size^2, (block_size, block_size)))
+        thres = imfilter(img, kernel)
+    elseif method == "median"
+        thres = mapwindow(median!, map(x->gray(x), img), (block_size, block_size))
+    elseif method == "gaussian"
+        kernel = Kernel.gaussian((sigma, sigma), (block_size, block_size))
```
timholy: Here too:
```julia
kernel = KernelFactors.gaussian((sigma, sigma), (block_size, block_size))
```
If you think folks will use a *really* big kernel, consider the `IIRGaussian` methods.

Finally, it seems to me that we might want to consider `NA()` boundary conditions for both calls to `imfilter` to reduce the impact of the edges?

src/algorithms.jl
```diff
@@ -908,3 +908,124 @@ function pyramid_scale(img, downsample)
...
+        thres = imfilter(img, kernel)
+    end
+    
+    return convert(Array{T, 2}, thres - offset)
```
timholy: What if the input image doesn't index starting at 1? Might be best to just skip the `convert`?

src/algorithms.jl
```diff
@@ -908,3 +908,124 @@ function pyramid_scale(img, downsample)
...
+end
+
+"""
+```
+threshold(img, thres, [method])
```
timholy: `imgbool = threshold(img, thresh, [method])` to document the return value?

src/algorithms.jl
tejus-gupta: ```imfilter``` doesn't seem to work with such arrays.

e.g,
```
img = OffsetArrays.OffsetArray(Float64, -10:10, -10:10)
block_size = 21
kernel = centered(fill(1/block_size^2, (block_size, block_size)))
imfilter(y, kernel)    # error
```

timholy: Hmm, didn't realize that. I'll fix it. Is there any other reason you're calling `convert`?

src/algorithms.jl
```diff
@@ -908,3 +908,124 @@ function pyramid_scale(img, downsample)
...
+```
+
+Computes threshold for grayscale image using Otsu's method.  
+  
+Parameters:  
+-    img         = Grayscale input image    
+-    bins        = Number of bins used to compute the histogram. Needed for floating-point images.    
+    
+"""
+function otsu_threshold{T<:Gray}(img::AbstractArray{T, 2}, bins::Int = 256)
```
tejus-gupta: If ```imhist()``` works for higher dimensional arrays, I doesn't need to do anything extra. I will update the functions to work for arbitrary dimension.

timholy: In particular, you can generalize this to multidimensional images just by getting rid of the `2`.

timholy: (In case you're wondering, abstract types are "bad" only for instances, not for function arguments. Dispatch will use the actual type, not the abstract type, when it compiles the method.)

src/algorithms.jl
tejus-gupta: Should I always use ```IIRGaussian``` or only for large window size? Has anyone done benchmarks for  ```KernelFactors.IIRGaussian```  and ```KernelFactors.gaussian``` as kernel length changes?

timholy: No serious benchmarks, so it would be great to test directly.

src/algorithms.jl
tejus-gupta: Any special tests I should add for higher-dimensional arrays (maybe corner cased often missed by such algorithms)? I have added a test to check if it returns otsu_threshold correctly for 3D array.

src/algorithms.jl
tejus-gupta: I thought the user would expect the threshold mask to be in same type as image. I think it's fine without ```convert ``` but we should document that it always returns a Float64 array.

tejus-gupta: I thought the user would expect the threshold mask to be in same type as image. We can still use ```map``` for it to work with array not using one-based indexing. I think it's fine without ```convert ``` but we should document that it always returns a Float64 array.

tejus-gupta: ```imfilter``` always returns ```Gray{Float64}``` array and mapwindow always returns ``` Float64``` array. I used ```convert```  to make the function type stable.I though the user would expect the threshold mask to be in same type as input image. I can still use ```map``` for arrays whose index don't start with one.

src/algorithms.jl
tejus-gupta: For a 1000x1000 image, I compared ```KernelFactors.IIRGaussian``` and ```KernelFactors.gaussian``` for kernel length = 11, 21, 31, ..., 201 and there was almost no difference in running time.

tejus-gupta: I agree that NA() boundary condition should work best. Should this(how to deal with boundary) be a function argument?

src/algorithms.jl
timholy: I don't think so, this case seems pretty straightforward.

src/algorithms.jl
timholy: If you pass either `Float32` or `N0f8` numbers to `mapwindow`, it returns `Float32`. Likewise, if you specify your `σ` as a `Float32` then `imfilter` will return the same eltype.

timholy: It's necessary to return something with floating-point eltype from `N0f8`, because `N0f8` can't encode negative numbers.

src/algorithms.jl
timholy: Right, this isn't returning a `Bool`. So you should name the output something different.

src/algorithms.jl
timholy: > and there was almost no difference in running time.

? I get about a 100x speed difference:
```julia
julia> a = rand(1000,1000);

julia> k = KernelFactors.gaussian((201, 201))
(ImageFiltering.KernelFactors.ReshapedOneD{Float64,2,0,OffsetArrays.OffsetArray{Float64,1,Array{Float64,1}}}([0.000281337, 0.000284147, 0.000286978, 0.00028983, 0.000292703, 0.000295597, 0.000298512, 0.000301449, 0.000304407, 0.000307386  …  0.000307386, 0.000304407, 0.000301449, 0.000298512, 0.000295597, 0.000292703, 0.00028983, 0.000286978, 0.000284147, 0.000281337]), ImageFiltering.KernelFactors.ReshapedOneD{Float64,2,1,OffsetArrays.OffsetArray{Float64,1,Array{Float64,1}}}([0.000281337, 0.000284147, 0.000286978, 0.00028983, 0.000292703, 0.000295597, 0.000298512, 0.000301449, 0.000304407, 0.000307386  …  0.000307386, 0.000304407, 0.000301449, 0.000298512, 0.000295597, 0.000292703, 0.00028983, 0.000286978, 0.000284147, 0.000281337]))

julia> imfilter(a, k);

julia> @time imfilter(a, k);
  3.202527 seconds (135 allocations: 65.347 MiB, 0.17% gc time)

julia> k2 = KernelFactors.IIRGaussian((201, 201));

julia> imfilter(a, k2);

julia> @time imfilter(a, k2);
  0.034880 seconds (39 allocations: 22.905 MiB, 18.09% gc time)
```

timholy: > Should this(how to deal with boundary) be a function argument?

If we think `NA` is basically always the right answer, I think it could be hard coded. But if you see a need to tweak it, that's fine too. Really, perhaps one should just pass the smoothing function in as a function...at which point this becomes pretty trivial.

test/algorithms.jl
```diff
@@ -490,6 +490,73 @@ using Base.Test
         @test sort(B)==sort(C)
     end
 
+    @testset "Thresholding" begin
+
+        #otsu_threshold
+        img = testimage("cameraman")
+        thres = otsu_threshold(img)
+        @test typeof(thres) == eltype(img)
+        @test_approx_eq_eps gray(thres) convert(N0f8, 87/256) eps(N0f8)
```
timholy: Are you using Julia 0.5? These cause deprecation warnings on 0.6. Probably want to switch these to `@test ≈(gray(thresh), convert(N0f8, 87/256), eps(N0f8))`.

src/algorithms.jl
```diff
@@ -908,3 +908,124 @@ function pyramid_scale(img, downsample)
...
+        kernel = kernelfactors(map(centered, ntuple(d->kernel1, ndims(img))))
+        thres = imfilter(img, kernel)
+    elseif method == "median"
+        thres = mapwindow(median!, map(x->gray(x), img), ntuple(x->block_size, ndims(img)))
+    elseif method == "gaussian"
+        kernel = Kernel.gaussian(ntuple(x->sigma, ndims(img)), ntuple(x->block_size, ndims(img)))
+        thres = imfilter(img, kernel)
+    end
+    
+    return convert(Array{T, N}, thres - offset)
```
timholy: Here's what I meant:

```julia
julia> using TestImages

julia> img = testimage("cameraman");

julia> thresh = adaptive_threshold(img, 3, 0.5)
ERROR: ArgumentError: element type FixedPointNumbers.Normed{UInt8,8} is an 8-bit type representing 256 values from 0.0 to 1.0,
  but the values (-0.017211328976034945,) do not lie within this range.
  See the READMEs for FixedPointNumbers and ColorTypes for more information.
Stacktrace:
 [1] throw_colorerror_(::Type{FixedPointNumbers.Normed{UInt8,8}}, ::Tuple{Float64}) at /home/tim/.julia/v0.6/ColorTypes/src/types.jl:632
 [2] throw_colorerror(::Type{FixedPointNumbers.Normed{UInt8,8}}, ::Float64) at /home/tim/.julia/v0.6/ColorTypes/src/types.jl:622
 [3] checkval at /home/tim/.julia/v0.6/ColorTypes/src/types.jl:600 [inlined]
 [4] ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}}(::Float64) at /home/tim/.julia/v0.6/ColorTypes/src/types.jl:347
 [5] copy!(::IndexLinear, ::Array{ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}},2}, ::IndexLinear, ::Array{Float64,2}) at ./abstractarray.jl:655
 [6] convert(::Type{Array{ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}},2}}, ::Array{Float64,2}) at /home/tim/.julia/v0.6/ImageCore/src/convert_reinterpret.jl:107
 [7] #adaptive_threshold#45(::Float64, ::Function, ::Array{ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}},2}, ::Int64, ::Float64, ::String) at /home/tim/.julia/v0.6/Images/src/algorithms.jl:985
 [8] adaptive_threshold(::Array{ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}},2}, ::Int64, ::Float64) at /home/tim/.julia/v0.6/Images/src/algorithms.jl:971
```

test/algorithms.jl
```diff
@@ -490,6 +490,73 @@ using Base.Test
...
+        thres = otsu_threshold(img, 512)
+        @test typeof(thres) == eltype(img)
+        @test_approx_eq_eps gray(thres) convert(N0f8, 87/256) eps(N0f8)
+
+        img = map(x->convert(ColorTypes.Gray{Float64}, x), img)
```
timholy: You can get rid of the `ColorTypes.` here.

test/algorithms.jl
```diff
@@ -490,6 +490,73 @@ using Base.Test
...
+            end
+        end
+
+        thres = colorview(Gray, fill(0.5, (5,5)))
+        img2 = threshold(img, thres)
+        @test eltype(img2) == eltype(img)
+        @test size(img2) == size(img)
+        for i in 1:5
+            for j in 1:5
+                @test_approx_eq gray(img2[i,j]) img[i,j]>thres[i,j]?1.0:0.0
```
timholy: `@test gray(img2[i,j]) == img[i,j]>thresh[i,j]`.

src/algorithms.jl
tejus-gupta: Got it. Without the convert, wouldn't the return type of the function would depend on the value of the argument ```method```?

test/algorithms.jl
tejus-gupta: Yes, I was using Julia 0.5. I have updated now.

src/algorithms.jl
timholy: Hmm, yes, I guess this is true. We either have to have a "safe" type (that can be computed inferrably) to `convert` to, or to pass in the method as a function.

Another thought: do we really need this function? It's a pretty thin wrapper around a filtering method.

src/algorithms.jl
tejus-gupta: I don't know what I did wrong earlier but I can see the difference right now( btw you seem to have set σ=201, not kernel size). It seems that ```IIRGaussian``` is always faster but it doesn't give accurate results for σ<=1.

timholy: Sorry, I don't understand this:

> btw you seem to have set σ=201, not kernel size

can you clarify?

src/algorithms.jl
tejus-gupta: I agree that the function doesn't do much and we can drop ```adaptive_threshold``` and ```threshold```.

timholy: Sometimes that's the easiest way to solve a problem like this.

tejus-gupta: I have removed ```adaptive_threshold``` and ```threshold```.

timholy:
Yay, thanks!

PR #642, opened 2017-06-13
Use IOContext hint in show

davidanthoff:


test/writemime.jl
```diff
@@ -19,6 +19,59 @@ using Base.Test
...
+        # end
+        # b = load(fn)
+        # @test b == img
+
+        # img = Gray.([0.1 0.2; -0.5 0.8])
+        # open(fn, "w") do file
+        #     show(IOContext(file, :full_fidelity=>true), MIME("image/png"), img, minpixels=0, maxpixels=typemax(Int))
+        # end
+        # b = load(fn)
+        # @test b == img
```
davidanthoff: @timholy I think these should in theory also work, but they don't. I think ``b`` and ``img`` end up having different types when the ``@test`` macro is called. I don't understand the package well enough to sort this out. Two options: either you tell me what to do :), or I just remove these three tests. I don't think we really need them, the other tests already test the whole ``:full_fidelity`` IOContext code path.

timholy: ImageMagick can't save floating-point images, and also not images with negative values. You can do something like `Gray.(N0f8[0.1 0.2; 0.5 0.8])` if you want. But if you want to remove them, that's fine; the really crucial test seems to be the "big image" test.

timholy:
This definitely is the right way to fix this, nice work!

test/writemime.jl
davidanthoff: Alright, I removed those.

davidanthoff:
I think if the tests pass this could be merged.

davidanthoff:
@timholy The errors on Mac seem unrelated to my changes here?

timholy:
Thanks!

PR #643, opened 2017-06-13
Wrap tests in one master testset

davidanthoff:
Makes it easier to see all test results in one view.


timholy:
OSX failure should be fixed with a new tag, see https://github.com/JuliaIO/QuartzImageIO.jl/issues/37.

PR #645, opened 2017-06-14
Blob_LoG

AShedko:
This aims to resolve [this issue](https://github.com/JuliaImages/Images.jl/issues/641)

Summary: 
---
blob_LoG function now works on N-dimensional images. 
the function signature changed to:
```blob_LoG(img, σbase, [edges], σfactors) -> Vector{BlobLoG}```
where `σbase` is a vector of magnitudes and `σfactors<:Vector` represents the scales in the image.

### CODE REVIEW

**Comment by AShedko** [comment]
Weirdly, this returns 1 blob
```
        A = zeros(Int, 9, 9, 9); A[5, 5, 5] = 1
        blobs = blob_LoG(A, 2.0.^[0.5, 0, 1])

>>> 1-element Array{Images.BlobLoG{Float64,Float64,3},1}:
 Images.BlobLoG{Float64,Float64,3}(CartesianIndex{3}((5,5,5)),1.0,0.190481)
```
while this returns 6 of them (1 normal and 5 with tiny amplitudes)

```
  A = zeros(Int, 9, 9, 9); A[3, 3, 3] = 1
  blobs = blob_LoG(A, 2.0.^[0.5, 0, 1])

>>> 6-element Array{Images.BlobLoG{Float64,Float64,3},1}:
 Images.BlobLoG{Float64,Float64,3}(CartesianIndex{3}((3,3,3)),1.0,0.190481)   
 Images.BlobLoG{Float64,Float64,3}(CartesianIndex{3}((8,6,3)),1.0,2.87645e-18)
 Images.BlobLoG{Float64,Float64,3}(CartesianIndex{3}((2,8,3)),1.0,5.82619e-18)
 Images.BlobLoG{Float64,Float64,3}(CartesianIndex{3}((8,4,6)),1.0,1.81186e-18)
 Images.BlobLoG{Float64,Float64,3}(CartesianIndex{3}((8,8,6)),1.0,1.60526e-18)
 Images.BlobLoG{Float64,Float64,3}(CartesianIndex{3}((2,3,8)),1.0,6.83169e-18)
```

An issue with kernels, perhaps?

File: src/algorithms.jl [hunk 1]
```diff
@@ -422,17 +422,22 @@ International Journal of Computer Vision, 30(2), 79–116.
 
 See also: [`BlobLoG`](@ref).
 """
-function blob_LoG{T,N}(img::AbstractArray{T,N}, σs, edges::Tuple{Vararg{Bool}}=(true, ntuple(d->false, Val{N})...))
-    sigmas = sort(σs)
+function blob_LoG{T,N}(img::AbstractArray{T,N}, σbase::Union{AbstractVector,Tuple},
```
timholy: This is just variable naming, but I would think of `σbase` as the "shape" of `σ` (defining the ratio of lengths along the different axes) and `σfactors` as holding the scale factors. In other words, I would lean towards swapping the names. Or better yet, coming up with new ones that are less confusing

src/algorithms.jl
```diff
@@ -422,17 +422,22 @@ International Journal of Computer Vision, 30(2), 79–116.
...
     @inbounds for isigma in eachindex(sigmas)
-        img_LoG[isigma,colons...] = (-sigmas[isigma]) * imfilter(img, Kernel.LoG(sigmas[isigma]))
+        img_LoG[isigma,colons...] = (-sigmas[isigma]) * imfilter(img, Kernel.LoG(ntuple(i->sigmas[isigma]*σfactors[i],Val{N})))
     end
     maxima = findlocalmaxima(img_LoG, 1:ndims(img_LoG), edges)
     [BlobLoG(CartesianIndex(tail(x.I)), sigmas[x[1]], img_LoG[x]) for x in maxima]
 end
-blob_LoG{T,N}(img::AbstractArray{T,N}, σs, edges::Bool) = blob_LoG(img, σs, (edges, ntuple(d->edges,Val{N})...))
+blob_LoG{T,N}(img::AbstractArray{T,N}, σbase, edges::Bool, σfactors=ntuple(d->1, Val{N})) =
+ blob_LoG(img, σbase, (edges, ntuple(d->edges,Val{N})...), σfactors)
```
timholy: indentation (4 spaces)

src/algorithms.jl
```diff
@@ -422,17 +422,22 @@ International Journal of Computer Vision, 30(2), 79–116.
 
 See also: [`BlobLoG`](@ref).
 """
-function blob_LoG{T,N}(img::AbstractArray{T,N}, σs, edges::Tuple{Vararg{Bool}}=(true, ntuple(d->false, Val{N})...))
-    sigmas = sort(σs)
+function blob_LoG{T,N}(img::AbstractArray{T,N}, σbase::Union{AbstractVector,Tuple},
+   edges::Tuple{Vararg{Bool}}=(true, ntuple(d->false, Val{N})...), σfactors=ntuple(d->1, Val{N}))
```
timholy: Maybe indent this so it's underneath the `(img::`, one character to the right of the `(`.

src/algorithms.jl
```diff
@@ -422,17 +422,22 @@ International Journal of Computer Vision, 30(2), 79–116.
...
     end
     maxima = findlocalmaxima(img_LoG, 1:ndims(img_LoG), edges)
     [BlobLoG(CartesianIndex(tail(x.I)), sigmas[x[1]], img_LoG[x]) for x in maxima]
 end
-blob_LoG{T,N}(img::AbstractArray{T,N}, σs, edges::Bool) = blob_LoG(img, σs, (edges, ntuple(d->edges,Val{N})...))
+blob_LoG{T,N}(img::AbstractArray{T,N}, σbase, edges::Bool, σfactors=ntuple(d->1, Val{N})) =
+ blob_LoG(img, σbase, (edges, ntuple(d->edges,Val{N})...), σfactors)
+
+blob_LoG{T,N}(img::AbstractArray{T,N}, σbase, σfactors=ntuple(d->1, Val{N})) =
+ blob_LoG(img, σbase, (true, ntuple(d->false,Val{N})...), σfactors)
```
timholy: indentation

timholy:
Oh, interesting that you get this problem. I wonder if we should put in a threshold? Note that this probably won't happen for any real image, where the pixel intensities won't be perfectly flat.

src/algorithms.jl
AShedko: Maybe just call them `σscale` and `σshape` respectively?

timholy: That would be fine. It should also be added to the docstring that the `i`th contender is `σscales[i] * σshape`.

src/algorithms.jl
```diff
@@ -402,12 +402,12 @@ immutable BlobLoG{T,S,N}
...
 
 """
-    blob_LoG(img, σs, [edges]) -> Vector{BlobLoG}
+    blob_LoG(img, σscales, [edges], σshape) -> Vector{BlobLoG}
 
 Find "blobs" in an N-D image using the negative Lapacian of Gaussians
 with the specifed vector or tuple of σ values. The algorithm searches for places
 where the filtered image (for a particular σ) is at a peak compared to all
-spatially- and σ-adjacent voxels.
+spatially- and σ-adjacent voxels, where σ is `σscales[i] * σshape` for some i.
```
timholy: Maybe add "By default, `σshape` is an ntuple of 1s" and place the variable in brackets to indicate that it's optional?

test/algorithms.jl
```diff
@@ -45,6 +45,17 @@ using Base.Test
...
+        A = zeros(Int, 9, 9, 9); A[5, 5, 5] = 1
+        blobs = blob_LoG(A, 2.0.^[0.5, 0, 1])
+        @test length(blobs) == 1
+        @test blobs[1].location == CartesianIndex((5,5,5))
+        # kinda anisotropic image
+        A = zeros(Int,9,9,9); A[5,4:6,5] = 1;
+        blobs = blob_LoG(A,[1.0], [1.,3.,1.])
+        @test length(blobs) == 1
+        @test blobs[1].location == CartesianIndex((5,5,5))
+
```
bjarthur: code coverage would be improved if the `σscales`, `edges`, and `σshape` inputs were used in the tests.

src/algorithms.jl
```diff
@@ -422,17 +422,22 @@ International Journal of Computer Vision, 30(2), 79–116.
 
 See also: [`BlobLoG`](@ref).
 """
-function blob_LoG{T,N}(img::AbstractArray{T,N}, σs, edges::Tuple{Vararg{Bool}}=(true, ntuple(d->false, Val{N})...))
-    sigmas = sort(σs)
+function blob_LoG{T,N}(img::AbstractArray{T,N}, σscales::Union{AbstractVector,Tuple},
+                       edges::Tuple{Vararg{Bool}}=(true, ntuple(d->false, Val{N})...), σshape=ntuple(d->1, Val{N}))
```
bjarthur: without a semi-colon, these are optional args, not keyword args.  that's fine with me, but, do we want to switch the order of `edges` and `σshape` in order to keep `σscales` and `σshape` next to each other?

AShedko: That would break backwards compatibility. And the scaling seem very niche. Is the performance penalty of keyword args justified?

src/algorithms.jl
AShedko: Done!

timholy:
Any further thoughts, @bjarthur? I'm fine with this as it is, but you raise good points and we should bring this to a conclusion.

If there is continued enthusiasm for switching the arg order, perhaps we could do something even more drastic to associate the two, e.g.,
```julia
blob_LoG(img, (σscales, σshape), edges)
```
where the main function is
```julia
function blob_LoG{T,N}(img::AbstractArray{T,N}, σinfo::Tuple{Union{AbstractVector,Tuple}, NTuple{N,Real}},
                       edges::Tuple{Vararg{Bool}}=(true, ntuple(d->false, Val{N})...))
    σscales, σshape = σinfo
    ...
```
and we could have a fallback for 2d of
```julia
blob_LoG(img::AbstractMatrix, σscales::Union{AbstractVector,Tuple},
                  edges::Tuple{Bool,Bool}=(true, false)) = blob_LoG(img, (σscales, (1, 1)), edges)
```

But switching to keyword args might also have merit. I suspect that compared to the cost of the filtering operations, the keyword penalty is trivial. If we do make a breaking change we can always use `@deprecate`.

But I'm also OK with this as is, so @bjarthur just let us know what you think.

bjarthur:
i'm fine with it as is too.  most important thing is that it now works on ND arrays.  we can always change the interface later if we decide it's too klunky.

timholy:
Thanks! OK, @AShedko I'll merge this shortly unless you tell me otherwise.

timholy:
Thanks so much, @AShedko!

AShedko:
You are welcome.

PR #648, opened 2017-06-18
Eliminate old deprecations

timholy:
We've had these deprecations since the release of Images 0.6 on Jan 30. Does it seem like long enough that we can now delete them?

This also shifts to Julia 0.6, but if folks object that can be reconsidered. ImageCore has already moved, so it might need some work to support 0.5.


timholy:
Feels pretty good to delete 3000 lines of code...

timholy:
Hmm, I was hoping this would fix the OSX timeout. Does anyone with a mac have any insights about what's happening?

Evizero:
Running the tests right now on my macbook and its taking its sweet time.

The merge of #643 surely doesn't help the situation as it suppresses the output of the individual testsets. Basically if the tests take a long time and nothing ever prints until they are done its a recipe for timeouts

That said the tests seem to take a really long time, so maybe there is a deeper issue

Evizero:
Ok so it seems that the tests actually hang after a while. I tracked it down to this section where it just seems to hang forever. everything before that (including all other test files) seem to execute fine

https://github.com/JuliaImages/Images.jl/blob/aa04e63487d46f0e41041e4ca65365ea3aa240ba/test/writemime.jl#L54-L58

This codeblock is sufficient to trigger the effect for me on my macbook:

```julia
using Images, Colors, FixedPointNumbers
using Base.Test

workdir = joinpath(tempdir(), "Images")
if !isdir(workdir)
    mkdir(workdir)
end
img = fill(HSV{Float64}(0.5, 0.5, 0.5), 1, 1)
fn = joinpath(workdir, "writemime.png")
open(fn, "w") do file
    show(file, MIME("image/png"), img, minpixels=0, maxpixels=typemax(Int))
end
```

if press `CTRL+C` it repeats a couple of error messages over and over, including

```
ErrorException("QuartzImageIO: tried to save unknown buffer type ColorTypes.HSV{Float64}")
```

timholy:
Can you do `Pkg.checkout("QuartzImageIO")`? See also https://github.com/JuliaLang/METADATA.jl/pull/9821.

Evizero:
solves the hang problem, but does cause one test to fail.

```
colorspace normalization: Test Failed
  Expression: b isa Matrix{RGB{N0f8}} && b[1] == RGB(1, 0, 0)
Stacktrace:
 [1] macro expansion at /Users/csto/.julia/v0.6/Images/test/writemime.jl:66 [inlined]
```

```julia
julia> b
1×1 Array{RGB4{N0f8},2}:
 RGB4{N0f8}(1.0,0.0,0.0)
```

timholy:
Fixed if we generalize this to `eltype(b) <: AbstractRGB && eltype(eltype(b)) == N0f8 && b[1] == RGB(1, 0, 0)`?

Evizero:
yes. fixed by

```diff
diff --git a/test/writemime.jl b/test/writemime.jl
index 42493bc..ff4d0f2 100644
--- a/test/writemime.jl
+++ b/test/writemime.jl
@@ -63,7 +63,7 @@ using Base.Test
             show(file, MIME("image/png"), img, minpixels=0, maxpixels=typemax(Int))
         end
         b = load(fn)
-        @test isa(b, Matrix{RGB{N0f8}}) && b[1] == RGB(1,0,0)
+        @test eltype(b) <: AbstractRGB && eltype(eltype(b)) == N0f8 && b[1] == RGB(1, 0, 0)  #isa(b, Matrix{RGB{N0f8}}) && b[1] == RGB(1,0,0)
         img = fill(RGBA{Float32}(1,0,0,0.5), 1, 1)
         open(fn, "w") do file
             show(file, MIME("image/png"), img, minpixels=0, maxpixels=typemax(Int))
```

REQUIRE
```diff
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.6.0-pre
```
tkelman: maybe release now?

timholy:
Once we get QuartzImageIO tagged, we'll be back to passing again.

PR #654, opened 2017-07-12
Juno plot pane display

pfitzseb:
Fixes https://github.com/JuliaImages/Images.jl/issues/435.

Open questions: 
- Should a `isdefined(Main, :Juno) && Main.Juno.isactive()` condition in `__init__` be used instead of Requires.jl?
- ~~Is it okay to use `stringmime` directly or would it make sense to call Images' `show` method beforehand to make sure image resizing etc happens?~~ Hm, this already seems to be happening...


timholy:
On 0.6 Requires.jl isn't quite so scary. Since I mostly care about 0.6, I think this is fine.

Thanks so much for your great work on this!

tkelman:
This package still supports 0.5, would it be better to drop 0.5 support before releasing this?

timholy:
@pfitzseb, I was just testing this myself. Do you have to do anything special to get the images to show? I can verify that the method is defined by `methods(Juno, render)` but with `img = testimages("lighthouse")` I'm not seeing anything.

pfitzseb:
Yeah, the display doesn't work in the console right now, only when inline evaluating in the editor. 
Also see the discussion in https://github.com/JunoLab/Atom.jl/pull/93.

PR #661, opened 2017-08-13
Add ImageMorphology as a dependency

### LINKED ISSUE #660: Move morphological operations to a separate package
The number of dependencies in Images.jl has increased a lot since its initial version. For my package ImageQuilting.jl I only need imfilter() and dilate(). The former is available in ImageFiltering.jl, but the latter is still implemented at Images.jl.

Could we move all morphology-related operations to a separate package like ImageMorphology.jl or even to ImageFiltering.jl? That would reduce the number of dependencies tremendously.

juliohm:
Fix #660. Please let me know if any other operation in Images.jl should be migrated to ImageMorphology.jl and I will take care of it.

Also, by looking at the test suite, I was wondering if we could just run the tests for each submodule:

```julia
@testset "Images" begin
  Pkg.test("ImageFiltering")
  Pkg.test("ImageMorphology")
  ...
end
```

Currently the tests are repeated in Images.jl, it would be great to have these encapsulated instead.


juliohm:
Hi all, could you please review and merge? Also, if you can tag a new version, that would be great.

timholy:
> Also, by looking at the test suite, I was wondering if we could just run the tests for each submodule:

That's probably a good idea. The ImageFiltering tests are *much* more thorough than anything here, and they take quite a long time. I could easily imagine the tests taking 5 minutes on a fast local machine, and 20 minutes or more on a travis worker. This would be the only disadvantage. But the bonus is insurance that everything works together, which is pretty big.

juliohm:
Agree Tim, the only issue it the time constraint on Travis, it will probably exceed the limits. We could assume that each submodule is passing the tests and don't test them at the Images.jl level?

In a user machine, however; we could activate all the tests or prompt the user for the tests he is interested in. This switch to find out if we are on Travis or not is very straightforward, and I have done it before here: https://github.com/juliohm/GeoStats.jl/blob/master/test/runtests.jl#L7-L12

We can activate all tests for JuliaImages maintainers or something like that.

PR #663, opened 2017-08-16
Extend dilate/erode for ImageMeta

juliohm:
As discussed in https://github.com/JuliaImages/ImageMorphology.jl/issues/1

I will remove the definitions from ImageMorphology.jl after they are merged here.

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -51,8 +51,9 @@ const is_little_endian = ENDIAN_BOM == 0x04030201
 @reexport using ImageFiltering
 @reexport using ImageMorphology
 
-import ImageTransformations: restrict
 using ImageMetadata: ImageMetaAxis
+import ImageMorphology: dilate, erode
+import ImageTransformations: restrict
 
 using Base.Cartesian  # TODO: delete this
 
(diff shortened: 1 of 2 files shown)


src/Images.jl
```diff
@@ -51,8 +51,9 @@ const is_little_endian = ENDIAN_BOM == 0x04030201
 @reexport using ImageFiltering
 @reexport using ImageMorphology
 
-import ImageTransformations: restrict
 using ImageMetadata: ImageMetaAxis
+import ImageMorphology: dilate, erode
+import ImageTransformations: restrict
```
timholy: What's this one about?

timholy:
I'll wait for it to get through CI, but then this seems fine to me.

src/Images.jl
juliohm: Yes, Git diffs are strange sometimes :)

juliohm:
Hi Tim, can I tag a new version for ImageMorphology.jl?

timholy:
Yes, go for it. It's attobot-enabled.

PR #668, opened 2017-09-20
Updates docs for `findlocalmaxima` to indicate they return `Vector{CartesianIndex}`

ssfrr:


timholy:
Indeed, thanks @ssfrr!

PR #678, opened 2017-11-14
improve component_centroids performance

Evizero:
As discussed in https://github.com/JuliaImages/Images.jl/issues/677#issuecomment-344284124 .

- makes the function type stable
- about 4x speed improvement on the example discussed in the link above

### DIFF
--- a/src/connected.jl
+++ b/src/connected.jl
@@ -272,17 +272,12 @@ function component_subscripts(img::AbstractArray{Int})
 end
 
 "`component_centroids(labeled_array)` -> an array of centroids for each label, including the background label 0"
-function component_centroids(img::AbstractArray{Int})
-    nd = ndims(img)
-    n = [zeros(nd+1) for i=0:maximum(img)]
-    s = size(img)
-    for i=1:length(img)
-        vcur = ind2sub(s,i)
-        vsum = n[img[i]+1]
-        for d=1:nd
-            vsum[d] += vcur[d]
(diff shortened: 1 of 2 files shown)


src/connected.jl
```diff
@@ -272,17 +272,15 @@ function component_subscripts(img::AbstractArray{Int})
...
-    n = [zeros(nd+1) for i=0:maximum(img)]
-    s = size(img)
-    for i=1:length(img)
-        vcur = ind2sub(s,i)
-        vsum = n[img[i]+1]
-        for d=1:nd
-            vsum[d] += vcur[d]
+function component_centroids(img::AbstractArray{Int,N}) where N
+    len = length(0:maximum(img))
+    n = zeros(Int, N+1, len)
```
timholy: I could also imagine making this `n = fill((zero(CartesianIndex{N}), 0), len)` and then writing the inner loop as
```julia
v = img[I] + 1
nv = n[v]
n[v] = (nv[1]+I, nv[2]+1)
```
Thoughts?

Evizero: Ohh, you just gave me an idea involving reshape and a trailing singleton dimenson

Evizero: neither really helps. :(

Evizero:
That's the best looking solution I think. performance doesn't make a difference

timholy:
Merge at will when CI finishes.

Issue #682, opened 2017-11-17
Port corners.jl to ImageFeatures.jl


zygmuntszpak:
I propose that corners.jl, which currently implements harris, shi-tomasi and kitchen_rosenfeld,  be moved to ImageFeatures.jl and be re-written so that it conforms to the conventions of the ImageFeatures.jl package.   I'd be happy to give this a go.

timholy:
Seems right to me. Some thinking about deprecation strategy could be warranted, though. Base has experienced some pains from moving code out to the standard library or packages.

Evizero:
We did do it before reasonably painlessly with ImageTransformations. Or do you not plan to reexport ImageFeatures here?

timholy:
ImageFeatures is not a dependency of Images (it's vice versa).

zygmuntszpak:
I've studied the code of ImageFeatures.jl some more and realised that it does not introduce any new corner detectors. Instead, the package uses the corner extraction functions in the `Images.jl` package. Each corner extraction function produces a binary matrix which indicates whether a particular pixel location is a corner.  As a first step, the `ImageFeatures.jl ` package converts this binary matrix into a `Keypoints` type, for example:  `Keypoints(fastcorners(img1, 12, 0.4))`.  Then, the package facilitates the extraction of a variety of local feature descriptors which are typically based on the neighbourhood pixels of the corner. The package also facilitates using the descriptors to match `Keypoints` between two views. 

The `ImageFeatures.jl ` package appears to make the distinction between a feature (i.e. a corner) and a descriptor for this feature. Hence, perhaps there is no need to move the code for detecting corners from `Images.jl` to `ImageFeatures.jl` after all? 


My ultimate aim is to introduce the ability to compute corners to sub-pixel precision. My current thinking is to keep the `imcorner` function in `Images.jl `and to extend it or to introduce another function such as `imcorner_subpixel`.  One possibility for extending would be to make the `imcorner` function return not just a binary matrix indicating whether a pixel was a corner, but also a matrix that stores the 'corner strength' of each pixel.  We could then introduce a function `pixel2subpixel` or `corner2subcorner` which takes as input the binary matrix and the corner matrix and does the appropriate interpolations to get sub-pixel precision. 

I would then make some changes in the ImageFeatures.jl package so that a Keypoint type need not be an integer. 

[...]

PR #685, opened 2017-12-01
Add sub-pixel corner detection capability #682

zygmuntszpak:
Introduces the function imcorner_subpixel as well as several requisite
support functions to facilitate the ability to detect corners to
sub-pixel precision.

The imcorner_subpixel uses the existing imcorner functionality to obtain
initial corner response values and a binary corner indicator matrix. The
corner response values and the indicator matrix are used in a univariate
quadratic interpolation scheme to determine the corner coordinates to
sub-pixel precision.

The new imcorner_subpixel function returns a length-3 NTuple of type
HomogeneousPoint2D, which represents the homogeneous coordinates of a
planar point.

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -13,11 +13,13 @@ import Base: abs, atan2, clamp, convert, copy, copy!, ctranspose, delete!,
              strides, sum, write, zero
 
 export float32, float64
+export HomogeneousPoint
 
 using Base: depwarn
 using Base.Order: Ordering, ForwardOrdering, ReverseOrdering
 
 using Compat
+using StaticArrays
 
 # "deprecated imports" are below
 
@@ -63,6 +65,52 @@ Indicate that `x` should be interpreted as a [percentile](https://en.wikipedia.o
 - `canny(img, 1.4, (Percentile(80), Percentile(20)))` uses percentiles of the edge magnitude image as threshold
 """
 struct Percentile{T} <: Real p::T end
+
+
+"""
+HomogeneousPoint(x::NTuple{N, T})
+
+In projective geometry [homogeneous coordinates](https://en.wikipedia.org/wiki/Homogeneous_coordinates) are the
+natural coordinates for describing points and lines. 
+
+For instance, the homogeneous coordinates for a planar point are a triplet of real numbers ``(u, v ,w)``, with ``w \\neq 0``. 
+This triple can be associated with a point ``P = (x,y)`` in Cartesian coordinates, where ``x = \\frac{u}{w}`` and ``y = \\frac{v}{w}`` 
+[(more details)](http://www.geom.uiuc.edu/docs/reference/CRC-formulas/node6.html#SECTION01140000000000000000).
+
+In particular, the `HomogeneousPoint((10.0,5.0,1.0))` is the standardised projective representation of the Cartesian 
+point `(10.0,5.0)`. 
+"""
+struct HomogeneousPoint{T <: AbstractFloat,N} 
+    coords::NTuple{N, T}
+end
+
+# By overwriting Base.to_indices we can define how to index into an N-dimensional array
+# given an (N+1)-dimensional [`HomogeneousPoint`](@ref) type. 
+# We do this by converting the homogeneous coordinates to Cartesian coordinates
+# and rounding to nearest integer.
+# 
+# For homogeneous coordinates of a planar point we return  
+# a tuple of permuted Cartesian coordinates, (y,x), since matrices 
+# are indexed  according to row and then column.
+# For homogeneous coordinates of other dimensions we do not permute 
+# the corresponding Cartesian coordinates. 
+Base.to_indices(A::AbstractArray, p::Tuple{<: HomogeneousPoint}) = homogeneous_point_to_indices(p[1])
+
+function homogeneous_point_to_indices(p::HomogeneousPoint{T,3}) where T 
+    if  p.coords[end] == 1
+        return round(Int,  p.coords[2]), round(Int, p.coords[1]) 
+    else
+        return round(Int,  p.coords[2] / p.coords[end]), round(Int, p.coords[1] / p.coords[end]) 
+    end    
+end
+
+function homogeneous_point_to_indices(p::HomogeneousPoint) 
+    if  p.coords[end] == 1
+        return round.(Int, p.coords)
+    else
+        return round.(Int, p.coords ./ p.coords[end])
+    end
+end
 
 include("labeledarrays.jl")
 include("algorithms.jl")
@@ -154,6 +202,8 @@ export # types
     forwarddiffx,
     forwarddiffy,
     imcorner,
+    imcorner_subpixel,
+    corner2subpixel,
     harris,
     shi_tomasi,
     kitchen_rosenfeld,
@@ -244,7 +294,7 @@ Algorithms:
     - Exposure : `imhist`, `histeq`, `adjust_gamma`, `histmatch`, `imadjustintensity`, `imstretch`, `imcomplement`, `clahe`, `cliphist`
     - Gradients: `backdiffx`, `backdiffy`, `forwarddiffx`, `forwarddiffy`, `imgradients`
     - Edge detection: `imedge`, `imgradients`, `thin_edges`, `magnitude`, `phase`, `magnitudephase`, `orientation`, `canny`
-    - Corner detection: `imcorner`, `harris`, `shi_tomasi`, `kitchen_rosenfeld`, `meancovs`, `gammacovs`, `fastcorners`
+    - Corner detection: `imcorner`,`imcorner_subpixel`, `harris`, `shi_tomasi`, `kitchen_rosenfeld`, `meancovs`, `gammacovs`, `fastcorners`
     - Blob detection: `blob_LoG`, `findlocalmaxima`, `findlocalminima`
     - Morphological operations: `dilate`, `erode`, `closing`, `opening`, `tophat`, `bothat`, `morphogradient`, `morpholaplace`, `feature_transform`, `distance_transform`, `convexhull`
     - Connected components: `label_components`, `component_boxes`, `component_lengths`, `component_indices`, `component_subscripts`, `component_centroids`
--- a/src/corner.jl
+++ b/src/corner.jl
@@ -37,6 +37,142 @@ end
 
 """
 ```
+corners = imcorner_subpixel(img; [method])
+         -> Vector{HomogeneousPoint{Float64,3}}
+corners = imcorner_subpixel(img, threshold, percentile; [method])
+         -> Vector{HomogeneousPoint{Float64,3}}
+```
+
+Same as [`imcorner`](@ref), but estimates corners to sub-pixel precision.
+
+Sub-pixel precision is achieved by interpolating the corner response values using
+the 4-connected neighbourhood of a maximum response value. 
+See [`corner2subpixel`](@ref) for more details of the interpolation scheme. 
+
+"""
+function imcorner_subpixel(img::AbstractArray; method::Function = harris, args...)
+    responses = method(img; args...)
+    corner_indicator = similar(img, Bool)
+    fill!(corner_indicator, false)
+    maxima = map(CartesianIndex, findlocalmaxima(responses))
+    for m in maxima corner_indicator[m] = true end
+    corners = corner2subpixel(responses,corner_indicator)
+end
+
+function imcorner_subpixel(img::AbstractArray, threshold; method::Function = harris, args...)
+    responses = method(img; args...)
+    corner_indicator = map(i -> i > threshold, responses)
+    corners = corner2subpixel(responses,corner_indicator)
+end
+
+function imcorner_subpixel(img::AbstractArray, thresholdp::Percentile; method::Function = harris, args...)
+    responses = method(img; args...)
+    threshold = StatsBase.percentile(vec(responses), thresholdp.p)
+    corner_indicator = map(i -> i > threshold, responses)
+    corners = corner2subpixel(responses,corner_indicator)
+end
+
+
+"""
+```
+corners = corner2subpixel(responses::AbstractMatrix,corner_indicator::AbstractMatrix{Bool})
+        -> Vector{HomogeneousPoint{Float64,3}}
+```
+Refines integer corner coordinates to sub-pixel precision.
+
+The function takes as input a matrix representing corner responses
+and a boolean indicator matrix denoting the integer coordinates of a corner
+in the image. The output is a vector of type [`HomogeneousPoint`](@ref)
+storing the sub-pixel coordinates of the corners.
+
+The algorithm computes a correction factor which is added to the original
+integer coordinates. In particular, a univariate quadratic polynomial is fit
+separately to the ``x``-coordinates and ``y``-coordinates of a corner and its immediate
+east/west, and north/south neighbours. The fit is achieved using a local
+coordinate system for each corner, where the origin of the coordinate system is
+a given corner, and its immediate neighbours are assigned coordinates of  minus
+one and plus one.
+
+The corner and its two neighbours form a system of three equations. For example,
+let  ``x_1 = -1``,  ``x_2 = 0`` and  ``x_3 = 1`` denote the local ``x`` coordinates
+of the west, center and east pixels and let the vector ``\\mathbf{b} = [r_1, r_2, r_3]``
+denote the corresponding corner response values. With
+
+```math 
+    \\mathbf{A} = 
+        \\begin{bmatrix}
+            x_1^2 & x_1  & 1  \\\\
+            x_2^2 & x_2  & 1 \\\\
+            x_3^2 & x_3  & 1 \\\\
+        \\end{bmatrix},
+```
+the coefficients of the quadratic polynomial can be found by solving the
+system of equations ``\\mathbf{b} = \\mathbf{A}\\mathbf{x}``. 
+The result is given by ``x = \\mathbf{A}^{-1}\\mathbf{b}``.
+
+The vertex of the quadratic polynomial yields a sub-pixel estimate of the
+true corner position. For example, for a univariate quadratic polynomial
+``px^2 + qx + r``, the ``x``-coordinate of the vertex is ``\\frac{-q}{2p}``. 
+Hence, the refined sub-pixel coordinate is equal to:
+ ``c +  \\frac{-q}{2p}``, where ``c`` is the integer coordinate. 
+
+!!! note
+    Corners on the boundary of the image are not refined to sub-pixel precision.
+
+"""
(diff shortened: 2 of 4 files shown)


src/Images.jl
```diff
@@ -18,6 +18,7 @@ using Base: depwarn
 using Base.Order: Ordering, ForwardOrdering, ReverseOrdering
 
 using Compat
+using StaticArrays
```
Evizero: minor comment: this needs a corresponding entry in the `REQUIRE` file

zygmuntszpak: Thanks, I will amend the commit and add the entry in the `REQUIRE` file.  I will also add some more tests as per the *codecov* suggestions. I'm not sure what to do about the failing tests on the win32 build. It seems to be a matter of precision because the approximately equal symbol is failing. Do you have any suggestions for that?

zygmuntszpak:
Looking at the failed build for `Win32 `and the debug info I printed I no longer think it is failing because of precision. Could it be something more fundamental in `Base`? The stack trace seems to suggest there is something wrong with the statement `all(ids[i].coords .≈ corner_pts[i].coords)` in general, and not that the approximately true statement doesn't hold. 

```
 Expression: all(ids[i].coords .≈ corner_pts[i].coords)
Stacktrace:
 [1] macro expansion at C:\Users\appveyor\.julia\v0.6\Images\test\corner.jl:110 [inlined]
 [2] macro expansion at .\test.jl:860 [inlined]
 [3] macro expansion at C:\Users\appveyor\.julia\v0.6\Images\test\corner.jl:33 [inlined]
 [4] macro expansion at .\test.jl:860 [inlined]
 [5] anonymous at .\<missing>:?
```

src/Images.jl
```diff
@@ -64,6 +66,33 @@ Indicate that `x` should be interpreted as a [percentile](https://en.wikipedia.o
...
+
+# By overwriting Base.to_indices we can define how to index into a matrix
+# given a `HomogeneousPoint2D` type. We do this by rounding the values of 
+# the `HomogeneousPoint2D` type  to the nearest integer.
+# We assume that the homogeneous coordinates have been standardised
+# so that the third coordinate equals one. Otherwise we would have to
+# divide each coordinate by the last coordinate everytime we attempt to index
+# into a matrix which could induce a performance hit. Note also that we return
+# a tuple of (y,x) since matrices are indexed according to row and then column.
+Base.to_indices(A::AbstractArray, p::Tuple{<: HomogeneousPoint2D}) = round(Int,  p[1].coords[2]), round(Int, p[1].coords[1])
```
Evizero: Just to double check: `p` is limited to a 1-element tuple, but this is on purpose right?

It may be interesting to note (but no need to change it here) that `Tuple` get special treatment in Julia. For example [they are a covariant type](https://docs.julialang.org/en/stable/manual/types/#Tuple-Types-1). This means that you could actually write `p::Tuple{HomogeneousPoint2D}` because for any `T <: HomogeneousPoint2D` it holds that `Tuple{T} <: Tuple{HomogeneousPoint2D}`

src/corner.jl
```diff
@@ -35,6 +35,140 @@ function imcorner(img::AbstractArray, thresholdp::Percentile; method::Function =
...
+```
+
+Same as [`imcorner`](@ref), but estimates corners to sub-pixel precision.
+
+Sub-pixel precision is achieved by interpolating the corner response values using
+the 4-connected neighbourhood of a maximum response value. 
+See [`corner2subpixel`](@ref) for more details of the interpolation scheme. 
+
+"""
+function imcorner_subpixel(img::AbstractArray; method::Function = harris, args...)
```
Evizero: @timholy Do (or should) we have a convention for naming catch-all optional positional or named arguments?

Personally I adapted the convention of using `args...` for positional arguments and `kw...` for keyword arguments.

src/corner.jl
```diff
@@ -35,6 +35,140 @@ function imcorner(img::AbstractArray, thresholdp::Percentile; method::Function =
...
+
+!!! note
+    Corners on the boundary of the image are not refined to sub-pixel precision.
+
+"""
+function corner2subpixel(responses::Matrix{Float64}, corner_indicator::Array{Bool,2})
+    nrows,ncolumns = size(corner_indicator)
+    row, col, _ = findnz(corner_indicator)
+    ncorners = length(row)
+    corners = fill(Images.HomogeneousPoint2D((0.0,0.0,0.0)),ncorners)
```
Evizero: Let's remove the `Images.` prefix here, since we are within the `Images` module

src/corner.jl
```diff
@@ -35,6 +35,140 @@ function imcorner(img::AbstractArray, thresholdp::Percentile; method::Function =
...
+
+
+"""
+```
+corners = corner2subpixel(responses::Matrix{Float64},corner_indicator::Array{Bool,2})
+        -> Vector{HomogeneousPoint2D{3,Float64}}
+```
+Refines integer corner coordinates to sub-pixel precision.
+
+The function takes as input a matrix reresenting corner responses
```
Evizero: typo `representing`

src/corner.jl
```diff
@@ -35,6 +35,140 @@ function imcorner(img::AbstractArray, thresholdp::Percentile; method::Function =
...
+```math 
+\mathbf{A} = 
+    \begin{bmatrix}
+        x_1^2 & x_1  & 1  \\
+        x_2^2 & x_2  & 1 \\
+        x_3^2 & x_3  & 1 \\
+    \end{bmatrix}
+```
+the coefficients of the quadratic polynomial can be found by solving the
+system of equations `b = Ax`. The result is given by `x = inv(A)b`.
```
Evizero: Since you used the math environment in the previous paragraph I think its worth considering doing it here for `b = Ax` and below as well (the two backticks environment instead of one)

test/corner.jl
```diff
@@ -29,6 +29,118 @@ using Base.Test, Images, Colors, FixedPointNumbers
     	for id in ids expected_corners[id] = 0 end
     	@test sum(corners .!= expected_corners) < 3
     end
+	
+	@testset "Corners Sub-pixel API" begin
```
Evizero: These are tabs I think. Could you change this to use spaces instead?

Evizero:
This looks like a nicely made addition. I left a couple of minor and style related comments. I am afraid I don't know much about the subject itself. Maybe someone else here could review the functionality.

src/Images.jl
zygmuntszpak: Yes, `p` is limited to a 1-element tuple. SimonDanish explained why in one of his replies to my [question](https://github.com/JuliaImages/ImageFeatures.jl/issues/45#issuecomment-345680172)

I have quoted him below: 

>That's indeed confusing and is the aftermath of indexing a multi dimensional array with a single element. `to_indices` takes a tuple because it gets called called like this in the abstract array code:

>#note, that indices isn't splatted, so it turns into a tuple:
`getindex(A::AbstractArray, indices...) = getindex(A, to_indices(indices)...)` 
so calling `getindex(A, Point(1.5, 1.5))` will call `to_indices(A, (Point(1.5, 1.5),))`.

Thanks for your remark about *covariant* types, I will do some more reading on the subject.

src/corner.jl
zygmuntszpak: In this instance I just copied and pasted the current definition of `imcorner`, so whatever change you suggest I will then also apply to the existing `imcorner` code.

src/corner.jl
zygmuntszpak: Thanks for spotting that, I thought I had systematically removed any reference to Images.

src/corner.jl
zygmuntszpak: Sure, no worries. Do you know how I might visualise what the LaTeX markup will look like? My version of Juno does not support Latex preview, and I didn't see any such functionality in Visual Studio Code.  I'm not sure how to build the documentation so that I can actually see the LaTeX markup.

Evizero: Do you have access to Jupyter? I think it does the appropriate rendering for docstrings

test/corner.jl
zygmuntszpak: How can you tell that they are tabs not spaces? I was using tab, but I thought that the editors (VS Code and Juno) were set up to automatically convert tabs to spaces.

Evizero: Just by selecting the text with the mouse basically. You'll see that you can just select one giant white space. In contrast if using spaces it would select a bunch of small whitespaces

src/corner.jl
timholy: Images developed over a long enough period of time that there isn't as much consistency as would be desirable. I also tend to use `args...` for positional, and `kw...` or `kwargs...` for keywords.

timholy:
I'll review this, but first I'd really like to thank @Evizero. In recent months my administrative responsibilities at work have increased enormously, to the point where it's hard for me to put much time into development. Having help with some of the review/bugfix/merge responsibilities will allow me to focus on higher-level issues, and collectively we'll ensure that JuliaImages keeps thriving!

Evizero:
happy to help where my skillset allows it. You were a great tutor to me when I started working with Images.jl, so I'd like to give back how I can (though my knowledge about image processing is very spotty)

src/Images.jl
```diff
@@ -64,6 +66,33 @@ Indicate that `x` should be interpreted as a [percentile](https://en.wikipedia.o
 """
 struct Percentile{T} <: Real p::T end
 
+
+"""
+HomogeneousPoint2D(x::NTuple{3, T})
```
timholy: Would it be worth calling this
```julia
struct HomogeneousPoint{T<:AbstractFloat,N}
    coords::NTuple{N,T}
end
```
The concept is a general one, even if its first application (`imcorner`) is currently limited to 2d.

src/Images.jl
timholy: This is one of these painful decisions. From a performance standpoint I completely understand why you've done it, OTOH it breaks the core equivalence-class property of `HomogeneousPoint`. How bad is the hit if you add `@assert p.coords[end] == 1`? `@assert` statements get skipped if you start julia with `julia -O3`.

src/corner.jl
```diff
@@ -35,6 +35,140 @@ function imcorner(img::AbstractArray, thresholdp::Percentile; method::Function =
...
+    threshold = StatsBase.percentile(vec(responses), thresholdp.p)
+    corner_indicator = map(i -> i > threshold, responses)
+    corners = corner2subpixel(responses,corner_indicator)
+end
+
+
+"""
+```
+corners = corner2subpixel(responses::Matrix{Float64},corner_indicator::Array{Bool,2})
+        -> Vector{HomogeneousPoint2D{3,Float64}}
```
timholy: I see you considered having an extra parameter to `HomogeneousPoint` already :slightly_smiling_face:. Whatever gets finally decided, make sure this docstring is in sync.

src/corner.jl
```diff
@@ -35,6 +35,140 @@ function imcorner(img::AbstractArray, thresholdp::Percentile; method::Function =
...
+```
+Refines integer corner coordinates to sub-pixel precision.
+
+The function takes as input a matrix reresenting corner responses
+and a boolean indicator matrix denoting the integer coordinates of a corner
+in the image. The output is a vector of type `HomogeneousPoint2D`
+(an `NTuple{3, Float64}`) storing the sub-pixel coordinates of the corners.
```
timholy: change to ``[`HomogeneousPoint`](@ref)`` and drop the statement in parentheses

src/corner.jl
```diff
@@ -35,6 +35,140 @@ function imcorner(img::AbstractArray, thresholdp::Percentile; method::Function =
...
+"""
+function corner2subpixel(responses::Matrix{Float64}, corner_indicator::Array{Bool,2})
+    nrows,ncolumns = size(corner_indicator)
+    row, col, _ = findnz(corner_indicator)
+    ncorners = length(row)
+    corners = fill(Images.HomogeneousPoint2D((0.0,0.0,0.0)),ncorners)
+    invA = @SMatrix [0.5 -1.0 0.5; -0.5 0.0 0.5; 0.0 1.0 -0.0]
+    for k = 1:ncorners
+        # Corners on the perimeter of the image will not be interpolated.
+        if  row[k] == 1 || row[k] == nrows || col[k] == 1 || col[k] == ncolumns
```
timholy: `imcorner` works even for [non-1 arrays](https://julialang.org/blog/2017/04/offset-arrays):
```julia
img = testimage("cam")
imgo = OffsetArray(img, (-2, 5))
imcorner(imgo)
```
It would be better to compare to the `first`/`last` values of `indices(corner_indicator)`.

src/corner.jl
```diff
@@ -35,6 +35,140 @@ function imcorner(img::AbstractArray, thresholdp::Percentile; method::Function =
...
+true corner position. For example, for a univariate quadratic polynomial
+`px^2 + qx + r`, the x-coordinate of the vertex is `-q/(2.0p)`. Hence, the refined
+sub-pixel coordinate is equal to: `c +  -q/(2.0p)`, where `c` is the integer
+coordinate. 
+
+!!! note
+    Corners on the boundary of the image are not refined to sub-pixel precision.
+
+"""
+function corner2subpixel(responses::Matrix{Float64}, corner_indicator::Array{Bool,2})
```
timholy: I would favor more general choices like `corner_indicator::AbstractMatrix{Bool}`. See below (re `indices`) to see how other types might arise.

src/corner.jl
zygmuntszpak: Since I have followed the current convention of `imcorner`, how about we leave it as `args` for the purpose of this commit? We could raise a new issue for systematically checking the Image package for consistent usage of args and kw? I just noticed that there appear to be several places where `kwargs` should have been used, such as gradcovs.

In the meantime, I will address the other issues that `Evizero` has pointed out and amend the current commit.

src/corner.jl
```diff
@@ -35,6 +35,140 @@ function imcorner(img::AbstractArray, thresholdp::Percentile; method::Function =
...
+Returns the value of a matrix at given coordinates together with the values
+of the north, south, east and west neighbours.
+
+This function does not perform bounds checking it is up to the user to ensure
+that the function is not called with indices that are on the boundary of the
+matrix.
+
+"""
+function unsafe_neighbourhood_4(matrix::Matrix{Float64},r::Int,c::Int)
+    @views center = matrix[r,c]
```
timholy: I don't think you need the `@views` here, you're just extracting a single value from the array.

test/corner.jl
```diff
@@ -29,6 +29,118 @@ using Base.Test, Images, Colors, FixedPointNumbers
...
+		@test pts[4].coords[3] ≈ 1
+
+		# Check imcorners_subpixel API.
+		img = zeros(20, 20)
+		img[4:17, 4:17] = 1
+		img[8:13, 8:13] = 0
+
+		ids = map(HomogeneousPoint2D,
+                    [(4.244432486132958, 4.244432486132958, 1.0),
+					 (4.244432486132958, 16.755567513867042, 1.0),
```
timholy: indentation issues

test/corner.jl
```diff
@@ -188,4 +300,4 @@ using Base.Test, Images, Colors, FixedPointNumbers
 
 end
 
-nothing
+nothing
```
timholy: Add blank line at end

src/corner.jl
zygmuntszpak: In light of your comment regarding `OffsetArray`, I presume I have to change `matrix::Matrix{Float64}` to `matrix::Union{OffsetArrays.OffsetArray{Float64,2,Array{Float64,2}}, Matrix{Float64}}`? 

Similarly, `responses::Matrix{Float64}` would need to become `responses::Union{OffsetArrays.OffsetArray{Float64,2,Array{Float64,2}}, Matrix{Float64}}`?

timholy: Why put a type on at all? (There is no performance advantage to specifying the type.) The only thing that adding the type does here is ensure that a `MethodError` gets thrown by an inappropriate call to this function, rather than some error thrown by, e.g., the indexing operations inside the body of this function.

It wouldn't be crazy to restrict it to two-dimensional `AbstractArray`s, though:
```julia
unsafe_neighbourhood_4(matrix::AbstractMatrix, r::Int, c::Int)
```

zygmuntszpak:
The latest pull-request addresses all the comments and suggestions. If I have left anything out I did so unintentionally.

src/Images.jl
```diff
@@ -64,6 +66,52 @@ Indicate that `x` should be interpreted as a [percentile](https://en.wikipedia.o
...
+# and rounding to nearest integer.
+# 
+# For homogeneous coordinates of a planar point we return  
+# a tuple of permuted Cartesian coordinates, (y,x), since matrices 
+# are indexed  according to row and then column.
+# For homogeneous coordinates of other dimensions we do not permute 
+# the corresponding Cartesian coordinates. 
+Base.to_indices(A::AbstractArray, p::Tuple{<: HomogeneousPoint}) = homogeneous_point_to_indices(A, p)
+         
+function homogeneous_point_to_indices(A::AbstractArray, p::Tuple{<: HomogeneousPoint})
```
timholy: This might be prettier if you extract the HomogeneousPoint in `to_indices` and just pass that:

```julia
Base.to_indices(A::AbstractArray, p::Tuple{<: HomogeneousPoint}) = homogeneous_point_to_indices(p[1])
```
because then you don't have to write `p[1]` everywhere in `homogeneous_point_to_indices`.

A more typical "Julian" style would be to avoid the branch and provide two methods:
```
function homogeneous_point_to_indices(p::HomogeneousPoint{T,3}) where T
    if p.coords[end] == 1
    ...
end
function homogeneous_point_to_indices(p::HomogeneousPoint)
    if p.coords[end] == 1
    ...
end
```
However, the compiler should be smart enough that it will evaluate your `if (length(p[1]).coords) == 3` at compile time so it avoids introducing another branch (in other words, in terms of generated code yours will be the same as mine). 

All these are optional, just wanted you to know.

src/corner.jl
```diff
@@ -35,6 +35,142 @@ function imcorner(img::AbstractArray, thresholdp::Percentile; method::Function =
...
+    threshold = StatsBase.percentile(vec(responses), thresholdp.p)
+    corner_indicator = map(i -> i > threshold, responses)
+    corners = corner2subpixel(responses,corner_indicator)
+end
+
+
+"""
+```
+corners = corner2subpixel(responses::AbstractMatrix,corner_indicator::AbstractMatrix{Bool})
+        -> Vector{HomogeneousPoint{3,Float64}}
```
timholy: Parameters are swapped (should be `Float64,3`).

zygmuntszpak:
Done. Thank you for all the excellent advice.

timholy:
Sorry I forgot to hit merge! Thanks again!

zygmuntszpak:
@timholy If I want this to appear when I do `Pkg.update("Images")` do I have to wait for you to tag a release? Do you have a policy for tagging releases?

timholy:
Just tagged (will have to wait for the METADATA.jl PR to be merged). Thanks for the reminder!

PR #689, opened 2017-12-15
Update exposure.jl

yakir12:
Added a comment to `complement`'s docstring

### DIFF
--- a/src/exposure.jl
+++ b/src/exposure.jl
@@ -26,7 +26,8 @@ imstretch(img::ImageMeta, m::Number, slope::Number) = shareproperties(img, imstr
     y = complement(x)
 
 Take the complement `1-x` of `x`.  If `x` is a color with an alpha channel,
-the alpha channel is left untouched.
+the alpha channel is left untouched. Don't forget to add a dot when `x` is
+an array: `complement.(x)`
 """
 complement(x) = oneunit(x)-x
 complement(x::TransparentColor) = typeof(x)(complement(color(x)), alpha(x))


Evizero:
was this something you yourself ran into?

I have mixed feeling about this, since there are a number of places where this little addition is equally relevant. Yet broadcast is a language level feature and pointing this out in a docstring seems on first glance redundant.

That said I suppose it doesn't hurt and if it helps users not to get stuck its a valuable addition

yakir12:
I ran into this today. Took me an embarrassingly long time to figure it out. 

The problem I guess is that `complement(x::Array)` doesn't issue a warning or throws an error, but does this:
```julia
julia> a = Gray.([.1 .5; .6 .9])
2×2 Array{Gray{Float64},2}:
 Gray{Float64}(0.1)  Gray{Float64}(0.5)
 Gray{Float64}(0.6)  Gray{Float64}(0.9)

julia> complement(a)
2×2 Array{Gray{Float64},2}:
 Gray{Float64}(0.9)   Gray{Float64}(-0.5)
 Gray{Float64}(-0.6)  Gray{Float64}(0.1)
``` 
Which is wrong (right?). 
So we need to either add a depreciation warning that in the future the function will not work on arrays without a dot, or change it's behavior somehow (or the docstring).

Evizero:
right. looks like you uncovered a deeper issue here. Maybe we should change the function to

```julia
complement(x::Union{Number,Colorant}) = oneunit(x)-x
```

and add a deprecated version 

```julia
@deprecate complement(x) complement.(x)
```

?

yakir12:
That sounds great (to me). Unrelated: wouldn't it be great to have an alias for `Union{Number,Colorant}`? I mean, I didn't check, but I bet there about a trillion instances of similar type signatures that could easily be aliased to something nicer... Just a thought.

Evizero:
> Unrelated: wouldn't it be great to have an alias for Union{Number,Colorant}? I mean, I didn't check, but I bet there about a trillion instances of similar type signatures that could easily be aliased to something nicer... Just a thought.

It's a natural idea, but I don't think this particular alias would add much value. Its already short and descriptive and I don't think it is that common.

Note that there are actually a couple of similar alias floating around the code base, so its not that far fetched. These are two I can think off the back of my head.

- https://github.com/JuliaImages/ImageCore.jl/blob/master/src/ImageCore.jl#L25
- https://github.com/JuliaImages/ImageTransformations.jl/blob/master/src/interpolations.jl#L9

---

Concerning the change I suggested. It would be interesting to hear the opinion of @timholy on this. It might be that making the docstring change you proposed is more in line with his intention of the function.

Thanks for taking the time to make this PR, by the way

yakir12:
I see about the alias 👍 

After getting help [here](https://discourse.julialang.org/t/simplifying-contributions-to-documentation/7785?u=yakir12), it became super easy. I'm looking forward to making more of these micro contributions.

Evizero:
I am going ahead with merging this for now as there is no downside to it.

We may want to discuss the complement behaviour in a new issue/pr

PR #692, opened 2017-12-20
Update algorithms.jl w/ laplacian pyramid methods

mcbaron:
Added methods to allow for a Laplacian pyramid decomposition and reconstruction of an image. 
Respects the convention of the Gaussian pyramid code

### DIFF
--- a/docs/src/core.md
+++ b/docs/src/core.md
@@ -56,7 +56,7 @@ instead using the functions `data` and `properties` to extract these.)  These
 are the only two fields in the first concrete image type, called `Image`:
 
 ```julia
-type Image{T,N,A<:AbstractArray} <: AbstractImageDirect{T,N}
+mutable struct Image{T,N,A<:AbstractArray} <: AbstractImageDirect{T,N}
     data::A
     properties::Dict{String,Any}
 end
@@ -79,7 +79,7 @@ images. More detail about this point can be found below.
 The only other concrete image type is for indexed images:
 
 ```julia
-type ImageCmap{T,N,A<:AbstractArray,C<:AbstractArray} <: AbstractImageIndexed{T,N}
+mutable struct ImageCmap{T,N,A<:AbstractArray,C<:AbstractArray} <: AbstractImageIndexed{T,N}
     data::A
     cmap::C
     properties::Dict{String,Any}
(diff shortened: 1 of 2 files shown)


zygmuntszpak:
Thank you for taking the time to make this contribution. Would it be possible for you to please add some corresponding tests in /test/algorithms.jl file?

zygmuntszpak:
@mcbaron Do you still have a bit of free time to work on this?

mcbaron:
Zygmunt, I would like to, but I am currently swamped with other work at the
moment. I will not be able to write appropriate tests until the end of
April.

I realize now that I should've written tests before submitting the pull
request. I will know to do that in the future!

rapus95:
Hey, may I bump this and ask whether there are some news? 🙈

timholy:
Just FYI we already have `restrict` which should be a much faster variant of the same thing (though it's limited to two-fold scale-reduction).

mcbaron:
@Tim Given how long this PR has been open, I figured that something else would've been written for this by now. 

I don't know how valuable my re-digging into the code will be, given the changes that have happened in Julia since the initial PR date.

timholy:
In fact `restrict` first appeared in 2014, though in 2017 it was moved to ImageTransformations (which gets loaded by Images)!

I will close this, but if anyone thinks this variant adds something not obtainable by `restrict` feel free to create an updated version.

timholy:
Ah wait, I was reading quickly (big backlog of email) and just noticed this is Laplacian pyramid rather than Gaussian pyramid. `restrict` is more along the lines of gaussian, this would add something new. But we can wait until it has a champion.

rapus95:
I don't know if that is a good idea but maybe there could be some collaboration between Images.jl and [Pyramids.jl](https://github.com/loganwilliams/Pyramids.jl)? This at least would prevent code duplication.

timholy:
We could delete `gaussian_pyramid` if that's what you have in mind? Though someone would have to put some effort into Pyramids.jl to make it work for, e.g., `Colorant` arrays (it doesn't currently).

`restrict` is so much faster than the alternatives, that's not on the table for removal. Using the `earth` image from TestImages (3k by 3k):
```julia
julia> @time gaussian_pyramid(img, 4, 2, 1.5);
  0.703564 seconds (5.70 k allocations: 823.213 MiB, 24.38% gc time)

julia> @time restrict(restrict(restrict(img)));
  0.088495 seconds (46 allocations: 101.612 MiB, 6.44% gc time)
```
I use it on much bigger images than that.

PR #694, opened 2017-12-26
Migrate Hausdorff distance to ImageDistances.jl

juliohm:
Dear all,

This pull request migrates the Hausdorff distance to ImageDistances.jl.

All the best,


Evizero:
I can see the sense in doing this, but we should think of some deprecation strategy.

juliohm:
Thank you @Evizero, what deprecation strategy to adopt?

1. Warning pointing to ImageDistances.jl
2. Warning + making ImageDistances.jl a dependency

Please let me know and I will update the PR.

Best,

Evizero:
This change is breaking, so maybe we should put the code deletion off until the 0.7 update.

To handle deprecation gracefully I suggest to not yet delete the function (or anything related to it), but instead just add a depwarn to the first line of `hausdorff_distance`.

Maybe something like this:

```julia
Base.depwarn("hausdorff_distance(imgA, imgB) is deprecated, use <equivalent signature in ImageDistances> of the package ImageDistances.jl instead" , :hausdorff_distance)
```

That would mean that for now the function doesn't actually use your new implementation, but it does tell the user to make the switch. When 0.7 comes, or sometime thereafter, we can then just remove all the related code (so basically we can then merge a PR like this). 

Note that we could at that later point also decide to alternatively just reexport ImageDistances and thus make it a Images.jl dependency. That said, let us consider that option at that later point.

timholy:
How big do you think ImageDistances will become? How many methods do you think it might ultimately export? If the answer is "a few" then perhaps we can just `@reexport` it. If you think it might include dozens of functions, then I think it's better to deprecate and move out.

juliohm:
I don't know of many distances between images, the answer should be a few. I am happy with any of the options, we can reassess them more carefully after Julia v0.7. :+1:

juliohm:
@Evizero could you please take a look at this PR? There is also #695.

Evizero:
oh, thanks for the reminder. forgot about this

Evizero:
I'll take care of this issue this week. sorry for the long delay

Evizero:
I'll go ahead and merge this and implement the reexport strategy outlined by Prof Holy in https://github.com/JuliaImages/Images.jl/pull/695#issuecomment-362914924

PR #696, opened 2017-12-27
make complement signature more strict. fix #690

### LINKED ISSUE #690: Behaviour of complement when passing an array
@yakir12 noticed in #689 that the function `complement` doesn't quite behave as expected when passing an array to it.

```julia
julia> A = [.1 .9; .6 .3]
2×2 Array{Float64,2}:
 0.1  0.9
 0.6  0.3

julia> complement.(A)
2×2 Array{Float64,2}:
 0.9  0.1
 0.4  0.7

julia> complement(A)
2×2 Array{Float64,2}:
  0.9  -0.9
 -0.6   0.7
```

Even though this behaviour doesn't really go against what the docstring says, I would still consider this behaviour a bug, or, given how easy it is to run into this, I would at least consider this very confusing.

A potential solution could be to change the function to:

```julia
complement(x::Union{Number,Colorant}) = oneunit(x)-x
@deprecate complement(x) complement.(x)
```

Evizero:
This is one attempt to improve the situation outlined in #690 

I have to admit I am not super happy about this either. That said, I do think we should not let `complement(::Array)` silently return something strange.

### DIFF
--- a/src/exposure.jl
+++ b/src/exposure.jl
@@ -29,8 +29,9 @@ Take the complement `1-x` of `x`.  If `x` is a color with an alpha channel,
 the alpha channel is left untouched. Don't forget to add a dot when `x` is
 an array: `complement.(x)`
 """
-complement(x) = oneunit(x)-x
+complement(x::Union{Number,Colorant}) = oneunit(x)-x
 complement(x::TransparentColor) = typeof(x)(complement(color(x)), alpha(x))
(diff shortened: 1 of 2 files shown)


Evizero:
CI error looks like https://github.com/JuliaImages/ImageAxes.jl/issues/25. We should probably fix that first before merging this

Evizero:
lets see if CI is happy now

timholy:
Thanks again!

PR #699, opened 2018-02-18
Adds more details and an example to the histeq documentation

zygmuntszpak:


zygmuntszpak:
@kmsquire When you get a moment, could you please have a read over this and merge it if you think it is ok?

src/exposure.jl
```diff
@@ -100,17 +100,96 @@ hist_equalised_img = histeq(img, nbins, minval, maxval)
...
+identically distributed random variables. Specifically, let the sample space
+``\\Omega`` be the set of all ``IJ``-tuples ``\\omega
+=(\\omega_{11},\\omega_{12},\\ldots,\\omega_{1J},\\omega_{21},\\omega_{22},\\ldots,\\omega_{2J},\\omega_{I1},\\omega_{I2},\\ldots,\\omega_{IJ})``,
+where each ``\\omega_{ij} \\in \\{0,1,\\ldots, L-1 \\}``. Furthermore, impose a
+probability measure on ``\\Omega`` such that the functions ``\\Omega \\ni
+\\omega \\to \\omega_{ij} \\in \\{0,1,\\ldots,L-1\\}`` are independent and
+identically distributed.
+
+One can then regard an image as a matrix of random variables ``\\mathbf{G} =
+[G_{i,j}(\\omega)]``, where each function ``G_{i,j}: \\Omega \\to \\mathbb{R}``
```
Evizero: are the `[` and `]` rendered ? I think to remember that one has to escape them in order to show up. I might be mixing up parenthesis though

zygmuntszpak: I can confirm that the `[` and `]` are rendering in IJulia.

Evizero:
I went ahead with the merge. Thanks!

PR #703, opened 2018-03-09
Added clearborder function

arijitkar98:
@zygmuntszpak @Evizero I have added the clearborder function that clears object that are connected to the border of the image.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -937,22 +937,17 @@ function clearborder(img::AbstractArray, width::Int=1, background::Int=0)
     outerrange = CartesianRange(map(i -> 1:i, dimensions))
     innerrange = CartesianRange(map(i -> 1+width:i-width, dimensions))
 
-    border_indices = Set{Int64}()
+    border_labels = Set{Int64}()
     for i in EdgeIterator(outerrange,innerrange)
-        push!(border_indices, labels[i])
+        push!(border_labels, labels[i])
     end
 
-    function vectorin(a, b)
-        bset = Set(b)
-        [i in bset for i in a]
-    end
-
-    label_mask = vectorin(range(0,number), border_indices)
-
-    new_img = copy(img)
+    new_img = similar(img)
     for itr in eachindex(labels)
-        if label_mask[labels[itr] + 1] == true
+        if labels[itr] in border_labels
             new_img[itr] = background
+        else
+        	new_img[itr] = img[itr]
         end
     end
 
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -863,3 +863,45 @@ function otsu_threshold(img::AbstractArray{T, N}, bins::Int = 256) where {T<:Uni
 
     return T((edges[thres-1]+edges[thres])/2)
 end
+
+function clearborder(img::AbstractArray, width::Int=1, background::Int=0)
+    
+    for i in size(img)
+        if(width > i)
+            throw(ArgumentError("Border width must not be greater than size of the image."))
+        end
+    end
+
+    connectivity = ntuple(i -> 3, ndims(img))
+    labels = label_components(img,trues(connectivity))
+    number = maximum(labels) + 1
+
+    dimensions = size(img)
+    outerrange = map(i -> 1:i, dimensions)
+    innerrange = map(i -> 1+width:i-width, dimensions)
+
+    outerrange = CartesianRange(outerrange)
+    innerrange = CartesianRange(innerrange)
+
+    border_elements = []
+    for i in EdgeIterator(outerrange,innerrange)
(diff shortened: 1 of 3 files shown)


arijitkar98:
Ping @zygmuntszpak @Evizero

test/algorithms.jl
```diff
@@ -565,6 +565,72 @@ using Base.Test
...
+        @test cleared_img == 10*ones(img)
+
+        #Grayscale input image Case
+        img = [1 2 3 1 2
+               3 3 5 4 2
+               3 4 5 4 2
+               3 3 2 1 2]
+        cleared_img = clearborder(img)
+        check_img = [0 0 0 0 0
+                     0 0 5 4 0
```
Evizero: shouldn't it be `0 3 5 4 0` ?

arijitkar98: The 3 at [2,2] will be a part of the label containing ([1,3], [2,1], [2,2], [3,1], [4,1], [4,2]) and thus is a part of an object touching the border.....so it should be cleared i.e. made 0

src/algorithms.jl
```diff
@@ -863,3 +863,62 @@ function otsu_threshold(img::AbstractArray{T, N}, bins::Int = 256) where {T<:Uni
...
+        bset = Set(b)
+        [i in bset for i in a]
+    end
+
+    label_mask = vectorin(indices, border_indices')
+    mask = map(i -> label_mask[i+1], labels[:])
+    reshape(mask, size(labels))
+
+    new_img = copy(img)
+    new_img[mask] = background
```
kmsquire: This works fine, and as a first pass is probably sufficient.  

However, if you wanted to make it more efficient, you could eliminate the `mask` allocation all together, by creating an empty image, and then filling each element in either with `background` or the original image value, depending on the label value.

The `eachindex` function is a nice way to generalize this to any number of dimensions.

arijitkar98:
@kmsquire I have made the required changes.

src/algorithms.jl
```diff
@@ -903,3 +903,59 @@ function yen_threshold(img::AbstractArray{T, N}, bins::Int = 256) where {T<:Unio
...
+    number = maximum(labels) + 1
+
+    dimensions = size(img)
+    outerrange = CartesianRange(map(i -> 1:i, dimensions))
+    innerrange = CartesianRange(map(i -> 1+width:i-width, dimensions))
+
+    border_indices = Set{Int64}()
+    for i in EdgeIterator(outerrange,innerrange)
+        push!(border_indices, labels[i])
+    end
```
kmsquire: `border_indices` contains the labels near the border, right?  How about renaming it to `border_labels`?

src/algorithms.jl
```diff
@@ -903,3 +903,59 @@ function yen_threshold(img::AbstractArray{T, N}, bins::Int = 256) where {T<:Unio
...
+    function vectorin(a, b)
+        bset = Set(b)
+        [i in bset for i in a]
+    end
+
+    label_mask = vectorin(range(0,number), border_indices)
+
+    new_img = copy(img)
+    for itr in eachindex(labels)
+        if label_mask[labels[itr] + 1] == true
```
kmsquire: Here, `== true` is redundant--you can just write 

```
if label_mask[labels[itr] +1]
```

That said, it seems that this is the same as
```
if labels[itr] in border_indices  # or border_labels, if you rename it
```

If you write it this way, you can dispense with `label_mask` and `vectorin` all together.

src/algorithms.jl
```diff
@@ -903,3 +903,59 @@ function yen_threshold(img::AbstractArray{T, N}, bins::Int = 256) where {T<:Unio
...
+    end
+
+    function vectorin(a, b)
+        bset = Set(b)
+        [i in bset for i in a]
+    end
+
+    label_mask = vectorin(range(0,number), border_indices)
+
+    new_img = copy(img)
```
kmsquire: Minor, but here, you copy the image, and below, you iterate through each index, which iterates through the whole image twice (once on the copy, and once below)

As an alternative, you could allocate `new_img = similar(img)`, and then fill in background or the original value in the loop below.

(The effect of this would be minor for small images, but might affect timing for large images/volumes, etc.)

src/algorithms.jl
arijitkar98: Yes...i will change it

arijitkar98:
@kmsquire I have made the changes you asked. Thank you for the code reviews

arijitkar98:
Ping @kmsquire @Evizero

zygmuntszpak:
I don't have write access to the repository so I can't merge this for you.

Evizero:
thank you @arijitkar98 for the contribution! and thank you for taking the time @kmsquire to review

PR #704, opened 2018-03-09
fix order of dimensions when showing an array of images

shashi:
Before 
screenshot from 2018-03-09 20-31-08

After:

screenshot from 2018-03-09 20-29-36

@alanedelman suggested the parenthetic in the vector display:

screenshot from 2018-03-09 19-58-16

### DIFF
--- a/src/showmime.jl
+++ b/src/showmime.jl
@@ -78,6 +78,7 @@ function _show_even(io::IO, m::MIME"text/html", imgs::AbstractArray{T, N}, cente
 end
 
 function Base.show(io::IO, m::MIME"text/html", imgs::AbstractArray{T, N}) where {T<:ColorantMatrix, N}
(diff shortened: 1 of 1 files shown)


rdeits:
This seems good. I didn't make this change in #627 precisely because of the inconsistency with the way a vector is displayed, but I agree that it's easier to explain why a vector is shown as a row than to explain why an N-dimensional array is shown transposed.

Evizero:
I like this change a lot. I also ran into this before, actually. Not that it was much of an issue to be fair. I have been a fan of your "show array of images as a table of images" either way ([example](https://nbviewer.jupyter.org/github/Evizero/Augmentor.jl/blob/gh-pages/generated/mnist_knet.ipynb))

Evizero:
Thanks!

shashi:
Cheers! :-)

PR #705, opened 2018-03-10
Adding radon transform

tosalonijain:
I have tried building a rough code for radon transform and am expecting guidance and changes from your side to finish it.

### DIFF
--- a/src/radon.jl
+++ b/src/radon.jl
@@ -1,0 +1,84 @@
+using Images, ImageProjectiveGeometry, ImageFiltering, ImageView, PaddedViews, Interpolations
+
+function warpfast(img::AbstractArray, M::AbstractArray, output_shape::Int64= 0, order::Int64 =1, mode="constant", cval::Int64 =0)
+    mode_c = Int64(uppercase(mode[1]))
+
+    println("axes  ", length(axes(img,1))," type  ", typeof(length(axes(img,1))))
+    if output_shape == 0
+        out_r = length(axes(img,1))
+        out_c = length(axes(img,2))
+    else
+        out_r = output_shape
+        out_c = output_shape
+    end
+
+    out = zeros(Int64,(out_r, out_c))
+
+    
+    newimg = imgtrans(img, M)
+    
+    if order == 3
+        
+        itp = interpolate(newimg, (BSpline(Constant()), Constant()), OnGrid())
+    elseif order == 1
+        
+        itp = interpolate(newimg, (BSpline(Linear()), Linear()), OnGrid())
+    elseif order == 2
+        itp = interpolate(newimg, (BSpline(Quadratic()), Quadratic()), OnGrid())
+    elseif order == 3
+        itp = interpolate(newimg, (BSpline(Cubic()), Cubic()), OnGrid())
+    end
+    for tfr in range(out_r)
+        for tfc in range(out_c)
+            out[tfr, tfc] = itp[tfr, tfc]
+        end
+    end
+    return out
+end
+
+ogrid(ys, xs) = [[[xs[i]] for i in 1:size(xs,1)], [ys[j] for j in 1:size(ys)]]
+
+function radon(image, theta= "nothing", circle= "nothing")
+    if theta == "nothing"
+        theta = collect(1:1:180)
+    end
+    if circle == "nothing"
+        diagonal = sqrt(2) * max(size(image,1),size(image,2))
+        pad = [Int64(ceil(diagonal - s)) for s in size(image)]
+        new_center = [Int64((s + p) / 2) for (s, p) in zip(size(image), pad)]
+        old_center = [Int64(s / 2) for s in size(image)]
+        pad_before = [nc - oc for (oc, nc) in zip(old_center, new_center)]
+        pad_width = [(pb, p - pb) for (pb, p) in zip(pad_before, pad)]
+        new_tup= (pad[1], pad[2])
+        println(-pad_width[1][1]+1,"   ", size(image,1)+pad_width[1][2],"  and  ", -pad_width[2][1]+1,"   ", size(image,2)+pad_width[2][2])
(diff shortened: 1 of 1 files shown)


Evizero:
Hi! Thank you for your interest to contribute!

Did you create this according to some reference? I am asking because I lack the domain knowledge to put this into context. From a pure package-maintainer standpoint I will say that I don't think its a good idea to introduce a couple new dependencies for a single function (especially `ImageView`).

zygmuntszpak:
Have a look at Algorithm 2.1 'The Discrete Normal Radon Transform` page 28 (47 in PDF) in
the PhD thesis of Peter Toft. I would be expecting something along those lines...

tosalonijain:
@Evizero @zygmuntszpak , yes I have used the scipy radon transform as a reference to frame this code. I am sorry  I did involved too many dependencies, I would cut the unnecessary ones. 
I had some queries to give this code the final finishing. It would be great if I could get some guidance from your side..:)

tosalonijain:
@Evizero @zygmuntszpak

zygmuntszpak:
Could you please outline the steps in the algorithm that you are trying to implement? Its not clear to me why you would need a projective transformation. The only explanation I can come up with for what you might be trying to do is that you are summing the rows for each column of the image, rotating the image slightly, summing the rows for each column of the rotated image etc. If that is the case, you don't need a projective warp but just a simple rotation. Presumably when you rotate the image slightly the new coordinates of the pixels don't fall onto the pixel grid, and so you need to fill the pixel grid of the rotated image by interpolating? 

The algorithm that I referenced in my previous post follows a different approach and seems like it can be implemented almost verbatim in Julia. Why not give that one a go? 

image

tosalonijain:
@zygmuntszpak , I am sorry I was just trying a different approach to find the radon transform. I was following the code base of scipy radon transform. 
Here, I have implemented a similar approach as suggested by you. Please provide your views and required changes, documentation etc for this code. 
If it seems okay to you, I shall soon write a test code for radon transform. 
And it would be great if you could provide some similar sources to apply inverse radon over the radon image produced by it.
Thankyou.

zygmuntszpak:
Chapter 8 of the thesis I referenced deals with inverting the radon transform using the _filtered back-projection_ technique. The procedure is split across 3 algorithms (8.1, 8.2 and 8.3). The implementation of the inverse radon transforms appears to be substantially more involved than the radon transform.

zygmuntszpak:
I was wondering why you implemented something that was _similar_ to the algorithm outlined above but not identical. I see that you are actually using [Tim Holy's implementation](https://discourse.julialang.org/t/radon-and-iradon-transforms/2407/19)...

tosalonijain:
@zygmuntszpak actually earlier only Tim had referenced me that implementation. But I was interested to employee interpolations on both x and y. Then finally since I could not build a feasible code for that so, I decided to use that only. I shall be writing the iradon code, as per your suggestions. Beside that the documentation and test for radon needs to be done also. Seeking your guidance.
Thankyou.

zygmuntszpak:
So far I haven't used the Radon transform in my own work so I don't have a deep understanding of it. Hence it would take me a fair bit of time to discern how Tim's implementation relates to the Algorithm that I cited above. Perhaps you could add some comments to clarify the steps of the algorithm? 

If I were doing this for the first time, I would start by implementing something where the key ideas are explained and the pseudo-code is detailed such as the Algorithm I pasted above. Then, upon receiving Tim's implementation, I would first check that both implementations give my roughly the same results. If they do, I would start teasing apart Tim's implementation to see why it is faster etc. 

There are some things that I personally find a bit odd in Tim's implementation. For example, instead of `w, h = size(A)` I would have thought that `h, w = size(A)` would be more appropriate since the usual convention is that rows of a matrix determine the height and the columns determine the width.

Perhaps someone else might have better insight into why `w,h` was chosen instead. The discourse thread I referenced above also has an implementation of the inverse radon transform which you might be able to use as a basis. 

Also, it might be worth checking out the ImageJ implementation of the [radon transforms](https://imagej.nih.gov/ij/plugins/radon-transform.html). Since ImageJ is used heavily by the medical imaging community I'm pretty sure a lot of thought went into the optimality and correctness of the implementation.

tosalonijain:
@zygmuntszpak I have added comments in order to help explanation of the code. 
There's slight variation from the reference you provided because the pseudo code mentioned in the thesis works for radon transform assuming the image to be square whereas Tim's code deals with rectangular images also. Although the governing idea is same for both. 
I am waiting for the changes to be made according to you and also shall I start writing the test code for this? I would work hard to get this PR get merged asap.
Thankyou.

zygmuntszpak:
Thank you, Saloni. I reckon you may as well start working on writing the tests for this.

tosalonijain:
@zygmuntszpak I am having trouble in how to claim for a given matrix or image, the radon transform found is correct without employing the inverse radon on that produced image?
It would be great if you could offer few suggestions in this regard for writing the test code.

zygmuntszpak:
You can test the radon transform on a simple shape for which an analytic expression of the radon transform is known. For example, you could consider a unit disk (filled circle) centred at the origin. 
A formula for what the radon transform should produce on such shape can be found on page 2 of [these lecture notes](https://statweb.stanford.edu/~candes/math262/Lectures/Lecture09.pdf). Look at the sentence "Then an easy calculation gives...". 

I'm sorry but I won't be able to assist you with the OpenCV project.

zygmuntszpak:
Thanks for adding such detailed comments to the code–they help tremendously!

tosalonijain:
@zygmuntszpak , I am sorry I am a a bit busy making my proposal for OpenCV.jl project, so could not implement the test code. But I would very soon be completing this work of mine.
Thank you.

zygmuntszpak:
Are you still busy with some of the tests?

tosalonijain:
@zygmuntszpak I am sorry I could not address your comment. Yes actually I am pretty swamped with my university tests and exams. I would for sure implement the test code soon. Pardon for the passiveness.

Thank you.

ashwani-rathee:
This is also very unlikely to be ever merged given radon transform is provided by https://github.com/JuliaImages/ImageReconstruction.jl

PR #706, opened 2018-03-18
Added imfill function

arijitkar98:
@zygmuntszpak @Evizero Added the imfill function that fills holes in the image that are smaller than a specified size.

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -224,6 +224,7 @@ export # types
     otsu_threshold,
     yen_threshold,
     clearborder,
+    imfill,
 
     #Exposure
     complement,
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -957,9 +957,9 @@ end
 
 """
 ```
-filled_img = clearborder(img, size)
-filled_img = clearborder(img, size, value)
-filled_img = clearborder(img, size, value, connectivity)
+filled_img = imfill(img, size)
+filled_img = imfill(img, size, value)
+filled_img = imfill(img, size, value, connectivity)
 ```
 
 Returns a copy of the original image after filling holes that are smaller than the specified size.
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -954,3 +954,51 @@ function clearborder(img::AbstractArray, width::Int=1, background::Int=0)
     return new_img
 
 end
+
+"""
+```
+filled_img = clearborder(img, size)
+filled_img = clearborder(img, size, value)
+filled_img = clearborder(img, size, value, connectivity)
+```
+
+Returns a copy of the original image after filling holes that are smaller than the specified size.
+
+Parameters:
+
+ -  img            = Binary/Grayscale input image (Integer valued)
+ -  size           = Minimum size of holes that are not filled.
+ -  value          = Value to be given to the holes that are filled (Default value is 1)
+ -  connectivity   = connectivity takes the same values as in label_components (Default value is 1:ndims(img))
+
+"""
+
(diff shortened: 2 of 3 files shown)


zygmuntszpak:
Thanks for yet another contribution! Two things immediately come to mind. First, you made a simple copy-paste mistake. The docstring for the function mentions 'clearborder'. 

The second potential issue could be the your representation of images as integer matrices and forcing the `value` parameter to be an integer. From what I understand the image data is always [a value between 0 and 1](https://juliaimages.github.io/latest/quickstart.html), no matter what the underlying type that is used to represent the image. Hence, I am not sure whether your model of an image as integer numbers is adequate here. Perhaps someone else with more familiarity might be able to correct me if I'm wrong.

arijitkar98:
@zygmuntszpak If I remove the constraint of images as integer matrices, I think the algorithm should work fine as label_components will work properly even for float values

arijitkar98:
If I do this, I am confused whether I should allow the user to be able to pass in another parameter which specifies a threshold and the pixels with intensities below this threshold would be considered as holes or should I just consider only pixels with zero intensities as holes?

kmsquire:
I'll verify what @zygmuntszpak said, that image values always take on values between 0 and 1.

Because this interpretation of images is quite different from what you see in other languages, I suggest reading the [Images.jl documentation](https://juliaimages.github.io/latest/index.html), especially the section [arrays, numbers, and colors](https://juliaimages.github.io/latest/arrays_colors.html).

Regarding dealing with gray-scale images, consider following the Matlab functionality, where a "hole" in a gray-scale image is defined as an area of dark pixels surrounded by lighter pixels.  
(Interpreting that mathematically/programmatically should hopefully be an interesting challenge!)

zygmuntszpak:
The particular technique you are after is called `Morphological Reconstruction'.  See [this excerpt](https://cn.mathworks.com/tagteam/64199_91822v00_eddins_final.pdf) and [this paper](http://www2.vincent-net.com/luc/papers/93ieeeip_recons.pdf)  for an example.  MATLAB implements the *fast hybrid reconstruction* algorithm which is described in the aforementioned paper by Luc Vincent.

In light of the link with mathematical morphology, I wonder if `imfill` function shouldn't live in ImageMorphology.jl instead?

zygmuntszpak:
Sorry @kmsquire, my intention for clicking on your name was not to request a review but to send you a PM.

kmsquire:
@zygmuntszpak, no worries!

ImageMorphology does seem like a reasonable place to put this, especially if it can use any of the machinery already there.  @juliohm, since that repo is all your work, any thoughts?

juliohm:
Thank you for pinging @kmsquire , the work in ImageMorphology.jl is actually Tim's work. We did split the package out of Images.jl to reduce the number of dependencies.

@arijitkar98 would it be fair to call your method an inpainting algorithm? We are just including another package in the JuliaImages organization soon for all sorts of "filling in" operations with texture: https://github.com/juliohm/ImageInpainting.jl Maybe it would be a good idea to submit the PR there? What do you think?

I have added a few basic examples of the API in the README. The idea is to have a type for an inpainting algorithm and dispatch on it.

juliohm:
@arijitkar98 it seems that the `imfill` function doesn't require a mask. In that case I think it is appropriate to have it in ImageMorphology.jl as @kmsquire suggested.

arijitkar98:
@zygmuntszpak @juliohm I am closing this PR and will send one in ImageMorphology.jl

I am currently going through the above mentioned papers and will try to implement that `holes` condition that @kmsquire described from MATLAB

PR #708, opened 2018-03-25
Added thinning function

### LINKED ISSUE #603: Thining and Skeletonozation
Is there a plan/objection to implement Thining, Skeletonozation and pruning as part morphological operations?

vernwalrahul:
Implemented thinning function using the technique of Zhang-Suen. fixes #603

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -224,6 +224,7 @@ export # types
     otsu_threshold,
     yen_threshold,
     clearborder,
+    thinning,
 
     #Exposure
     complement,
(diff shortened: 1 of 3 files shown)


kmsquire:
@vernwalrahul, thank you for the contribution. 

Before this is reviewed, I’ll point out that morphological operations have been moved to ImageMorphology.jl, and that this algorithm should probably be submitted there. 

(The corresponding issue and any other image morphology related issues should also probably be moved there.)

PR #711, opened 2018-03-28
Fixes adjust_gamma so that it works on offset arrays 

### LINKED ISSUE #686: adjust_gamma doesn't work on offset arrays
The title:
```julia
using Images
img = rand(Gray, 5,5)
I = adjust_gamma(img, 1/2.2) # works fine
imgp = padarray(img, Fill(0, (2,2)))
I = adjust_gamma(imgp, 1/2.2) # error:
ERROR: size not supported for arrays with indices (-1:7, -1:7); see http://docs.julialang.org/en/latest/devdocs/offset-arrays/
```

zygmuntszpak:
Fixes #686 
###  Implementation Note
Calling `indices(img)` on a canonical array returns a tuple of type `Base.OneTo`. Unfortunately, there is currently no `fill` method defined in `Base` which can take a tuple of `Base.OneTo` as its dimension argument. Hence, I had to explicitly convert the tuple of `Base.OneTo` to a tuple of integers.

### DIFF
--- a/src/exposure.jl
+++ b/src/exposure.jl
@@ -270,7 +270,7 @@ adjust_gamma(img::AbstractArray{T}, gamma::Number) where {T<:Number} = _adjust_g
 adjust_gamma(img::AbstractArray{T}, gamma::Number) where {T<:Colorant} = _adjust_gamma(img, gamma, T)
 
 function _adjust_gamma(img::AbstractArray, gamma::Number, C::Type)
-    gamma_corrected_img = zeros(C, size(img))
+    gamma_corrected_img = _fill(oneunit(C), indices(img))
     for I in eachindex(img)
         gamma_corrected_img[I] = _gamma_pixel_rescale(img[I], gamma)
     end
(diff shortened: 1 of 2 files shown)


src/exposure.jl
```diff
@@ -270,15 +270,25 @@ adjust_gamma(img::AbstractArray{T}, gamma::Number) where {T<:Number} = _adjust_g
 adjust_gamma(img::AbstractArray{T}, gamma::Number) where {T<:Colorant} = _adjust_gamma(img, gamma, T)
 
 function _adjust_gamma(img::AbstractArray, gamma::Number, C::Type)
-    gamma_corrected_img = zeros(C, size(img))
+    if eltype(indices(img)) <: Base.OneTo
+        dim = map((x) -> x.stop, indices(img))
+    else
+        dim = indices(img)
+    end
+    gamma_corrected_img = fill(oneunit(C), dim)
```
Evizero: I think the more julian approach would be to map a custom `fill` function that solves this via dispatch

e.g.:

```julia
_fill(val, dim) = fill(val, dim) # fallback
_fill(val, dim::NTuple{N,Base.OneTo}) where {N} = fill(val, map(length, dim))
```

and then use this function instead:

```julia
gamma_corrected_img = _fill(oneunit(C), indices(img))
```

src/exposure.jl
```diff
@@ -270,15 +270,25 @@ adjust_gamma(img::AbstractArray{T}, gamma::Number) where {T<:Number} = _adjust_g
...
 end
 
 function adjust_gamma(img::AbstractArray{T}, gamma::Number, minval::Number, maxval::Number) where T<:Number
-    gamma_corrected_img = zeros(Float64, size(img))
+    if eltype(indices(img)) <: Base.OneTo
+        dim = map((x) -> x.stop, indices(img))
+    else
+        dim = indices(img)
+    end
+    gamma_corrected_img = fill(oneunit(T), dim)
```
Evizero: same here

Evizero:
thanks!

PR #714, opened 2018-04-11
add support for FileIO's new metadata method

bjarthur:
https://github.com/JuliaIO/ImageMagick.jl/pull/111 might have to be merged and tag for this to pass tests.

### DIFF
--- a/test/writemime.jl
+++ b/test/writemime.jl
@@ -19,6 +19,7 @@ using Test
         end
         b = load(fn)
         @test b == A
+        @test Images.metadata(fn) == (reverse(size(A)), Gray{N0f8})
 
         img = fill(RGB{N0f16}(1,0,0), 1, 1)
         open(fn, "w") do file
@@ -26,6 +27,7 @@ using Test
         end
         b = load(fn)
         @test b == img
+        @test Images.metadata(fn) == (reverse(size(img)), RGB{N0f16})
 
         A = N0f8[0.01 0.99; 0.25 0.75]
         fn = joinpath(workdir, "writemime.png")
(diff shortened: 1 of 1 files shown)


bjarthur:
the relevant PRs to ImageMagick and FileIO are now merged.  this PR should work after they are both tagged.

Evizero:
Thanks! Can you (or someone) post here when the tags are in metadata? I will then retrigger CI and merge

Oh, speaking of tags, the minimum version of FileIO would then need a bumb in REQUIRE. But I could also take care of that little detail after this PR is merged

bjarthur:
please retrigger CI as FileIO and ImageMagick have been tagged:

https://github.com/JuliaLang/METADATA.jl/pull/15632#event-1724807383
https://github.com/JuliaLang/METADATA.jl/pull/15257

thanks!

bjarthur:
running CI again

bjarthur:
tests pass on julia 0.6 for windows and linux.  quartzimageio.jl just needs metadata support for this PR to work on os x...

rsrock:
Is there any reason this has to be in Images?  I have a few issues:

1) `metadata` is too useful of a name to grab of this.  I use it all the time for more elaborate information about experimental conditions. Something like `imgsize` and `imgtype` may be better?  I wold also suggest _not_ exporting these names.

2) I don't think that QuartzImageIO can be refactored to separately grab image info, then the image data, without repeating a lot of work.  It would be like peeling an onion to get to that layer you want (with the metadata), reassembling it, then peeling it again to a deeper layer to get the data.  I don't look forward to that 😄

timholy:
I think I'm with @rsrock on this one. As a non-exported method of ImageMagick it's fine, but as a general method it would need a more specific name.

bjarthur:
i've removed the export of `metadata` and modified the tests accordingly.  can think about an alternate name if desired at another time.

johnnychen94:
Oops... I made a mistake by accidentally dropped the commits when resolving the conflicts locally. @bjarthur do you happen to have a local copy of it?

PR #716, opened 2018-04-14
Adds boundaries function

arijitkar98:
Added the boundaries function that returns a Boolean array where the boundaries between the labeled regions in the image are true.

I will add the tests soon.


arijitkar98:
I have added the subpixel method and all the test cases. I have used a primitive padding function so as not to add the dependency of PaddedViews

arijitkar98:
@zygmuntszpak @Evizero

zygmuntszpak:
Hi Arijit, 

Unfortunately I've fallen ill and don't have the energy at the moment to study this. I will try to get back to you on it later this week when  I am, hopefully, feeling better.

zygmuntszpak:
Hi Arijit, 

Could you please clarify the purpose of these functions some more? What are the typical use cases of having a 'thick' border, 'inner' border and 'subpixel' border? Are you basing this particular function and terms  on some publication or textbook?  

I think the function needs more detailed documentation. I couldn't determine what thick, inner or subpixel meant without studying your test cases. It might be worthwhile including the test cases as example cases in the docstring. 

The thick and inner border tests currently don't test with more than one labelled region. What is the expected behaviour for something like
```
00000000
00000000
00112200
00112200
00000000
```
or 

```
00000000
00002200
00110000
00110000
00000000
```
I guess this will depend on whether one is using 4 or 8 connectivity?

What happens with the 'thick' option when a label is right up against the border? e.g.

```
11000000
11000000
00000000
00000000
00000000
```

Does this function work with `Offset` arrays? 

You mentioned you don't want to add a dependency on `paddedviews`, but `paddedviews` is already [part of the Images package](https://juliaimages.github.io/latest/conversions_views.html#Adding-padding-1). Perhaps it might be worthwhile to reuse its functionality after all?  

At the moment a user of the function indicates that type of border they want based on an integer. I think the Images API tends to use either a string, such as "Inner" or "Thick", or alternatively a Symbol such as `:inner` or `:thick`. Check out  `padarray` from the `ImageFiltering` package for an example of what I mean. It would be great to keep consistency between the packages. 

Thank you for all your efforts.

arijitkar98:
Hi @zygmuntszpak 

> Are you basing this particular function and terms on some publication or textbook?

I was going through similar implementations of this function in MATLAB and Scipy to see the different use cases of this function when I came accross these different methods in the Scipy implementation and so I included them to provide greater functionality.

I will add more extensive test cases and change padding function and argument types soon.

kmsquire:
Hi @arijitkar98,

> I was going through similar implementations of this function in MATLAB and Scipy to see the different use cases of this function when I came accross these different methods in the Scipy implementation and so I included them to provide greater functionality.

When referencing other code, even just to read it, anything you write in Julia will likely be considered "derivative code", and must be placed under the original license.  (If you just looked at the interface to the code, without actually reading the implementation, that's fine.)

In particular, it is highly recommended not to look at MATLAB code unless it is available under a permissive license and not obtained from Matlab File Exchange (see the [TOU, section 2.a.iii](https://www.mathworks.com/matlabcentral/termsofuse.html#content).  Original code included with MATLAB cannot be read at all if you want to implement similar functions.

(An example of "good" matlab code is Peter Kovesi's http://www.peterkovesi.com/matlabfns/, which are available under a permissive license directly from the author.)

If you did read any MATLAB code not publicly available under a permissive license, then this PR cannot be merged.  If the code is publicly available, then that license needs to be included here.

Similarly, if you based any of this code off of Scipy, then you must include the full copyright from that code along with this PR (and update the LICENSE file in this repository accordingly).

arijitkar98:
@kmsquire I had mainly referred to the scipy docs for the API; and about the Licence, the scikit-image repo contains a [modified BSD licence](https://github.com/scikit-image/scikit-image#license-modified-bsd).

I am not sure what is to be done here. Can this PR be merged?

kmsquire:
@arijitkar98, I apologize for not getting back to you sooner.  This fell off of my radar.

> @kmsquire I had mainly referred to the scipy docs for the API; and about the Licence, the scikit-image repo contains a modified BSD licence.

Okay.  If you didn't actually look at any code, then it seems fine to me!  If you based any part of it off of the scikit-image code, then you need to include the `scikit-image` license, and indicate that this function is under that license.  

If you need to do that (it doesn't sound like it), then the license should be included in the `LICENSE.md` file at the end, and should indicate which function(s) it applies to.  You should also include something like the following in the comment in the comment block for the function.

```
This code was translated from the scikit-image version of the same function, and is available under a modified BSD license.  See LICENSE.md for details.
```

@timholy was mostly absent when we were discussing this.  If he and @zygmuntszpak do not have any additional comments, then it can be merged.  Can you fix the merge conflicts?

timholy:
Since scikit-image is BSD, that's fine.

> @kmsquire I had mainly referred to the scipy docs for the API; and about the Licence, the scikit-image repo contains a modified BSD licence.

But [earlier in this thread](https://github.com/JuliaImages/Images.jl/pull/716#issuecomment-386261728) you mentioned you had "gone through" implementations in MATLAB. We need to respect copyright; my hunch is that we can't merge this. Really sorry. Hopefully someone else will contribute a "clean" implementation.

PR #722, opened 2018-05-23
Adds OffsetArrays capability to gaussian_pyramid

arijitkar98:
Ping @zygmuntszpak @Evizero

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1,4 +1,5 @@
 using Base: indices1, tail
+using OffsetArrays
 
 Compat.@dep_vectorize_2arg Gray atan2
 Compat.@dep_vectorize_2arg Gray hypot
@@ -791,25 +792,19 @@ by a factor `downsample` and `sigma` for the gaussian kernel.
 
 """
 
-function gaussian_pyramid(img::AbstractArray{T,N}, n_scales::Int, downsample::Real, sigma::Real) where {T, N}
+function gaussian_pyramid(img::AbstractArray{T,N}, n_scales::Int, downsample::Real, sigma::Real) where {T,N}
     kerng = KernelFactors.IIRGaussian(sigma)
     kern = ntuple(d->kerng, Val{N})
     gaussian_pyramid(img, n_scales, downsample, kern)
 end
 
-function gaussian_pyramid(img::AbstractArray{T,N}, n_scales::Int, downsample::Real, kern::NTuple{N,Any}) where {T, N}
+function gaussian_pyramid(img::AbstractArray{T,N}, n_scales::Int, downsample::Real, kern::NTuple{N,Any}) where {T,N}
     # To guarantee inferability, we make sure that we do at least one
     # round of smoothing and resizing
     img_smoothed_main = imfilter(img, kern, NA())
-    if all(x->isa(x, Base.OneTo), indices(img))
-        img_scaled = pyramid_scale(img_smoothed_main, downsample)
-        prev = convert(typeof(img_scaled), img)
-        pyramid = typeof(img_scaled)[prev]
-    else
-        img_scaled = pyramid_scale(img_smoothed_main, downsample, map(i -> i.start, indices(img_smoothed_main)))
-        prev = convert.(typeof.(img_scaled)[1], img)
-        pyramid = [prev]
-    end
+    img_scaled = pyramid_scale(img_smoothed_main, downsample)
+    prev = convert(typeof(img_scaled), img)
+    pyramid = typeof(img_scaled)[prev]
     if n_scales ≥ 1
         # Take advantage of the work we've already done
         push!(pyramid, img_scaled)
@@ -817,28 +812,22 @@ function gaussian_pyramid(img::AbstractArray{T,N}, n_scales::Int, downsample::Re
     end
     for i in 2:n_scales
         img_smoothed = imfilter(prev, kern, NA())
-        if all(x->isa(x, Base.OneTo), indices(img))
-            img_scaled = pyramid_scale(img_smoothed, downsample)
-        else
-            img_scaled = pyramid_scale(img_smoothed, downsample, map(i -> i.start, indices(img_smoothed)))
-        end
+        img_scaled = pyramid_scale(img_smoothed, downsample)
         push!(pyramid, img_scaled)
         prev = img_scaled
     end
     pyramid
 end
 
-function pyramid_scale(img::AbstractArray{T,N}, downsample::Real) where {T, N}
+function pyramid_scale(img, downsample)
     sz_next = map(s->ceil(Int, s/downsample), size(img))
     imresize(img, sz_next)
 end
 
-function pyramid_scale(img::AbstractArray{T,N}, downsample::Real, start::Tuple) where {T, N}
-    sz_next = map(s->ceil(Int, s/downsample), length.(indices(img)))
-    temp_img = imresize(img, sz_next)
-    trans = Translation(map(i -> ceil((1-i)/2), start)...)
-    resized_img = warp(temp_img, trans)
-    return resized_img
+function pyramid_scale(img::OffsetArray, downsample)
+    sz_next = map(s->ceil(Int, s/downsample), length.(Compat.axes(img)))
+    off = (.-ceil.(Int,(.-start.(Compat.axes(img).-(1,1)))./downsample))
+    OffsetArray(imresize(img, sz_next), off)
 end
 
 """
@@ -941,7 +930,7 @@ Parameters:
 """
 
 function clearborder(img::AbstractArray, width::Int=1, background::Int=0)
-    
+
     for i in size(img)
         if(width > i)
             throw(ArgumentError("Border width must not be greater than size of the image."))
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -790,19 +790,26 @@ Returns a  gaussian pyramid of scales `n_scales`, each downsampled
 by a factor `downsample` and `sigma` for the gaussian kernel.
 
 """
-function gaussian_pyramid(img::AbstractArray{T,N}, n_scales::Int, downsample::Real, sigma::Real) where {T,N}
+
+function gaussian_pyramid(img::AbstractArray{T,N}, n_scales::Int, downsample::Real, sigma::Real) where {T, N}
     kerng = KernelFactors.IIRGaussian(sigma)
     kern = ntuple(d->kerng, Val{N})
     gaussian_pyramid(img, n_scales, downsample, kern)
 end
 
-function gaussian_pyramid(img::AbstractArray{T,N}, n_scales::Int, downsample::Real, kern::NTuple{N,Any}) where {T,N}
+function gaussian_pyramid(img::AbstractArray{T,N}, n_scales::Int, downsample::Real, kern::NTuple{N,Any}) where {T, N}
     # To guarantee inferability, we make sure that we do at least one
     # round of smoothing and resizing
     img_smoothed_main = imfilter(img, kern, NA())
-    img_scaled = pyramid_scale(img_smoothed_main, downsample)
-    prev = convert(typeof(img_scaled), img)
-    pyramid = typeof(img_scaled)[prev]
+    if all(x->isa(x, Base.OneTo), indices(img))
+        img_scaled = pyramid_scale(img_smoothed_main, downsample)
+        prev = convert(typeof(img_scaled), img)
+        pyramid = typeof(img_scaled)[prev]
+    else
+        img_scaled = pyramid_scale(img_smoothed_main, downsample, map(i -> i.start, indices(img_smoothed_main)))
+        prev = convert.(typeof.(img_scaled)[1], img)
+        pyramid = [prev]
+    end
     if n_scales ≥ 1
         # Take advantage of the work we've already done
         push!(pyramid, img_scaled)
@@ -810,16 +817,28 @@ function gaussian_pyramid(img::AbstractArray{T,N}, n_scales::Int, downsample::Re
     end
     for i in 2:n_scales
         img_smoothed = imfilter(prev, kern, NA())
-        img_scaled = pyramid_scale(img_smoothed, downsample)
+        if all(x->isa(x, Base.OneTo), indices(img))
+            img_scaled = pyramid_scale(img_smoothed, downsample)
+        else
+            img_scaled = pyramid_scale(img_smoothed, downsample, map(i -> i.start, indices(img_smoothed)))
+        end
         push!(pyramid, img_scaled)
         prev = img_scaled
     end
     pyramid
 end
 
-function pyramid_scale(img, downsample)
+function pyramid_scale(img::AbstractArray{T,N}, downsample::Real) where {T, N}
     sz_next = map(s->ceil(Int, s/downsample), size(img))
     imresize(img, sz_next)
+end
+
+function pyramid_scale(img::AbstractArray{T,N}, downsample::Real, start::Tuple) where {T, N}
+    sz_next = map(s->ceil(Int, s/downsample), length.(indices(img)))
+    temp_img = imresize(img, sz_next)
+    trans = Translation(map(i -> ceil((1-i)/2), start)...)
+    resized_img = warp(temp_img, trans)
+    return resized_img
 end
 
 """
--- a/REQUIRE
+++ b/REQUIRE
@@ -20,3 +20,4 @@ FileIO
 Compat 0.19
 StatsBase 0.14   # for histrange
 Requires
+OffsetArrays
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -54,6 +54,7 @@ using ImageMetadata: ImageMetaAxis
 import ImageMorphology: dilate, erode
 import ImageTransformations: restrict
 using TiledIteration: EdgeIterator
+using CoordinateTransformations: Translation
 
 using Base.Cartesian  # TODO: delete this
 
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -54,7 +54,6 @@ using ImageMetadata: ImageMetaAxis
 import ImageMorphology: dilate, erode
 import ImageTransformations: restrict
 using TiledIteration: EdgeIterator
-using CoordinateTransformations: Translation
 
 using Base.Cartesian  # TODO: delete this
 
--- a/test/algorithms.jl
+++ b/test/algorithms.jl
@@ -232,6 +232,30 @@ using Base.Test
         @test size.(pyramidlevel1) == [(32,32), (16,16)]
     end
 
+    @testset "gaussian_pyramid" begin
+        #Tests for OffsetArrays
+        img = zeros(70, 70)
+        img[20:51, 20:51] = 1
+        imgo = OffsetArray(img, 0, 0)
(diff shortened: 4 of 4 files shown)


Evizero:
Using an `if` like this looks to me like it would break type-inference. Could you test that by any chance?

zygmuntszpak:
Hi Evizero,

how would we go about writing a test for this? I'm still having trouble grasping how type inference works. Is the issue that when the function gets compiled with an `OffsetArray` type, the compiler will not know whether `if all(x->isa(x, Base.OneTo), indices(img))` is true or not, because it would actually have to execute that line at run-time to be sure? Without knowing if this line is true or not, it cannot determine whether the output will be a canonical array, or an offsetarray? Is this what you mean by breaking type-inference?

With the following code
```julia
using Images, Base.Test, ImageView, OffsetArrays

img = zeros(70, 70)
img[20:51, 20:51] = 1
pyramid = gaussian_pyramid(img, 3, 2, 1.0)

@code_warntype gaussian_pyramid(img, 3, 2, 1.0)

imgo = OffsetArray(img, -1, -1)
pyramido = gaussian_pyramid(imgo, 3, 2, 1.0)

@code_warntype gaussian_pyramid(imgo, 3, 2, 1.0)
```

the statement  `@code_warntype gaussian_pyramid(img, 3, 2, 1.0)` yields

```julia
Variables:
  #self# <optimized out>
  img::Array{Float64,2}
  n_scales::Int64
  downsample::Int64
  sigma::Float64
  #39 <optimized out>
  kerng::ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}
  kern::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.T
riggsSdika{Float64,3,3,9}}

Body:
  begin
      kerng::ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9} = $(Expr(:invoke, MethodInstanc
e for (::ImageFiltering.KernelFactors.#kw##IIRGaussian)(::Array{Any,1}, ::ImageFiltering.KernelFacto
rs.#IIRGaussian, ::Type{Float64}, ::Float64), :($(QuoteNode(ImageFiltering.KernelFactors.#IIRGaussia
n))), :($(Expr(:invoke, MethodInstance for vector_any(::Any, ::Vararg{Any,N} where N), :(Base.vector
_any), :(:emit_warning), true))), :(ImageFiltering.KernelFactors.IIRGaussian), Float64, :(sigma))) #
 line 796:
      SSAValue(2) = kerng::ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}
      kern::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFacto
rs.TriggsSdika{Float64,3,3,9}} = (Core.tuple)(SSAValue(2), SSAValue(2))::Tuple{ImageFiltering.Kernel
Factors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}} # line 7
97:
      return $(Expr(:invoke, MethodInstance for gaussian_pyramid(::Array{Float64,2}, ::Int64, ::Int6
4, ::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.Trig
gsSdika{Float64,3,3,9}}), :(Images.gaussian_pyramid), :(img), :(n_scales), :(downsample), :(kern)))
  end::Array{Array{Float64,2},1}
```

which seems to suggest that the return type was correctly inferred?

The second statement `@code_warntype gaussian_pyramid(imgo, 3, 2, 1.0)`

yields

```julia
Variables:
  #self# <optimized out>
  img::OffsetArrays.OffsetArray{Float64,2,Array{Float64,2}}
  n_scales::Int64
  downsample::Int64
  sigma::Float64
  #39 <optimized out>
  kerng::ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}
  kern::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}}

Body:
  begin
      kerng::ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9} = $(Expr(:invoke, MethodInstance for (::ImageFiltering.KernelFactors.#kw##IIRGaussian)(::Array{Any,1}, ::ImageFiltering.KernelFactors.#IIRGaussian, ::Type{Float64}, ::Float64), :($(QuoteNode(ImageFilt
ering.KernelFactors.#IIRGaussian))), :($(Expr(:invoke, MethodInstance for vector_any(::Any, ::Vararg{Any,N} where N), :(Base.vector_any), :(:emit_warning), true))), :(ImageFiltering.KernelFactors.IIRGaussian), Float64, :(sigma))) # line 796:
      SSAValue(2) = kerng::ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}
      kern::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}} = (Core.tuple)(SSAValue(2), SSAValue(2))::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.Trigg
sSdika{Float64,3,3,9}} # line 797:
      return $(Expr(:invoke, MethodInstance for gaussian_pyramid(::OffsetArrays.OffsetArray{Float64,2,Array{Float64,2}}, ::Int64, ::Int64, ::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}}), :(Image
s.gaussian_pyramid), :(img), :(n_scales), :(downsample), :(kern)))
  end::Array{_,1} where _
```
The last line `end::Array{_,1} where _`  is in red in the REPL, so presumably something went wrong here? 
How should one interpret this? 

I suppose one could get rid of the `if` statement by writing a second specialised `pyramid_scale` on the `OffsetArray` type and using dispatch to pick the relevant one? Would that resolve type-instability? If so, howcome?

Evizero:
> how would we go about writing a test for this? I'm still having trouble grasping how type inference works.

The main way I usually test inference is with `@inferred`. example: https://github.com/JuliaImages/ImageTransformations.jl/blob/master/test/interpolations.jl#L5 . docs: https://docs.julialang.org/en/stable/stdlib/test/#Base.Test.@inferred . Note that it technically it doesn't check type stability, only that the return type is concrete. In most cases though that is the highest priority aspect of that.

>  Is the issue that when the function gets compiled with an OffsetArray type, the compiler will not know whether if `all(x->isa(x, Base.OneTo), indices(img))` is true or not, because it would actually have to execute that line at run-time to be sure? Without knowing if this line is true or not, it cannot determine whether the output will be a canonical array, or an offsetarray? Is this what you mean by breaking type-inference?

In general yes, though it seems the compiler is getting better and better in figuring things out at compile time. So depending on how complex the condition-statement is in the `if` and if said statement only requires type information to evaluate to a `Bool`, the compiler might be able to figure out which branch of the `if` to take and thus remove the `if`. My suspicion was that in this case the expression is too complex for it to do so.

> which seems to suggest that the return type was correctly inferred?

yes. Though the actual method of interest is hidden behind the method invocation and it is not quite clear to me how the compiler figured out the return type correctly. Especially because in the second example it did not do so.

> The last line end::Array{_,1} where _ is in red in the REPL, so presumably something went wrong here?

Correct. type inference is "broken" here. I say broken because I am confident it can be avoided, not because it is inherently bad, which of course it isn't. It is a useful feature to allow for code that is not type stable.

> I suppose one could get rid of the if statement by writing a second specialised pyramid_scale on the OffsetArray type and using dispatch to pick the relevant one? Would that resolve type-instability? If so, howcome?

That is basically what I would suggest/do. Basically AFAIK there are two types of dispatch, static and dynamic. If the type of an object is known at compile time (because is is inferred correctly), then dispatch can also be done at compile time. If not, the dispatch has do be done during run time dynamically. We usually aim for static dispatch, because that allows the compiler to see which exact method of the called function it would use, and try to infer the return type of that method (and so on). If the compiler has to resort to dynamic dispatch it can know during compile time which method it would invoke, which implies it likely can't figure out the return type either.

So basically as soon as some object's type can't be inferred, it will likely poison return-type inference for function call it is involved in, and so will the objects those functions return. In performance sensitive code we want to avoid that where we can.

src/algorithms.jl
```diff
@@ -790,38 +790,57 @@ Returns a  gaussian pyramid of scales `n_scales`, each downsampled
...
     pyramid
 end
 
-function pyramid_scale(img, downsample)
+function pyramid_scale(img::AbstractArray{T,N}, downsample::Real) where {T, N}
     sz_next = map(s->ceil(Int, s/downsample), size(img))
     imresize(img, sz_next)
 end
 
+function pyramid_scale(img::AbstractArray{T,N}, downsample::Real, start::Tuple) where {T, N}
```
Evizero: why is this method for offset arrays so radically different from the normal array one? (i.e. requires `Translation` and `warp`)

src/algorithms.jl
```diff
@@ -790,38 +790,57 @@ Returns a  gaussian pyramid of scales `n_scales`, each downsampled
...
-    img_scaled = pyramid_scale(img_smoothed_main, downsample)
-    prev = convert(typeof(img_scaled), img)
-    pyramid = typeof(img_scaled)[prev]
+    if all(x->isa(x, Base.OneTo), indices(img))
+        img_scaled = pyramid_scale(img_smoothed_main, downsample)
+        prev = convert(typeof(img_scaled), img)
+        pyramid = typeof(img_scaled)[prev]
+    else
+        img_scaled = pyramid_scale(img_smoothed_main, downsample, map(i -> i.start, indices(img_smoothed_main)))
+        prev = convert.(typeof.(img_scaled)[1], img)
```
Evizero: why use broad cast here?

Evizero:
@zygmuntszpak as a side note, please don't shy away from pinging me with `@evizero` when you would like me to read something. I get quite a lot of notifications per day, because I watch too many repositories/threads (my own fault basically), which means I miss quite a lot posts depending on how busy I am. I do see pings though, as those trigger email notifications as well.

zygmuntszpak:
@Evizero Thank you much for the wonderful explanation. 

@arijitkar98 Taking into account what Evizero said, all that you need to do is add the following function to what was previously there. That is, you can discard everything you wrote in the current pull-request and just add the following piece of code:

```julia
function pyramid_scale(img::OffsetArray, downsample)
    sz_next = map(s->ceil(Int, s/downsample), length.(Compat.axes(img)))
    OffsetArray(imresize(img, sz_next), start.(Compat.axes(img).-(1,1)))
end
```

You will also need to add `using OffsetArrays` to the top of the `algorithms.jl` file. 

The following sanity check affirms that it works: 
```julia
using Images, Base.Test,  OffsetArrays, BenchmarkTools

img = zeros(70, 70)
img[20:51, 20:51] = 1
pyramid = gaussian_pyramid(img, 3, 2, 1.0)

imgo = OffsetArrays.OffsetArray(img, 0, 0)
pyramido = gaussian_pyramid(imgo, 3, 2, 1.0)
```
The indices of  `pyramid` and `pyramido` agree:
```julia
Compat.axes.(pyramid)
4-element Array{Tuple{Base.OneTo{Int64},Base.OneTo{Int64}},1}:
 (Base.OneTo(70), Base.OneTo(70))
 (Base.OneTo(35), Base.OneTo(35))
 (Base.OneTo(18), Base.OneTo(18))
 (Base.OneTo(9), Base.OneTo(9))

Compat.axes.(pyramido)
4-element Array{Tuple{UnitRange{Int64},UnitRange{Int64}},1}:
 (1:70, 1:70)
 (1:35, 1:35)
 (1:18, 1:18)
 (1:9, 1:9)
```

Also, `@code_warntype` suggests that everything is now inferred correctly.
```julia
 @code_warntype gaussian_pyramid(img, 3, 2, 1.0)
Variables:
  #self# <optimized out>
  img::Array{Float64,2}
  n_scales::Int64
  downsample::Int64
  sigma::Float64
  #39 <optimized out>
  kerng::ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}
  kern::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}}

Body:
  begin
      kerng::ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9} = $(Expr(:invoke, MethodInstance for (::ImageFiltering.KernelFactors.#kw##IIRGaussian)(::Array{Any,1}, ::ImageFiltering.KernelFactors.#IIRGaussian, ::Type{Float64}, ::Float64), :($(QuoteNode(ImageFiltering.KernelFactors.#IIRGaussian))), :($(Expr(:invoke, MethodInstance for vector_any(::Any, ::Vararg{Any,N} where N), :(Base.vector_any), :(:emit_warning), true))), :(ImageFiltering.KernelFactors.IIRGaussian), Float64, :(sigma))) # line 854:
      SSAValue(2) = kerng::ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}
      kern::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}} = (Core.tuple)(SSAValue(2), SSAValue(2))::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}} # line 855:
      return $(Expr(:invoke, MethodInstance for gaussian_pyramid(::Array{Float64,2}, ::Int64, ::Int64, ::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}}), :(Images.gaussian_pyramid), :(img), :(n_scales), :(downsample), :(kern)))
  end::Array{Array{Float64,2},1}
```

```julia
@code_warntype gaussian_pyramid(imgo, 3, 2, 1.0)
Variables:
  #self# <optimized out>
  img::OffsetArrays.OffsetArray{Float64,2,Array{Float64,2}}
  n_scales::Int64
  downsample::Int64
  sigma::Float64
  #39 <optimized out>
  kerng::ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}
  kern::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}}

Body:
  begin
      kerng::ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9} = $(Expr(:invoke, MethodInstance for (::ImageFiltering.KernelFactors.#kw##IIRGaussian)(::Array{Any,1}, ::ImageFiltering.KernelFactors.#IIRGaussian, ::Type{Float64}, ::Float64), :($(QuoteNode(ImageFiltering.KernelFactors.#IIRGaussian))), :($(Expr(:invoke, MethodInstance for vector_any(::Any, ::Vararg{Any,N} where N), :(Base.vector_any), :(:emit_warning), true))), :(ImageFiltering.KernelFactors.IIRGaussian), Float64, :(sigma))) # line 854:
      SSAValue(2) = kerng::ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}
      kern::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}} = (Core.tuple)(SSAValue(2), SSAValue(2))::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}} # line 855:
      return $(Expr(:invoke, MethodInstance for gaussian_pyramid(::OffsetArrays.OffsetArray{Float64,2,Array{Float64,2}}, ::Int64, ::Int64, ::Tuple{ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9},ImageFiltering.KernelFactors.TriggsSdika{Float64,3,3,9}}), :(Images.gaussian_pyramid), :(img), :(n_scales), :(downsample), :(kern)))
  end::Array{OffsetArrays.OffsetArray{Float64,2,Array{Float64,2}},1}
```
It would be great if you could add a few further tests for the case of the `OffsetArray`. You could recycle the tests that currently exists by wrapping  `img` with zero offset, e.g. OffsetArrays.OffsetArray(img, 0, 0)`, so that only the type has changed.

zygmuntszpak:
@Evizero 
Arijit took a closer look at my suggestion above and discovered that when one uses an image of type
`ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}}` and wraps it with an `OffsetArray` as per the example code below:

```julia
using Images, Base.Test, ImageView, OffsetArrays, TestImages
img1 = Gray.(testimage(test_image))
img1o = OffsetArray(img1, 0, 0)
gaussian_pyramid(img1o, 3, 2, 1.0)
```

then one gets a conversion error:

```julia
MethodError: Cannot `convert` an object of type OffsetArrays.OffsetArray{ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}},2,Array{ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}},2}} to an object of type OffsetArrays.OffsetArray{ColorTypes.Gray{Float64},2,Array{ColorTypes.Gray{Float64},2}}
This may have arisen from a call to the constructor OffsetArrays.OffsetArray{ColorTypes.Gray{Float64},2,Array{ColorTypes.Gray{Float64},2}}(...),
since type constructors fall back to convert methods.
in gaussian_pyramid at Images\src\algorithms.jl:855
in gaussian_pyramid at Images\src\algorithms.jl:865
```
This arises because `imfilter` will produce an output image of type `Float64` even if the original was `UInt8`. It seems like the relevant conversion and promotion rules are not defined for the cases when a Grayscale image of type `UInt8` is wrapped by an `OffsetArray`. 
 
If one started with a `Float64` grayscale image, e.g.
```julia
 gaussian_pyramid(Float64.(img1o), 3, 2, 1.0)
```
then everything works correctly. So it seems like it is a matter of somewhere defining a conversion rule which applies `Float64` to each element of the array.

Where would one define such a rule? Should it sit in ColorTypes or FixedPointNumbers?

Evizero:
Good catch! I just looked into it myself out of curiosity. Looks like the relevant methods for `Array` live here: https://github.com/JuliaImages/ImageCore.jl/blob/master/src/convert_reinterpret.jl#L85-L88 . Luckily `ImageCore` already depends on `OffsetArray`, so I would say that such addidional convert methods should live there as well

arijitkar98:
Ping @zygmuntszpak

zygmuntszpak:
@Evizero When you have a moment could you please tag a new release for us?

Evizero:
will do when I am back tomorrow. as a side note to myself and others. we should start the habit of updating https://github.com/JuliaImages/Images.jl/blob/master/NEWS.md on the PRs. I will take a look tomorrow and add the changes in a new commit

Evizero:
https://github.com/JuliaLang/METADATA.jl/pull/15121

PR #723, opened 2018-06-11
update NEWS.md and REQUIRE

Evizero:
Small updates in order to tag the latest master.

Note that we already skipped documenting a bunch of tags in NEWS.md (I am sure much of which is my oversight). We should try to pay more attention to this before merging PRs.


Evizero:
I forgot to mention:

Relevant PR is #722 from @arijitkar98 

cc: @zygmuntszpak

PR #725, opened 2018-07-10
Add gaussian_pyramid downsampling domain warning

stillyslalom:
For downsample <= 1, images are instead upscaled; Gaussian filtering serves no purpose in such instances.


oxinabox:
Why `warn` and not just `throw(DomainError("..."))`?

stillyslalom:
As implemented, the gaussian pyramid algorithm is still valid for `downsample < 1`; it just upscales instead of downscaling. I can't think of a reason why anyone would want that, but since it's valid, it shouldn't throw an error. That said, `downsample <= 0` *should* throw a `DomainError`!

Evizero:
oh sorry. yes go ahead. Thanks for the contribution!

PR #727, opened 2018-07-30
update for Julia v0.7

RalphAS:
This depends on updates in some dependencies (notably ImageTransformations and ImageFiltering, along with their dependencies) for which PRs are pending at this time.

The vast majority of locally-caused deprecations are eliminated here.

With a set of bleeding-edge dependencies I am getting ~~4~~ ~~2~~ no errors and ~6~ no broken tests out of the test suite (except 4 tests needing fixes for ImageAxes; they are suppressed here).

The broken tests ~are~ were associated with `yen_threshold` which I have ~not studied~ apparently fixed.

2 errors ~~are~~ were associated with obsolete broadcast usage related to the `sumfinite!` function in `algorithms.jl`; ~~advice~~ review would be welcome.

~~The other 2 errors are in restriction and convex hull tests; I don't understand them yet.~~ Edit: fixed

### DIFF
--- a/docs/src/core.md
+++ b/docs/src/core.md
@@ -56,7 +56,7 @@ instead using the functions `data` and `properties` to extract these.)  These
 are the only two fields in the first concrete image type, called `Image`:
 
 ```julia
-type Image{T,N,A<:AbstractArray} <: AbstractImageDirect{T,N}
+mutable struct Image{T,N,A<:AbstractArray} <: AbstractImageDirect{T,N}
     data::A
     properties::Dict{String,Any}
 end
@@ -79,7 +79,7 @@ images. More detail about this point can be found below.
 The only other concrete image type is for indexed images:
 
 ```julia
-type ImageCmap{T,N,A<:AbstractArray,C<:AbstractArray} <: AbstractImageIndexed{T,N}
+mutable struct ImageCmap{T,N,A<:AbstractArray,C<:AbstractArray} <: AbstractImageIndexed{T,N}
     data::A
     cmap::C
     properties::Dict{String,Any}
@@ -199,7 +199,7 @@ function imfilter{T,N}(A::Array{T,N}, kernel::Array{T,N}, options...)
 The first step might be to simply provide a version for `AbstractImage` types:
 
 ```julia
-function imfilter{T,N}(img::AbstractImage{T,N}, kernel::Array{T,N}, options...)
+function imfilter(img::AbstractImage{T,N}, kernel::Array{T,N}, options...) where {T,N}
     out = imfilter(data(img), kernel, options...)
     shareproperties(img, out)
 end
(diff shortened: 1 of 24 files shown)


docs/src/core.md
```diff
@@ -56,7 +56,7 @@ instead using the functions `data` and `properties` to extract these.)  These
 are the only two fields in the first concrete image type, called `Image`:
 
 ```julia
-type Image{T,N,A<:AbstractArray} <: AbstractImageDirect{T,N}
+mutable struct Image{T,N,A<:AbstractArray} <: AbstractImageDirect{T,N}
```
Evizero: lets either leave this file untouched, or delete it. The `Image` type has been gone for quite some time now, so this file must be quite old

src/Images.jl
```diff
@@ -271,7 +274,7 @@ export # types
     # phantoms
     shepp_logan
 
-_length(A::AbstractArray) = length(linearindices(A))
+_length(A::AbstractArray) = length(LinearIndices(A))
```
Evizero: I noticed this doesn't always work as `linearindices` did in `0.6`.  probably safer to do `eachindex` here

src/corner.jl
```diff
@@ -121,8 +121,9 @@ Hence, the refined sub-pixel coordinate is equal to:
 
 """
 function corner2subpixel(responses::AbstractMatrix, corner_indicator::AbstractMatrix{Bool})
-    row_range, col_range = indices(corner_indicator)
-    row, col, _ = findnz(corner_indicator)
+    row_range, col_range = axes(corner_indicator)
+    idxs = findall(!iszero, corner_indicator) # findnz
+    row, col = (getindex.(idxs,1),getindex.(idxs,2))
```
Evizero: I suspect this workaround doesn't always work.

I say this because I also replaced that deprecation somewhere and, well, it didn't work in all cases. There I had the luxury of using `SparseArrays.findnz`, but I am not sure if that works for non-sparse arrays.

docs/src/core.md
RalphAS: Ok. I'll revert. Updating the documentation should be a separate effort.

src/Images.jl
RalphAS: Ok. And it saves a few nanoseconds for OffsetArrays!

src/corner.jl
RalphAS: Since `corner_indicator` is a matrix from the signature, there's no dimensionality problem, and I don't immediately see where other trouble would arise. Your experience seems surprising. (But then I specialize in edge cases, not corner cases.)

Can we use the proposed revision until we identify a specific need for something else?

Evizero: Yes. This was not a change request, just a comment to share my anecdotal evidence

RalphAS:
Many of the dependencies have recently been updated, not all are merged and tagged.  IIUC the biggest remaining problems are ImageAxes/AxisArrays which have many deprecations under 0.7 (so don't work under 1.0). Tests require ImageMagick which needs a decision on what to merge.

cormullion:
about 40 packages depend on Images (https://juliaobserver.com/packages?dependent_id=Images) so it would help the upgradathon stats.

RalphAS:
Thanks @rsrock. I do think that merging this to master at this point will make it easier for folks working on related packages.  Did you want further reviews before merging?

Evizero:
I agree. go for it and merge. Even if there are errors, at least those will be errors that aren't already fixed in an open PR (namely this one). So at the very least it will make it a bit more obvious where the remaining issues reside.

PR #735, opened 2018-08-18
a couple of last minute fixes for Julia v0.7

RalphAS:
This should get the MIME show stuff working for IJulia and fixes a test.


Evizero:
Nice. I was just looking at the `AxisArrays` error locally

PR #736, opened 2018-08-18
update/fix showable

Evizero:
complements the changes from @RalphAS at #735


RalphAS:
Quite right, I just realized that my test run was using an alternate path.

PR #739, opened 2018-08-19
end hausdorff deprecation

Evizero:
Good as time as any to end outstanding deprecation periods.

cc: @juliohm


juliohm:
Awesome @Evizero , thank you. Feel free to merge anytime.

PR #740, opened 2018-08-19
Using Statistics.var

tknopp:
Its used later in this file and I get a deprecation warning in my code.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1,7 +1,7 @@
 using Base: axes1, tail
 using OffsetArrays
 import Statistics
-using Statistics: mean
+using Statistics: mean, var
 import AxisArrays
 
 # Compat.@dep_vectorize_2arg Gray atan2


RalphAS:
We are extending `var` and `std` below, so they are qualified in all uses AFAICT. An additional `using` shouldn't hurt, but if you're using a (fairly recent) commit from master it shouldn't help either.

tknopp:
no its not qualified in this line:
https://github.com/JuliaImages/Images.jl/blob/master/src/algorithms.jl#L80

RalphAS:
My mistake.  You are of course correct.  Maybe when the test requirements are all tagged we can do a coverage analysis to avoid this kind of embarrassment.

tknopp:
you can see here that its not covered: https://codecov.io/gh/JuliaImages/Images.jl/src/master/src/algorithms.jl#L72.
But don't worry 90% code coverage is pretty good.

PR #741, opened 2018-08-22
Fix juno integration

pfitzseb:
No need for all of that now that Juno properly integrates with Base's display system.


Evizero:
nice! Is this the only reason we needed Requires.jl ?

pfitzseb:
I think so, yes.

Issue #737, opened 2018-08-18
Proposal: move MIME show methods to another package


RalphAS:
Background:

Under a Julia interface with display graphics, a command like
```julia
using FileIO; img = load("myfile.png")
```
wants to display the result inline, via a `show` method. A particularly important example is IJulia/Jupyter, where the default `show` fallback does an expensive conversion to SVG (which takes minutes for a large image and may not even give a usable result). (cf. #619 #676)

Currently the appropriate `show` method is defined in Images; it uses `restrict` from ImageTransformations to make the display size manageable for large images.

The disadvantage of this is that packages which load image files (ImageMagick or QuartzImageIO) have needed an Images dependency to insure that the appropriate `show` method
is registered when a `load` call returns, even if the caller of `load` has not imported Images.

Developers of those packages have apparently decided that the long tail of an Images dependency is undesirable. (https://github.com/JuliaIO/ImageMagick.jl/issues/127)

Proposal:

ISTM that there are two good options: (a) move the MIME show code to ImageCore, at the expense of adding a FileIO and Requires dependency to ImageCore, or (b) register a new package which just deals with this, and suggest that ImageMagick and QuartzImageIO depend on the new package.

I have made a draft of such a new package here: [ImageShow](https://github.com/RalphAS/ImageShow.jl)

timholy:
I really like the idea of a separate package. Should we make it `@require IJulia` in Images.jl?

RalphAS:
@timholy Could you elaborate? Do you mean that Images.jl would only load the new package (which implements `show`) when/if IJulia is also loaded?  That is feasible with some hackery.

Do you know if any other IDEs (e.g. Juno) make use of the PNG streamer?

timholy:
Right. The `restrict` based methods are there basically to make sure that display of large images isn't too slow in IJulia.

Not sure, perhaps @pfitzseb can tell us?

Evizero:
I like the sketch of ImageShow. 

Concerning Images.jl. I think the save bet would be for Images to always import ImageShow. One never knows who want to use the html mime.

pfitzseb:
> Do you know if any other IDEs (e.g. Juno) make use of the PNG streamer?

Not sure, what is that? :)

Juno currently only calls `show` methods for the following mime types (and only if the plot pane has been enabled by the user), in that order:
```
  "application/juno+plotpane", # basically text/html
  "image/svg+xml"
  "image/png",
  "image/jpeg",
[...]

PR #746, opened 2018-09-02
Transfer responsibility for MIME-writing to ImageShow

RalphAS:
This implements the proposal in #737 

Note that this is a bit messy because ImageShow implements a fallback restriction scheme (downsampling) but uses the old approach (from ImageTransformations) when available.  Thus it seems prudent to run the relevant tests under both conditions. 
Edit: For now, the old show tests stay in place (using the code in ImageShow); eventually the tests might be consolidated.


Evizero:
I think we can live with only testing the "with Images" scenario in Images.jl. I think the foremost goal here in this package should be to be the least breaking to he current behaviour.

RalphAS:
Agreed; full compatibility is the design goal here.  (There is confusion due to a namespace mistake in the ImageShow test script, which I will fix shortly.)  I am revising this PR to use the pre-existing Images.jl `writemime` test script as is, so the test is representative of real use.

PR #749, opened 2018-09-12
Augments documentation for contrast limited histogram equalisation

zygmuntszpak:


timholy:
Very impressive as usual! I'm wondering, do you think the "random variable" part is likely to be clarifying or obscuring for your average reader? One can alternatively make an appeal the gut, that you want to spread out the intensity distribution so as to use your dynamic range more effectively.

zygmuntszpak:
Thanks Tim! I'm not actually sure how to characterize the archetypical  user/reader of this library since nowadays image processing is used in so many different scientific disciplines. I've tried to give a high-level description first in the spirit of using "your dynamic range effectively":

> Histogram equalisation was initially conceived to  improve the contrast in a
> single-channel grayscale image. The method transforms the
> distribution of the intensities in an image so that they are as uniform as
> possible [1]. The natural justification for uniformity
> is that the image has better contrast  if the intensity levels of an image span
> a wide range on the intensity scale. As it turns out, the necessary
> transformation is a mapping based on the cumulative histogram.

Only then do I introduce the random variable part is to explain why the cumulative histogram is the appropriate transformation. An analogous exposition of histogram equalization (in terms of random variables) appears in the classic textbook [1].  Since the discussion of the random variables  is under the "Details" subheading I figured that anyone that found it confusing would just skip the "details" and move on to the description of the parameters and the use-case example. On the other hand, people familiar with probability theory might appreciate the theoretical underpinnings and assumptions of the CLAHE method and would continue to read the details.

zygmuntszpak:
@kmsquire I'd love your feedback on this if you have a moment.

timholy:
How about this: instead of repeating all the intro stuff, how about delete that part and refer the reader to the identical description in `histeq`? Advantages:
- [DRY principles](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) suggest it's better to have things in just one place
- for non-IJulia users this will reduce the amount of scrolling that's needed to see the argument summary at the top

 Just focus on the part specific to `clahe`, then I think we can merge this.

zygmuntszpak:
Thanks Tim! I've done as you suggested.

timholy:
Thanks!

PR #752, opened 2018-10-05
Expands `histmatch` documentation

zygmuntszpak:


src/exposure.jl
```diff
@@ -454,8 +454,86 @@ _fill(val, dim::NTuple{N,Base.OneTo}) where {N} = fill(val, map(length, dim))
 hist_matched_img = histmatch(img, oimg, nbins)
 ```
 
-Returns a grayscale histogram matched image with a granularity of `nbins` number of bins. `img` is the image to be
-matched and `oimg` is the image having the desired histogram to be matched to.
+Returns a histogram matched image with a granularity of `nbins` number
+of bins. The first parameter `img` is the image to be matched, and the second
```
timholy: Maybe call these `argument`s since `parameter` typically means type-parameter, like in `Array{Float32,2}`.

timholy:
For test failures see #753

zygmuntszpak:
Thanks for the suggestion---I've changed "parameter" to "argument". 

I'd be happy to build on your work to revise the `histmatch` implementation and I will start a separate branch for that. Could you please clarify what you would like the tests with non-trivial histograms to look like? 

I've noticed something odd with `imhist` which has to do with how  [`StatsBase.histrange`](https://github.com/JuliaStats/StatsBase.jl/blob/master/src/hist.jl) constructs its bins.

Consider the following code snippet
```julia
using Images

imgA = 0:1:255
edgesA, countsA  = imhist(imgA,256,0,255)
length(countsA) # 258 bins

imgB = 0:1/255:1
edgesB, countsB  = imhist(imgB,256,0,1)
length(counts) # 203 bins
```

You don't get the number of bins you asked for when working in the unit interval (which we typically do with our images). The fact that you get 258 bins and not 256 for imgA is explained in the `imhist` documentation. In particular, 

* `count[1]` is the number satisfying `x < edges[1]`, and
* `count[end]` is the number satisfying `x >= edges[end]`.

However, I believe we ought to obtain the same number of bins when working in the unit interval. Do you also consider this a bug?

In `Images.jl` there is a remark: `using StatsBase  # TODO: eliminate this dependency`. 

Shall we try to get rid of StatsBase in `imhist`?

timholy:
Yes, agreed, having it not be scale-invariant is bad. It's not great that it doesn't just give the user the number of bins requested, though that goal is a little bit at odds with the goal of having "nice" bin boundaries. Perhaps for "nice" boundaries we want to force the user to supply `edges` directly, and always use the number of bins requested?

Also, in the brave new world where Julia can return arrays indexed starting from something other than 1, it occurs to me that we'd ideally use `count[0]` to store the number of values that fall in the range `< edges[1]`, i.e., shift the indexing of `count` down by 1. But I don't know how to transition to that without breaking people's code, unless we simultaneously change the name of the function (which perhaps we should do, as the mix of "reasonable" and Matlab-inspired names is a little strange).

With regards to `histmatch`, here's a pretty scary example (the last section is using [Rebugger](https://github.com/timholy/Rebugger.jl) so you can easily see how it chooses the bins; in brief, they are `0.0:0.2:1.2`):
```julia
julia> img1 = Gray.([0.4 0.45; 0.5 0.55])
2×2 Array{Gray{Float64},2} with eltype Gray{Float64}:
 Gray{Float64}(0.4)  Gray{Float64}(0.45)
 Gray{Float64}(0.5)  Gray{Float64}(0.55)

julia> img2 = Gray.([0.1 0.3; 0.7 0.9])
2×2 Array{Gray{Float64},2} with eltype Gray{Float64}:
 Gray{Float64}(0.1)  Gray{Float64}(0.3)
 Gray{Float64}(0.7)  Gray{Float64}(0.9)

julia> histmatch(img1, img2, 6)
2×2 Array{Gray{Float64},2} with eltype Gray{Float64}:
 Gray{Float64}(0.0)  Gray{Float64}(0.0)
 Gray{Float64}(0.0)  Gray{Float64}(0.0)

histmatch(img::AbstractArray{T,N} where N, oimg::AbstractArray, nbins::Integer) where T<:Colorant in Images at /home/tim/.julia/dev/Images/src/exposure.jl:474
  img = Gray{Float64}[Gray{Float64}(0.4) Gray{Float64}(0.45); Gray{Float64}(0.5) Gray{Float64}(0.55)]
  oimg = Gray{Float64}[Gray{Float64}(0.1) Gray{Float64}(0.3); Gray{Float64}(0.7) Gray{Float64}(0.9)]
  nbins = 6
  T = Gray{Float64}
rebug> @eval Images let (img, oimg, nbins, T) = Main.Rebugger.getstored("39106ca8-c97a-11e8-3d8a-0fe5129cb876")
       begin
           el_gray = graytype(eltype(img))
           (oedges, ohist) = imhist(oimg, nbins, zero(el_gray), oneunit(el_gray))
           @show oedges ohist
           _histmatch(img, oedges, ohist)
       end
       end
oedges = 0.0:0.2:1.2
ohist = [0, 1, 1, 0, 1, 1, 0, 0]
2×2 Array{Gray{Float64},2} with eltype Gray{Float64}:
 Gray{Float64}(0.0)  Gray{Float64}(0.0)
 Gray{Float64}(0.0)  Gray{Float64}(0.0)
```
The bin containing 0 is occupied by only one pixel in the "reference" image, and it is not the most appropriate value for *any* of the input pixels.

As far as getting rid of StatsBase, I'd be fine with that. Originally I felt we should re-use its code wherever possible, but over time I have become convinced it may not be worth it. (It's also used by [ColorVectorSpace](https://github.com/JuliaGraphics/ColorVectorSpace.jl/blob/4ee990cbe2682f471ba67ced1b4c23d1d1856cf8/src/ColorVectorSpace.jl#L13) so getting it completely out of Images will take some effort.)

PR #757, opened 2018-10-13
Adds `build_histogram` to address limitations of `imhist`

zygmuntszpak:
In contrast to `imhist` the `build_histogram` function will return the number of bins that the user requested.  

Furthermore, when working with an 8-bit image the `build_histogram` function will return the same bin boundaries whether the intensities are represented with a `FixedPointNumbers` (`N0f8`) in the interval [0,1] or by unsigned integers in the interval [0,255]. 

The histogram `counts` are returned as an `OffsetArray` and the **zero** index of  `counts` now stores the number of intensities that are less than the first bin boundary. This constitutes a breaking change from `imhist` which did not utilise an `OffsetArray`, and instead utilised the **first** index of `counts` to store the number of intensities that were less than the first bin boundary.

The choice of the name `build_histogram` is in anticipation of further refactoring of the `exposure.jl` API where I intend to introduce the function `adjust_histogram` see (#752).

### DIFF
--- a/src/exposure.jl
+++ b/src/exposure.jl
@@ -179,6 +179,169 @@ function imhist(img::AbstractArray, edges::AbstractRange)
     edges, histogram
 end
 
+
+function partition_interval(nbins::Integer, minval::Real, maxval::Real)
+    edges = (range(0,step=(maxval - minval) / nbins,length=nbins)) .+ minval
+end
+
+function partition_interval(nbins::Integer, minval::AbstractGray, maxval::AbstractGray)
+    partition_interval(nbins, minval.val, maxval.val)
+end
+
+"""
+```
+edges, count = build_histogram(img, nbins)
+edges, count = build_histogram(img, nbins, minval, maxval)
+edges, count = build_histogram(img, edges)
+```
+
+Generates a histogram for the image over nbins spread between `[minval, maxval]`.
+Color images are automatically converted to grayscale.
+
+# Output
+Returns `edges` which is a [`range`](@ref) type that specifies how the  interval
+`[minval, maxval]` is divided into bins, and an array `count` which records the
+concomitant bin frequencies. In particular, `count` has the following
+properties:
+
+* `count[0]` is the number satisfying `x < edges[1]`
+* `count[i]` is the number of values `x` that satisfy `edges[i] <= x < edges[i+1]`
+* `count[end]` is the number satisfying `x >= edges[end]`.
+* `length(count) == length(edges)+1`.
+
+# Details
+
+One can consider a histogram as a piecewise-constant model of a probability
+density function ``f`` [1]. Suppose that ``f`` has support on some interval ``I =
+[a,b]``.  Let ``m`` be an integer and ``a = a_1 < a_2 < \\ldots < a_m < a_{m+1} =
+b`` a sequence of real numbers. Construct a sequence of intervals
+
+```math
+I_1 = [a_1,a_2], I_2 = (a_2, a_3], \\ldots, I_{m} = (a_m,a_{m+1}]
+```
+
+which partition ``I`` into subsets ``I_j`` ``(j = 1, \\ldots, m)`` on which
+``f`` is constant. These subsets satisfy ``I_i \\cap I_j = \\emptyset, \\forall
+i \\neq j``, and are commonly referred to as *bins*. Together they encompass the
+entire range of data values such that ``\\sum_j |I_j | = | I |``. Each bin has
+width ``w_j = |I_j| = a_{j+1} - a_j`` and height ``h_j`` which is the constant
+probability density over the region of the bin. Integrating the constant
+probability density over the width of the bin ``w_j`` yields a probability mass
+of ``\\pi_j = h_j w_j`` for the bin.
+
+For a sample ``x_1, x_2, \\ldots, x_N``, let
+
+
+```math
+n_j = \\sum_{n = 1}^{N}\\mathbf{1}_{(I_j)}(x_n),
+\\quad \\text{where} \\quad
+\\mathbf{1}_{(I_j)}(x) =
+\\begin{cases}
+ 1 & \\text{if} x \\in I_j,\\\\
+ 0 & \\text{otherwise},
+\\end{cases},
+```
+represents the number of samples falling into the interval ``I_j``. An estimate
+for the probability mass of the ``j``th bin is given by the relative frequency
+``\\hat{\\pi} = \\frac{n_j}{N}``, and the histogram estimator of the probability
+density function is defined as
+```math
+\\begin{aligned}
+\\hat{f}_n(x)  & = \\sum_{j = 1}^{m}\\frac{n_j}{Nw_j} \\mathbf{1}_{(I_j)}(x) \\\\
+& = \\sum_{j = 1}^{m}\\frac{\\hat{\\pi}_j}{w_j} \\mathbf{1}_{(I_j)}(x) \\\\
+& = \\sum_{j = 1}^{m}\\hat{h}_j \\mathbf{1}_{(I_j)}(x).
+\\end{aligned}
+```
+
+The function ``\\hat{f}_n(x)`` is a genuine density estimator because ``\\hat{f}_n(x)  \\ge 0`` and
+```math
+\\begin{aligned}
+\\int_{-\\infty}^{\\infty}\\hat{f}_n(x) \\operatorname{d}x & = \\sum_{j=1}^{m} \\frac{n_j}{Nw_j} w_j \\\\
+& = 1.
+\\end{aligned}
+```
+
+# Options
+Various options for the parameters of this function are described in more detail
+below.
+
+## Choices for `nbins`
+You can specify the number of discrete bins for the histogram. When specifying
+the number of bins consider the maximum number of greylevels that your image
+type supports. For example, with an image of type `N0f8` there is a maximum
+of 256 possible graylevels. Hence, if you request more than 256 bins for
+that type of image you should expect to obtain zero counts for numerous bins.
+
+
+## Choices for `minval`
+You have the option to specify the lower bound of the interval over which the
+histogram will be computed.  If `minval` is not specified then the minimum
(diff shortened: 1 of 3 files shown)


src/exposure.jl
```diff
@@ -179,6 +179,159 @@ function imhist(img::AbstractArray, edges::AbstractRange)
     edges, histogram
 end
 
+
+function partition_interval(img::AbstractArray, nbins::Integer, minval::RealLike, maxval::RealLike)
```
timholy: What's the `img` argument for?

src/exposure.jl
```diff
@@ -179,6 +179,159 @@ function imhist(img::AbstractArray, edges::AbstractRange)
...
+Color images are automatically converted to grayscale.
+
+# Output
+Returns `edges` which is a [`range`](@ref) type that specifies how the  interval
+`(minval, maxval]` is divided into bins, and an array `count` which records the
+concomitant bin frequencies. In particular, `count` has the following
+properties:
+
+* `count[i]` is the number of values `x` that satisfy `edges[i] <= x < edges[i+1]`.
+* `count[0]` is the number satisfying `x < edges[1]`, and
```
timholy: One interesting case is when the user supplies `edges` as an argument, and `edges` doesn't start indexing at 1.
I'm not certain how to handle this pithily in the documentation: `first(count) is the number satisfying `x < first(edges)` doesn't quite seem to have the same clarity, though there are some good things about it.

src/exposure.jl
```diff
@@ -179,6 +179,159 @@ function imhist(img::AbstractArray, edges::AbstractRange)
...
+end
+
+"""
+```
+edges, count = build_histogram(img, nbins)
+edges, count = build_histogram(img, nbins, minval, maxval)
+edges, count = build_histogram(img, edges)
+```
+
+Generates a histogram for the image over nbins spread between `(minval, maxval]`.
```
timholy: Is it really proper to describe this as being open on the left? Your examples below stash 0 in the expected `count[1]` bin.

test/exposure.jl
```diff
@@ -39,6 +39,91 @@ eye(m,n) = Matrix{Float64}(I,m,n)
...
+        # of bins used in the histogram in powers of two. With the exception of the
+        # first bin (with index 0), all other bins should have equal counts.
+        expected_counts = [2^i for i = 0:7]
+        bins = [2^i for i = 8:-1:1]
+        for i = 1:length(bins)
+            for T in (Gray{N0f8}, Gray{N0f16}, Gray{Float32}, Gray{Float64})
+                edges, counts  = build_histogram(T.(collect(0:1/255:1)),bins[i],0,1)
+                @test length(edges) == length(counts) - 1
+                @test all(counts[1:end] .== expected_counts[i]) && counts[0] == 0
+                @test typeof(counts) <: OffsetArray
```
timholy: Rather than check the type, how about we check `axes(counts)`? That's really the fundamental property we're after here.

test/exposure.jl
```diff
@@ -39,6 +39,91 @@ eye(m,n) = Matrix{Float64}(I,m,n)
...
+                edges, counts  = build_histogram(0:1/255:1,4,128/255,192/255)
+                @test length(edges) == length(counts) - 1
+                all(counts == [0; 16; 16; 16; 208])
+                @test typeof(counts) <: OffsetArray
+            end
+
+            if T == Int
+                edges, counts  = build_histogram(0:1:255,4,120,140)
+                @test length(edges) == length(counts) - 1
+                all(counts == [0; 5; 5; 5; 241])
```
timholy: I'm surprised this passes.
```julia
julia> counts = OffsetArray([0; 5; 5; 5; 241], 0:4)
OffsetArray(::Array{Int64,1}, 0:4) with eltype Int64 with indices 0:4:
   0
   5
   5
   5
 241

julia> counts == [0; 5; 5; 5; 241]
false
```

src/exposure.jl
zygmuntszpak: At the moment the function does not accept an `OffsetArray` for `edges`, only an `AbstractRange`. If we want to support an `OffsetArray` for `edges` then things will be a little more complicated. We will have to shift the indices of `count` so that they match whatever the user has shifted their `edges` by. 

I imagine something along the lines of the description given below.

Let `m = first(edges)`  and  `o = last(edges)` denote the first and last index of the `edges` array, respectively. Then the output array `count` has the following properties:

* `count[m]` is the number satisfying `x < edges[m]`
* `count[n]` is the number of values `x` that satisfy `edges[n] <= x < edges[n+1]` (m < n < o)
* `count[o]` is the number satisfying `x >= edges[o]`
* `length(count) == length(edges)+1`.

For the typical case where  `edges` is specified by a canonical array for which the first element is indexed by one, the `count` array has the properties:

* `count[0]` is the number satisfying `x < edges[1]`
* `count[i]` is the number of values `x` that satisfy `edges[i] <= x < edges[i+1]`
* `count[end]` is the number satisfying `x >= edges[end]`
* `length(count) == length(edges)+1`.

src/exposure.jl
zygmuntszpak: Thank you, I will fix that.

test/exposure.jl
zygmuntszpak: Do you mean something like 
```julia
typeof(axes(counts)) <: Tuple{Base.Slice{UnitRange{Int64}}}
```

test/exposure.jl
zygmuntszpak: I forgot to add `@test` to that line in which case I would have seen that it fails---sorry. Actually, I just noticed that I did a few `copy/paste` errors with the `counts ==` stuff in the tests which I will fix as well.

Should we do something like `@test parent(counts) == [0; 5; 5; 5; 241]` instead?

src/exposure.jl
timholy: > At the moment the function does not accept an OffsetArray for edges, only an AbstractRange

```julia
julia> using IdentityRanges

julia> r = IdentityRange(-2:5)
IdentityRange(-2:5)

julia> isa(r, AbstractRange)
true

julia> first(axes(r, 1))
-2

julia> r = Base.Slice(-3:3)
Base.Slice(-3:3)

julia> isa(r, AbstractRange)
true

julia> first(axes(r, 1))
-3
```

Your proposed docstring works. Keep in mind that `count[end]` always works, so I don't know if you need to introduce `o`. Also `length(count) == length(edges)+1` is true in both cases. Perhaps all you have to say is "For the typical case ... the number satisfying `x < edges[1]` is stored in `count[0]`."

test/exposure.jl
timholy: No, just `axes(counts) == (0:length(edges),)`. In general I try to check values rather than types, which are typically a means to an end rather than the end in and of themselves.

Here's a way of thinking about why: the property we want to enforce is the existence of a `count[0]`, so who cares what we happen to call this type of object? If `Vector` some day starts supporting unconventional indexing---which perhaps it might, when it gets rewritten in pure Julia---do we really want this test to fail?

test/exposure.jl
timholy: Better would be `@test collect(counts) == ...`, again trying to focus behavior rather than the specific details about how objects are implemented.

src/exposure.jl
zygmuntszpak: It turns out that the function `searchsortedlast` insists on the range being indexed from 1 (https://github.com/JuliaLang/julia/blob/master/base/sort.jl). The function throws `Argument Error: Range must be indexed starting with 1`. 

If we want to facilitate the user passing `edges`  which do not start indexing from 1 then we will have to write our own `searchsortedlast` implementation. Alternatively, one would need to make a pull-request to base.

timholy: It just needs a better implementation.

My recommendation is to incorporate `has_offset_axes(edges) && throw(ArgumentError("edges must be indexed starting with 1"))` into `build_histogram` and then we can fix this later if it becomes an issue.

src/exposure.jl
```diff
@@ -179,6 +179,169 @@ function imhist(img::AbstractArray, edges::AbstractRange)
...
+using TestImages, FileIO, ImageView
+
+img =  testimage("mandril_gray");
+edges, counts  = build_histogram(img,256,0,1)
+```
+
+Given a color image, compute the histogram of the red channel.
+```julia
+img = testimage("mandrill")
+r = red(img)
```
timholy: `red.(img)`

src/exposure.jl
```diff
@@ -179,6 +179,169 @@ function imhist(img::AbstractArray, edges::AbstractRange)
...
+    edges = partition_interval(nbins, minval, maxval)
+    build_histogram(img, edges)
+end
+
+function build_histogram(img::AbstractArray, nbins::Integer=200)
+    build_histogram(img, nbins, minfinite(img), maxfinite(img))
+end
+
+function build_histogram(img::AbstractArray, edges::AbstractRange)
+    lb = first(axes(edges,1))-1
```
timholy: Oh, wait, I don't think you have to check whether `edges` has offset indices, because it looks to me like your implementation uses generic indexing. :+1:

zygmuntszpak: The implementation uses generic indexing but `searchsortedlast` currently does not support it and so the user will encounter an error thrown by `searchsortedlast`. Hence, I will add the check to determine if `edges` has offset indices into `build_histogram` as per your original suggestion.

timholy: Not necessary---I don't care whether the error is thrown by `build_histogram` or `searchsortedlast`. And if we enhance the implementation of `searchsortedlast` to use generic indexing then this starts working automatically.

The only real issue would be the docstring. Again, though, I don't think this is an urgent priority, especially now that this implementation *is* generic.

zygmuntszpak:
I didn't wait for the tests to pass because I just fixed the typo in the docstring `red.(img)`.

zygmuntszpak:
@timholy Perhaps you could tag a release, not necessarily for this commit, but for one sometime after you removed `FactCheck`. I read on Slack that some people were running `test Images` after they installed the Images package and were encountering all of the failed tests which left them very confused.

PR #761, opened 2018-10-24
Adds `build_histogram(Equalization(),...)`

zygmuntszpak:
Removed the dependence on `searchsortedlast` from `build_histogram` to improve the performance of histogram construction. Added a fresh implementation of histogram equalization which  does not require `_histeq_pixel_rescale`.

### DIFF
--- a/src/exposure.jl
+++ b/src/exposure.jl
@@ -318,28 +318,34 @@ function build_histogram(img::AbstractArray, nbins::Integer=200)
     build_histogram(img, nbins, minfinite(img), maxfinite(img))
 end
 
+function build_histogram(img::AbstractArray{T}, edges::AbstractRange) where {T<:AbstractRGB}
+    build_histogram(Gray.(img),edges)
+end
+
 function build_histogram(img::AbstractArray, edges::AbstractRange)
+    Base.has_offset_axes(edges) && throw(ArgumentError("edges must be indexed starting with 1"))
     lb = first(axes(edges,1))-1
     ub = last(axes(edges,1))
+    first_edge = first(edges)
+    step_size = edges.step.hi
     counts = fill(0, lb:ub)
-    o = Base.Order.Forward
-    G = graytype(eltype(img))
-    for v in img
-        if isnan(v)
-            continue
-        else
-            val = convert(G, v)
+    for val in img
+         if isnan(val)
+             continue
+         else
             if val >= edges[end]
                 counts[end] += 1
                 continue
+            elseif val < first(edges)
+                counts[lb] += 1
+            else
+                index = Int(Base.div(val-first_edge,step_size)) + 1
+                counts[index] += 1
             end
-            index = searchsortedlast(edges, val, o)
-            counts[index] += 1
         end
     end
     edges, counts
 end
-
 
 
 function _histeq_pixel_rescale(pixel::T, cdf, minval, maxval) where T<:NumberLike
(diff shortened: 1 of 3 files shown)


src/exposure.jl
```diff
@@ -482,6 +488,154 @@ end
...
+    edges, histogram = build_histogram(img, nbins, minval, maxval)
+    lb = first(axes(histogram,1))
+    ub = last(axes(histogram,1))
+    N = length(img)
+    cdf = cumsum(histogram[lb:ub]/N)
+    transform_density!(img, edges, histogram, cdf)
+end
+
+function transform_density!(img::AbstractArray,edges::AbstractArray, histogram::AbstractArray, cdf::AbstractArray)
+    o = Base.Order.Forward
```
zygmuntszpak: Forgot to remove this.

src/exposure.jl
```diff
@@ -482,6 +488,154 @@ end
...
+# References
+1. R. C. Gonzalez and R. E. Woods. *Digital Image Processing (3rd Edition)*.  Upper Saddle River, NJ, USA: Prentice-Hall,  2006.
+
+See also: [histmatch](@ref),[clahe](@ref), [build_histogram](@ref) and  [adjust_gamma](@ref).
+
+"""
+function adjust_histogram(operation::Equalization, img::AbstractArray{T}, nbins::Integer, minval::RealLike = 0, maxval::RealLike = 1) where {T<:AbstractRGB}
+    yiq = convert.(YIQ, img)
+    yiq_view = channelview(yiq)
+    yeq = adjust_histogram!(Equalization(),view(yiq_view,1,:,:),nbins,minval,maxval)
```
zygmuntszpak: Don't need to declare `yeq` since `adjust_histogram!` modifies its argument. I forgot to remove the `yeq`  after refactoring my initial implementation.

src/exposure.jl
```diff
@@ -318,30 +318,36 @@ function build_histogram(img::AbstractArray, nbins::Integer=200)
...
+function build_histogram(img::AbstractArray{T}, edges::AbstractRange) where {T<:AbstractRGB}
+    build_histogram(Gray.(img),edges)
+end
+
 function build_histogram(img::AbstractArray, edges::AbstractRange)
+    Base.has_offset_axes(edges) && throw(ArgumentError("edges must be indexed starting with 1"))
     lb = first(axes(edges,1))-1
     ub = last(axes(edges,1))
+    first_edge = first(edges)
+    step_size = edges.step.hi
```
timholy: This seems to assume a `TwicePrecision` element. But I can construct an AbstractRange as `0:11` or `StepRangeLen(0.0, 0.1, 11)`.

src/exposure.jl
```diff
@@ -482,6 +488,154 @@ end
...
+
+# References
+1. R. C. Gonzalez and R. E. Woods. *Digital Image Processing (3rd Edition)*.  Upper Saddle River, NJ, USA: Prentice-Hall,  2006.
+
+See also: [histmatch](@ref),[clahe](@ref), [build_histogram](@ref) and  [adjust_gamma](@ref).
+
+"""
+function adjust_histogram(operation::Equalization, img::AbstractArray{T}, nbins::Integer, minval::RealLike = 0, maxval::RealLike = 1) where {T<:AbstractRGB}
+    yiq = convert.(YIQ, img)
+    yiq_view = channelview(yiq)
```
timholy: Better would be something like `y = map(c->YIQ(c).y, img)` (untested).

src/exposure.jl
```diff
@@ -482,6 +488,154 @@ end
...
+imshow(imgeq)
+```
+
+# References
+1. R. C. Gonzalez and R. E. Woods. *Digital Image Processing (3rd Edition)*.  Upper Saddle River, NJ, USA: Prentice-Hall,  2006.
+
+See also: [histmatch](@ref),[clahe](@ref), [build_histogram](@ref) and  [adjust_gamma](@ref).
+
+"""
+function adjust_histogram(operation::Equalization, img::AbstractArray{T}, nbins::Integer, minval::RealLike = 0, maxval::RealLike = 1) where {T<:AbstractRGB}
```
timholy: Does it have to be `AbstractRGB` or does any `Color3` suffice?

src/exposure.jl
```diff
@@ -318,30 +318,36 @@ function build_histogram(img::AbstractArray, nbins::Integer=200)
     build_histogram(img, nbins, minfinite(img), maxfinite(img))
 end
 
+function build_histogram(img::AbstractArray{T}, edges::AbstractRange) where {T<:AbstractRGB}
+    build_histogram(Gray.(img),edges)
+end
+
 function build_histogram(img::AbstractArray, edges::AbstractRange)
+    Base.has_offset_axes(edges) && throw(ArgumentError("edges must be indexed starting with 1"))
```
timholy: Seems generic so not sure this is necessary (but I have no objection to including it, either).

src/exposure.jl
```diff
@@ -482,6 +488,154 @@ end
...
+end
+
+function transform_density!(img::AbstractArray,edges::AbstractArray, histogram::AbstractArray, cdf::AbstractArray)
+    o = Base.Order.Forward
+    first_edge = first(edges)
+    step_size = edges.step.hi
+    map!(img,img) do val
+        if val >= edges[end]
+            newval = cdf[end]
+        else
```
timholy: do you need to check `val < first(edges)` here too?

test/exposure.jl
```diff
@@ -77,7 +77,8 @@ eye(m,n) = Matrix{Float64}(I,m,n)
                 @test collect(counts) == [128; 16; 16; 16; 80]
                 @test axes(counts) == (0:length(edges),)
             else
-                edges, counts  = build_histogram(0:1/255:1,4,128/255,192/255)
+                img = collect(0:1/255:1)
+                edges, counts  = build_histogram(T.(img),4,128/255,192/255)
```
timholy: Just curious, what's the reason for this change?

src/exposure.jl
```diff
@@ -482,6 +488,154 @@ end
...
+struct Equalization <: AbstractHistogramOperation end
+
+"""
+```
+hist_equalised_img = adjust_histogram(Equalization(),img, nbins)
+hist_equalised_img = adjust_histogram(Equalization(),img, nbins, minval, maxval)
+```
+
+Returns a histogram equalised image with a granularity of approximately `nbins`
+number of bins.
```
timholy: Can you remind me the reason for the "approximately"?

src/exposure.jl
zygmuntszpak: It's not quite generic yet. The manner in which I calculate the index currently still assumes that edges start from 1.

src/exposure.jl
zygmuntszpak: `Color3` is what we want. I occasionally have trouble determining what the most abstract types are---sorry.

src/exposure.jl
zygmuntszpak: I actually need not only the `y` value, but also the `i` and `q` values since I need to reconstruct the original colour image. That's why I convert to `YIQ` and pass the `Y` as a view to be equalised.

src/exposure.jl
zygmuntszpak: Yes I do :D.

test/exposure.jl
zygmuntszpak: I wanted to convert the `img` to all the different types `T` that we loop over in the outer loop. That was my original intention but I had forgotten to actually convert the `img` to type `T` in the previous commit.

src/exposure.jl
zygmuntszpak: This is a copy/paste mistake. I copied the description from `imhist`. I recently added the word _approximately_ to `imhist`  because `imhist` doesn't guarantee that the number of bins that the user specifies actually are utilised in the construction of the histogram. Thanks for spotting this---I've fixed it in the latest version.

src/exposure.jl
zygmuntszpak: Thanks. I think I've handled the general case in the new version of the code.

src/exposure.jl
timholy: Good. But wouldn't it still be better to access the fields you want directly rather than using `channelview`? To me the meaning of `A[i,j].y` seems much more evident than `A[1,i,j]`.

timholy:
No problem about the new branch, I know how it goes. I will close this but note some responses above.

PR #762, opened 2018-10-26
Adds `adjust_histogram(Equalization(),...)`

zygmuntszpak:
Removed the dependence on `searchsortedlast` from `build_histogram` to improve the performance of histogram construction. Added a fresh implementation of histogram equalization which  does not require `_histeq_pixel_rescale`.

### DIFF
--- a/src/exposure.jl
+++ b/src/exposure.jl
@@ -318,29 +318,34 @@ function build_histogram(img::AbstractArray, nbins::Integer=200)
     build_histogram(img, nbins, minfinite(img), maxfinite(img))
 end
 
+function build_histogram(img::AbstractArray{T}, edges::AbstractRange) where {T<:AbstractRGB}
+    build_histogram(Gray.(img),edges)
+end
+
 function build_histogram(img::AbstractArray, edges::AbstractRange)
+    Base.has_offset_axes(edges) && throw(ArgumentError("edges must be indexed starting with 1"))
     lb = first(axes(edges,1))-1
     ub = last(axes(edges,1))
+    first_edge = first(edges)
+    step_size = step(edges)
     counts = fill(0, lb:ub)
-    o = Base.Order.Forward
-    G = graytype(eltype(img))
-    for v in img
-        if isnan(v)
-            continue
-        else
-            val = convert(G, v)
+    for val in img
+         if isnan(val)
+             continue
+         else
             if val >= edges[end]
                 counts[end] += 1
                 continue
+            elseif val < first(edges)
+                counts[lb] += 1
+            else
+                index = Int(Base.div(val-first_edge,step_size)) + 1
+                counts[index] += 1
             end
-            index = searchsortedlast(edges, val, o)
-            counts[index] += 1
         end
     end
     edges, counts
 end
-
-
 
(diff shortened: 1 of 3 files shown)


src/exposure.jl
```diff
@@ -482,6 +499,161 @@ end
...
+
+# References
+1. R. C. Gonzalez and R. E. Woods. *Digital Image Processing (3rd Edition)*.  Upper Saddle River, NJ, USA: Prentice-Hall,  2006.
+
+See also: [histmatch](@ref),[clahe](@ref), [build_histogram](@ref) and  [adjust_gamma](@ref).
+
+"""
+function adjust_histogram(operation::Equalization, img::AbstractArray{T}, nbins::Integer, minval::RealLike = 0, maxval::RealLike = 1) where {T<:Color3}
+    yiq = convert.(YIQ, img)
+    yiq_view = channelview(yiq)
```
timholy: Oh wait, I see now: you're doing this in-place, so by modifying the view you can then reconstruct the original. Clever. No objections to the use of `channelview` anymore, sorry I didn't get this the first time I looked at it.

src/exposure.jl
```diff
@@ -482,6 +499,161 @@ end
...
+abstract type AbstractHistogramOperation end
+struct Equalization <: AbstractHistogramOperation end
+
+"""
+```
+hist_equalised_img = adjust_histogram(Equalization(),img, nbins)
+hist_equalised_img = adjust_histogram(Equalization(),img, nbins, minval, maxval)
+```
+
+Returns a histogram equalised image with a granularity of approximately `nbins`
```
timholy: Still says "approximately"

src/exposure.jl
```diff
@@ -318,29 +318,46 @@ function build_histogram(img::AbstractArray, nbins::Integer=200)
...
     edges, counts
 end
 
+function determine_stepsize(edges::UnitRange)
+    1
+end
 
+function determine_stepsize(edges::AbstractRange)
+    if typeof(edges.step) <: Base.TwicePrecision
+        edges.step.hi
```
timholy: Looks much better, but out of curiosity why are you not just calling `step(edges)`? That will actually be more accurate, it will return `step.hi + step.lo` which in some cases can differ (typically in the least-significant bits).

zygmuntszpak: I didn't know about `step(edges)`. I was scratching my head wondering why there isn't a `step` field for every `AbstractRange`. Thanks for drawing my attention to `step`!

src/exposure.jl
zygmuntszpak: *facepalm*

zygmuntszpak:
Thanks Tim! I will add a few more tests over the weekend before I merge.

src/exposure.jl
timholy: AbstractRange "core" interface: `first`, `step`, `last`, `length`. Also useful are `minimum` and `maximum` if you don't want to distinguish between ranges with increasing vs decreasing step. Also supports `intersect` and other fancier operations you can build from these. `base/ranges.jl` makes for fairly easy reading---the only oddity is that it's so fundamental that it's used very early in Julia's bootstrap self-building process, so certain things like `@inline` are not yet available (so there are a few hoops here and there you have to jump through).

PR #764, opened 2018-10-30
Adds adjust_histogram(Matching(),...)

zygmuntszpak:
Provides a fresh implementation of histogram matching which addresses some failure cases of the `histmatch` implementation.


src/exposure.jl
```diff
@@ -641,6 +642,150 @@ function transform_density!(img::AbstractArray,edges::AbstractArray, histogram::
...
+```math
+Q(x) =  T^{-1}\\left( S(x) \\right),
+```
+
+where ``T^{-1}(y) = \\operatorname{min} \\{ x \\in \\mathbb{R} : y \\leq T(x)
+\\}`` is the inverse cumulative distribution function of ``T(x)``.
+
+The mapping suggests that one can conceptualise histogram matching as performing
+histogram equalisation on the source and target image and relating the two
+equalised histograms. Refer to [adjust_histogram](@ref) for more details on
```
zygmuntszpak: Is there any way to link to a particular function based on the type of its first parameter? I would like to point the reader to `adjust_histgoram(Equalization(),...)` but `[adjust_histogram](@ref)` will just link to the list of all possible `adjust_histogram` functions.

zygmuntszpak:
I'm not sure if I am interpreting the `CodeCov` report correctly, but I think it is erroneous in this instance.

rdeits:
The code coverage issue is almost certainly due to: https://github.com/JuliaCI/Coverage.jl/issues/187

src/exposure.jl
timholy: I don't know. Anyone else?

src/exposure.jl
```diff
@@ -641,6 +642,150 @@ function transform_density!(img::AbstractArray,edges::AbstractArray, histogram::
 end
 
 
+"""
+```
+hist_matched_img = adjust_histogram(Matching(),img, oimg, nbins)
```
timholy: Would `targetimg` or `referenceimg` be clearer than `oimg`?

src/exposure.jl
```diff
@@ -641,6 +642,150 @@ function transform_density!(img::AbstractArray,edges::AbstractArray, histogram::
...
+
+# Options
+
+Various options for the parameters of this function are described in more detail
+below.
+
+## Choices for `img` and `oimg`
+
+The `adjust_histogram(Matching(),...)` function can handle a variety of input
+types. The returned image depends on the input type. If the input is an `Image`
```
timholy: What's this reference to `Image`?

src/exposure.jl
```diff
@@ -641,6 +642,150 @@ function transform_density!(img::AbstractArray,edges::AbstractArray, histogram::
...
+then the resulting image is of the same type and has the same properties.
+
+For coloured images, the inputs are converted to
```
timholy: I don't feel strongly about this, but perhaps use American spelling here? (since `ColorTypes` etc. use it.) But if you'd like to champion some diversity here, I won't complain.

src/exposure.jl
```diff
@@ -641,6 +642,150 @@ function transform_density!(img::AbstractArray,edges::AbstractArray, histogram::
...
+    end
+    # Construct empirical cummulative density function for both images.
+    edges , histogram = build_histogram(img, nbins, imin, imax)
+    oedges, ohistogram = build_histogram(oimg, nbins, imin, imax)
+    step_size = step(edges)
+    first_edge = first(oedges)
+    last_edge = last(oedges)
+    lb = first(axes(histogram,1))
+    ub = last(axes(histogram,1))
+    cdf = cumsum(histogram[lb:ub]/length(img))
```
timholy: I'm not sure I understand why the indexing is needed, and this isn't just `histogram`. If it's to ensure 1-based indexing, where are you getting an error? `lookup_icdf` seems pretty easy to write with general indexing.

test/exposure.jl
```diff
@@ -467,6 +467,57 @@ eye(m,n) = Matrix{Float64}(I,m,n)
...
+            imgo = T.(adjust_gamma(img,0.5))
+
+            result = adjust_histogram(Matching(),img,imgo,256)
+
+            edges_target, counts_target = build_histogram(imgo,256,0,1)
+            edges_result, counts_result = build_histogram(result,256,0,1)
+            @test all(counts_target .== counts_result)
+        end
+
+        for T in (Gray{N0f8}, Gray{N0f16}, Gray{Float16}, Gray{Float32})
```
timholy: Is `T` used?

test/exposure.jl
```diff
@@ -467,6 +467,57 @@ eye(m,n) = Matrix{Float64}(I,m,n)
...
+            img = Gray{N0f8}.([i < 64 ? i/255 : 0 for i = 0:255])
+            imgo = Gray{N0f8}.([i > 64 && i < 128 ? i/255 : 0 for i = 0:255])
+            result = adjust_histogram(Matching(),img,imgo,256)
+            edges_source, counts_source = build_histogram(img,256,0,1)
+            edges_target, counts_target = build_histogram(imgo,256,0,1)
+            edges_result, counts_result = build_histogram(result,256,0,1)
+            @test all(counts_target .== counts_result)
+        end
+
+        for T in (RGB{N0f8}, RGB{N0f16}, RGB{Float32}, RGB{Float64})
```
timholy: Likewise here

src/exposure.jl
```diff
@@ -641,6 +642,150 @@ function transform_density!(img::AbstractArray,edges::AbstractArray, histogram::
...
+
+# References
+1. W. Burger and M. J. Burge. *Digital Image Processing*. Texts in Computer Science, 2016. [doi:10.1007/978-1-4471-6684-9](https://doi.org/10.1007/978-1-4471-6684-9)
+
+
+See also: [adjust_histogram](@ref), [adjust_histogram!](@ref) and [clahe](@ref).
+
+"""
+function adjust_histogram(operation::Matching, img::AbstractArray, oimg::AbstractArray, nbins::Integer = 256)
+    adjust_histogram!(Matching(), copy(img), copy(oimg), nbins)
```
timholy: Is the `copy(oimg)` necessary if it's only acting as a reference?

src/exposure.jl
```diff
@@ -641,6 +642,150 @@ function transform_density!(img::AbstractArray,edges::AbstractArray, histogram::
...
+    # to the distribution of oimg.
+    map!(img,img) do val
+        if val >= last_edge
+            newval = oedges[last(lookup_table)-1] + step(oedges)
+        elseif val < first_edge
+            newval = oedges[first(lookup_table)]
+        else
+            index = Int(Base.div(val-first_edge,step_size)) + 1
+            newval = oedges[lookup_table[index]]
+        end
```
timholy: For clarity consider adding `return newval`

src/exposure.jl
```diff
@@ -641,6 +642,150 @@ function transform_density!(img::AbstractArray,edges::AbstractArray, histogram::
...
+
+function adjust_histogram!(operation::Matching, img::AbstractArray, oimg::AbstractArray, nbins::Integer = 256 )
+    if eltype(img) <: AbstractGray
+        imin, imax = 0, 1
+    else
+        imin, imax = min(minfinite(img), minfinite(oimg)), max(maxfinite(img), maxfinite(oimg))
+    end
+    # Construct empirical cummulative density function for both images.
+    edges , histogram = build_histogram(img, nbins, imin, imax)
+    oedges, ohistogram = build_histogram(oimg, nbins, imin, imax)
```
timholy: Would it make sense instead to use `_, ohistogram = build_histogram(oimg, edges)` here? It would more obviously enforce alignment of the histograms. (I'm not saying they aren't aligned now, just that for anyone reading the code, it will be more obvious.)

src/exposure.jl
```diff
@@ -641,6 +642,150 @@ function transform_density!(img::AbstractArray,edges::AbstractArray, histogram::
...
+end
+
+function adjust_histogram!(operation::Matching, img::AbstractArray{T}, oimg::AbstractArray{T}, nbins::Integer = 256 ) where T <: Color3
+    yiq = convert.(YIQ, img)
+    yiq_view = channelview(yiq)
+    adjust_histogram!(Matching(),view(yiq_view,1,:,:), map(c->YIQ(c).y, oimg), nbins)
+    convert.(T, yiq)
+end
+
+function adjust_histogram!(operation::Matching, img::AbstractArray, oimg::AbstractArray, nbins::Integer = 256 )
```
timholy: Might it be worth having a version where we pass `edges` as an argument?

src/exposure.jl
```diff
@@ -641,6 +642,150 @@ function transform_density!(img::AbstractArray,edges::AbstractArray, histogram::
...
+    oedges, ohistogram = build_histogram(oimg, nbins, imin, imax)
+    step_size = step(edges)
+    first_edge = first(oedges)
+    last_edge = last(oedges)
+    lb = first(axes(histogram,1))
+    ub = last(axes(histogram,1))
+    cdf = cumsum(histogram[lb:ub]/length(img))
+    ocdf = cumsum(ohistogram[lb:ub]/length(oimg))
+    # Precompute the inverse cummulative distribution function of ocdf.
+    lookup_table = lookup_icdf(cdf::AbstractArray, ocdf::AbstractArray)
```
timholy: Type annotations leftovers from copy/paste

src/exposure.jl
```diff
@@ -641,6 +642,150 @@ function transform_density!(img::AbstractArray,edges::AbstractArray, histogram::
...
+    # Transform the intensities in img so that they are distributed according
+    # to the distribution of oimg.
+    map!(img,img) do val
+        if val >= last_edge
+            newval = oedges[last(lookup_table)-1] + step(oedges)
+        elseif val < first_edge
+            newval = oedges[first(lookup_table)]
+        else
+            index = Int(Base.div(val-first_edge,step_size)) + 1
```
timholy: What will this do for `val=NaN`?

test/exposure.jl
zygmuntszpak: Fixed it.

test/exposure.jl
zygmuntszpak: Fixed it.

zygmuntszpak:
I've added additional tests to cover `NaN` and to cover the case of passing `edges` directly. Also refactored the code somewhat.

src/exposure.jl
Evizero: - Does this work? I think the function name needs to be code escaped to even link to the docstring
  (i.e. ``[`adjust_histogram`](@ref)``)
- I am just guessing, but did you try 
  ```markdown
   [`adjust_histogram`](@ref adjust_histogram(::Matching, ::AbstractArray, ::AbstractArray, ::AbstractRange)
  ```
  (or whatever the desired signature is)

zygmuntszpak: In the Atom browser it works without escaping the code, but I will revise the code to escape the function name as per your suggestion. I just didn't notice the lack of escaping because it was working for me in the Atom documentation browser. 

Unfortunately adding `@ref adjust_histogram(::Matching, ::AbstractArray, ::AbstractArray, ::AbstractRange)` doesn't work. It doesn't link to anything then.

zygmuntszpak: Actually, Christof it turns out that your suggestion *does* work when you build the documentation using `Documenter.jl` provided you add the full function signature to `function_reference.md`. I  managed to figure it out thanks to the advice of [fredrikekre in a discourse post](https://discourse.julialang.org/t/add-cross-references-foo-ref-to-functions-with-same-name-but-different-arguments/17113). 

Unfortunately, it doesn't work inside the `Juno` (Atom) editor. I might open an issue there. 

I'll work on fixing up all the docstrings of the `build_histogram` and `adjust_histogram` functions before I merge this.

PR #777, opened 2019-01-07
Add Boundary filling algorithm

aquatiko:
**Closes:** https://github.com/JuliaImages/Images.jl/issues/775
This is the initial version of the algorithm and I have only added support for 2D type images.
If this seems good, I will add a generalized method including 3D images.

**Input Image:**
yw7sd
**Result:** 
screenshot from 2019-01-07 17-58-31

### DIFF
--- a/src/connected.jl
+++ b/src/connected.jl
@@ -281,3 +281,37 @@ function component_centroids(img::AbstractArray{Int,N}) where N
     end
     map(v -> n[v][1].I ./ n[v][2], 1:len)
 end
+
+"""
+ ```
+ filled_img = fill_boundaries(img)
+ filled_img = fill_boundaries(img, connectivity)
+ ```
+ Return bool array where boundaries between labeled regions are True.
+ Parameters:
+  -  img            = Input image (Boolean array type)
+  -  connectivity   = connectivity takes the same values as in label_components (Default value is 1:ndims(img))
+ """
+function fill_boundaries(img::AbstractArray{Bool},connectivity::Union{Dims, AbstractVector{Int}, BitArray}=1:ndims(img))
+    
+    labels = label_components(img,connectivity)
+    boundaries = zeros(Bool,(size(img)))
+    
+    if ndims(img) == 2
+        M = size(img)[1]
+        N = size(img)[2]
+        for ind in eachindex(img)
+            if ind%M == 1 || ind%M == 0 || ind <= M || ind+M >= M*N     # mark end boundaries
+                 boundaries[ind] = true
+            
+            elseif labels[ind] == labels[ind+1] == labels[ind-1] == labels[ind+M] == labels[ind-M] 
+                continue                                                # ignoring non Boundary pixels
+            
+            else
+                boundaries[ind] = true                              
+            end
+        end
+    end
+    
+    return boundaries
+end
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -186,6 +186,7 @@ export
     component_indices,
     component_subscripts,
     component_centroids,
+    fill_boundaries,
     magnitude,
     magnitude_phase,
     meanfinite,


src/connected.jl
```diff
@@ -281,3 +281,37 @@ function component_centroids(img::AbstractArray{Int,N}) where N
...
+    boundaries = zeros(Bool,(size(img)))
+    
+    if ndims(img) == 2
+        M = size(img)[1]
+        N = size(img)[2]
+        for ind in eachindex(img)
+            if ind%M == 1 || ind%M == 0 || ind <= M || ind+M >= M*N     # mark end boundaries
+                 boundaries[ind] = true
+            
+            elseif labels[ind] == labels[ind+1] == labels[ind-1] == labels[ind+M] == labels[ind-M]
```
tejus-gupta: This condition will depend on the `connectivity`.

```
julia> img = zeros(Bool, (3,3))
julia> img[2, 1:3] .= true
julia> fill_boundaries(img, [2])
3×3 Array{Bool,2}:
 true  true  true
 true  true  true
 true  true  true
```

The `(2,2)` pixel isn't at the boundary according to the specified connectivity pattern.

aquatiko: Thanks for the suggestion @tejus-gupta. After changing the line as
```
elseif (1 in connectivity && labels[ind] == labels[ind+1] == labels[ind-1]) || (2 in connectivity && labels[ind] == labels[ind+M] == labels[ind-M]) 
       boundaries[ind] =false                                                # ignoring non Boundary pixels
```
it seems to be working as you said.

tejus-gupta: Check out `CartesianIndices` https://julialang.org/blog/2016/02/iteration. It makes writing multidimensional algorithm very simple.

Here is the modified `find_boundaries` function using CartesianIndices.
```
# connectivity is a list indicating whether neighbors in each direction is used.
function fill_boundaries2(img::AbstractArray{Bool}, connectivity::AbstractVector{Int} = trues(ndims(img)))

    win = CartesianIndex(Tuple(connectivity))
    boundaries = zeros(Bool,(size(img)))
    R = CartesianIndices(img)
    Ifirst, Ilast = first(R), last(R)

    for I in R
        val = img[I]
        for J in CartesianIndices(map((i,j)->i:j, Tuple(max(Ifirst, I-win)), Tuple(min(Ilast, I+win))))
            if img[J] != val
                boundaries[I] = true 
                break
            end
        end
    end

    boundaries
end
```

This version is also faster since we don't label components first.

```
julia> using BenchmarkTools

julia> img = rand(Bool, (1000, 1000));

julia> @btime fill_boundaries(img, [2]);
  33.989 ms (72 allocations: 13.50 MiB)

julia> @btime fill_boundaries2(img, [0,1]);
  20.593 ms (6 allocations: 976.95 KiB)
```

aquatiko: @tejus-gupta I have updated the method and checked it's validity with different cases. Should the tests for the same be added to a  new file as there is no separate file foe testing `connected.jl` ?

tlnagy:
ping @timholy, what's the status of this? I've been doing `Images.dilate(mask) .- mask` to accomplish this which is pretty wasteful. Would be great to have this merged.

src/connected.jl
```diff
@@ -281,3 +281,33 @@ function component_centroids(img::AbstractArray{Int,N}) where N
...
+function fill_boundaries(img::AbstractArray{Bool}, connectivity::AbstractVector{Int} = ones(Int, ndims(img)))
+
+    win = CartesianIndex(Tuple(connectivity))
+    boundaries = zeros(Bool,(size(img)))
+    R = CartesianIndices(img)
+    Ifirst, Ilast = first(R), last(R)
+
+    for I in R
+        val = img[I]
+        for J in CartesianIndices(map((i,j)->i:j, Tuple(max(Ifirst, I-win)), Tuple(min(Ilast, I+win))))
```
johnnychen94: I think the following two are equivalent here. (except that the second isn't supported until julia 1.1 IIRC)
```diff
- for J in CartesianIndices(map((i,j)->i:j, Tuple(max(Ifirst, I-win)), Tuple(min(Ilast, I+win))))
+ for J in max(Ifirst, I-win):min(Ilast, I+win)
```

and we can add a line of code to make this compatible:
```julia
_colon(I, J) = I:J
if v"1.0" <= VERSION < v"1.1"
    _colon(I::CartesianIndex{N}, J::CartesianIndex{N}) where N =
        CartesianIndices(map((i,j) -> i:j, Tuple(I), Tuple(J)))
end
```

src/connected.jl
```diff
@@ -281,3 +281,33 @@ function component_centroids(img::AbstractArray{Int,N}) where N
...
+ ```
+ Return bool array where boundaries between labeled regions are True.
+ Parameters:
+  -  img            = Input image (Boolean array type)
+  -  connectivity   = connectivity takes the same values as in label_components (Default value is 1:ndims(img))
+ """
+function fill_boundaries(img::AbstractArray{Bool}, connectivity::AbstractVector{Int} = ones(Int, ndims(img)))
+
+    win = CartesianIndex(Tuple(connectivity))
+    boundaries = zeros(Bool,(size(img)))
```
johnnychen94: I prefer this to be `zeros(T, size(img))`, and loose type annotation `img::AbstractArray{Bool}` to `img::AbstractArray{T}` where `T<:Union{Bool, Gray{Bool}`. By doing this we support both `Array{Bool}` and `Array{Gray{Bool}}`.

src/connected.jl
```diff
@@ -281,3 +281,33 @@ function component_centroids(img::AbstractArray{Int,N}) where N
...
+
+"""
+ ```
+ filled_img = fill_boundaries(img)
+ filled_img = fill_boundaries(img, connectivity)
+ ```
+ Return bool array where boundaries between labeled regions are True.
+ Parameters:
+  -  img            = Input image (Boolean array type)
+  -  connectivity   = connectivity takes the same values as in label_components (Default value is 1:ndims(img))
```
johnnychen94: The default value of `connectivity` isn't `1:ndims(img)`. So please clarify which is used here, also you might need to describe how `connectivity` changes the results.

PR #779, opened 2019-02-01
Fix `restrict` on an AxisArray with non-1 axes

timholy:
This may seem like a bit of a rareified combination, but if you're running ImageFiltering 0.5.3+ and OffsetArrays 0.10.0+, you can trigger it on Images master with

```julia
using Images, TestImages
img = centered(testimage("mri"))
restrict(img)
```

I'm a bit unhappy about the implementation of `ofaxes` (it basically only supports Base and OffsetArrays), but I don't have a better idea ATM.


timholy:
If someone has better ideas we can always improve this later, so I'll merge this.

PR #801, opened 2019-04-22
revert CI version to keep julia 0.7 compatible

johnnychen94:
According to discussion in https://github.com/JuliaImages/ImageCore.jl/pull/71#discussion_r277172826, when we drop the CI test on 0.7 we should drop the compatibility as well.

Since there's already a register PR: https://github.com/JuliaRegistries/General/pull/162, we can keep this version `0.18.0` still compatible with julia `v0.7`. And drop `v0.7` compatibility in the next minor version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ os:
     - linux
     - osx
 julia:
+    - 0.7
     - 1.0
-    - 1.1
     - nightly
 matrix:
   allow_failures:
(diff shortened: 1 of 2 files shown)


timholy:
Any particular reason not to run on 1.1?

johnnychen94:
I meant to make CI cover only one non-LTS version.

timholy:
Since we're passing on nightly, there doesn't seem to be much to worry about.

johnnychen94:
It's also a simple revert of https://github.com/JuliaImages/Images.jl/pull/800/files

PR #809, opened 2019-05-17
Port graph-related codes to ImageMorphology

### LINKED ISSUE #807: Port convexhull.jl and connected.jl to ImageMorphology.jl
I propose to move the following two files to `ImageMorphology.jl`

* [`connected.jl`](https://github.com/JuliaImages/Images.jl/blob/master/src/connected.jl)
* [`convexhull.jl`](https://github.com/JuliaImages/Images.jl/blob/master/src/convexhull.jl)

My reason is: graphs are heavily used in morphology operations

ported functions:

* [`label_components`](https://github.com/JuliaImages/Images.jl/blob/master/src/connected.jl#L3-L29) and `label_components!` (This blocks PR: [add `imfill` function](https://github.com/JuliaImages/ImageMorphology.jl/pull/9))
* [`convexhull`](https://github.com/JuliaImages/Images.jl/blob/master/src/convexhull.jl)
* [`component_boxes`](https://github.com/JuliaImages/Images.jl/blob/bdfd044420fa6ffcd34760f804f0c3ce12186945/src/connected.jl#L228-L244), `component_lengths`, `component_indices`, `component_subscripts`, `component_centroids`

All of these functions are exported by `Images`, but it would quite painless since `ImageMorphology` is reexported by `Images`

Possibly, [`labeledarrays.jl`](https://github.com/JuliaImages/Images.jl/blob/master/src/labeledarrays.jl) can be ported to `ImageMorphology.jl` as well.

cc: @juliohm, @aquatiko

johnnychen94:
It's a trivial copy-and-paste operation to ImageMorphology.

More details: https://github.com/JuliaImages/Images.jl/issues/807

Blocking PR:

- [x] https://github.com/JuliaImages/ImageMorphology.jl/pull/17
- [x] #840 

closes #807


johnnychen94:
The reason to CI failure is: `ImageMorphology 0.2` is not compatible to Julia 0.7.

IMO the best time for this PR being merged is `Images v0.19.0`, i.e., the time we plan to drop julia 0.7 compatibility.

`using Images` or `using ImageMorphology` alone will be okay, but using both would introduce 
```
WARNING: using ImageMorphology.label_components in module Main conflicts with an existing identifier.
```

timholy:
Happy to merge this as long as the appropriate minimum version of ImageMorphology gets added to the `[compat]` section.

timholy:
Doesn't need all the patch versions added, see the docs: https://julialang.github.io/Pkg.jl/v1/compatibility/

I was more worried about the *lower* bound. Is 0.2.2 sufficient?

johnnychen94:
> Is 0.2.2 sufficient?

Yes. Thanks for the information, I thought `v0.2.3` and `v0.2.4` are incompatible if only `v0.2.2` is specified.

PR #821, opened 2019-09-21
Improve eltype generality in imstretch test

timholy:
We seem to be failing tests, this should fix it.


timholy:
The problem is ImageTransformations 0.8 doesn't support 0.7, and that upper-bounds ColorVectorSpace, and the older version of ColorVectorSpace didn't give consistent output for `float(::Gray{N0f8})`.

The easiest thing to do is drop 0.7 support. OK?

PR #839, opened 2019-11-28
remove ssdn, sadn, ncc, difftype and sumdiff

johnnychen94:
* `sad`, `ssd` ported in https://github.com/JuliaImages/ImageDistances.jl/pull/30
* `ssdn`, `sadn` and `ncc` ported in https://github.com/JuliaImages/ImageDistances.jl/pull/35
* `difftype` and `sumdiff` are no longer used and not exported.


timholy:
Seems like it should add a minimum version of ImageDistances 0.2.4

johnnychen94:
> Seems like it should add a minimum version of ImageDistances 0.2.4

May I read it as you’re suggesting the following versioning routine?

1. sub-package tag a new release that introduces new feature
2. Images.jl register the new sub-package version in [compat] section, which currently is handled by CompatHelper
3. Images.jl tag a new version when necessary

timholy:
I think that's right, but really my point was the following: you're removing functionality from Images. It's covered in ImageDistances, which gets installed when you install Images, so that should be OK. But it's OK only if the user is running an appropriate version of ImageDistances (at least version 0.2.4), and you can't be sure of that unless you declare it to be a requirement for getting this change in Images itself.

For example, the user might have `dev`ed ImageDistances back at version 0.2.3 and is still running that version. In such cases, you want this user to still be running the older version of Images, despite the availability of the newer version, otherwise you'll break his/her packages.

PR #840, opened 2019-11-28
add upper bounds to dependencies and drop 0.7 compatibility

johnnychen94:
closes #827, #828, #829, #830, #831, #832, #833, #834, #835, #836, #837, #838

I guess nobody is using Julia 0.7 now, so let's just drop the Julia v0.7 compatibility.

P.S., some versions I listed in compat section are 0.7 compatible, but I think that's okay

This PR also enables #809


timholy:
Let's retest now that #821 is merged.

Issue #825, opened 2019-11-22
Sort out our dependencies and moving to version 1.0


timholy:
For automatic registration we need to add upper bounds to all dependencies across JuliaImages. As some of you know from experience, without preparation this can cause a bit of a train wreck when it comes to Pkg's resolver. I just went through a whole bunch of the packages in JuliaImages and added [CompatHelper](https://github.com/bcbi/CompatHelper.jl) actions, which currently is our best mechanism to determine when versions need to be bumped. For those of you who maintain packages, I may not have done this with yours, I only did this for ones where I feel like I have a reasonable "ownership" stake. On https://github.com/JuliaImages the packages are sorted by recency of modification, so you can easily see whether you might want to do the same for one of your packages.

Kristoffer Carlsson may have posted a useful script [here](https://discourse.julialang.org/t/please-be-mindful-of-version-bounds-and-semantic-versioning-when-tagging-your-packages/30708/95) but discourse seems to be down ATM so I can't easily check.

timholy:
Also, since changing bounds requires a bump in minor version number, I think it makes sense to make the transition throughout the whole JuliaImages stack to version 1.0:
- [semver](https://semver.org/#how-do-i-know-when-to-release-100) says we should have done that a long time ago
- [discourse](https://discourse.julialang.org/t/how-can-we-encourage-julia-package-developers-to-release-version-1-0-0/29124) says we should
- 1.0 could be psychologically important for grant proposals and papers

This does not mean that we can't make breaking changes in the future; package versions will just go to 2.x or higher.

Now is a good time, however, to propose any breaking changes. I'll post an issue in some packages, feel free to add new issues or add to those.

johnnychen94:
For packages that aren't well built at present(e.g., `ImageDraw.jl`, `ImageMorphorlogy.jl`), will it be too early to set up a 1.0 goal? I'm afraid that I'm currently too busy with my daily work to catch up with this schedule.

Another related question is: assume that `Images.jl` had bumped to `1.0`, what should be done if we intend to bump a release of `ImageTransformation.jl`. According to SemVer, looks like every minor version bump in `ImageTransformation.jl` should also bump a minor version in `Images.jl`.

timholy:
> According to SemVer, looks like every minor version bump in ImageTransformation.jl should also bump a minor version in Images.jl.

[...]

PR #843, opened 2019-12-03
make docs a submodule and add documentation stage in CI

johnnychen94:
I'm not very sure whether it works as expected, but here's what I expect:

* enable documentation build on `Images.jl` side,
* and furthermore, versioning documentation

If this strategy works, we might also need to adjust the structure of https://github.com/JuliaImages/juliaimages.github.io so that the general folder structure looks normal, it's `docs/docs/` at present.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "docs"]
+	path = docs
+	url = git@github.com:JuliaImages/juliaimages.github.io.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "docs"]
 	path = docs
-	url = git@github.com:JuliaImages/juliaimages.github.io.git
+	url = https://github.com/JuliaImages/juliaimages.github.io.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,12 +17,12 @@ after_success:
     - julia -e 'cd(Pkg.dir("Images")); Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder())'
 
 jobs:
-    include:
-        - stage: "Documentation"
-        julia: 1.0
-        os: linux
-        script:
-            - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
-                                                    Pkg.instantiate()'
-            - julia --project=docs/ docs/docs/make.jl
-        after_success: skip
+  include:
+    - stage: "Documentation"
+      julia: 1.0
+      os: linux
+      script:
+      - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
+                                             Pkg.instantiate()'
+      - julia --project=docs/ docs/docs/make.jl
+      after_success: skip
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,8 @@ jobs:
     - stage: "Documentation"
       julia: 1.0
       os: linux
+      env:
+        - TRAVIS_REPO_SLUG=github.com/JuliaImages/juliaimages.github.io.git
       script:
       - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
                                              Pkg.instantiate()'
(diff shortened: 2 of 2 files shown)


johnnychen94:
Great that it works, so the next step is to add SSH key pair generated by `DocumenterTools` to travis's secrets(both Images and juliaimages.github.io) and juliaimages.github.io's deploy key(with write permission), which I don't have permission to. @timholy @Evizero 

After that, we could merge this PR and then manually make a temp tag to check if it successfully versions the documentation.

Or we could use Github Actions for documentation build. An example can be found in [DemoCards](https://github.com/johnnychen94/DemoCards.jl/blob/master/.github/workflows/docs.yml)

timholy:
I am not sure I fully understand the aim here; what's the goal of building the docs with Images.jl?

johnnychen94:
Any updates(especially new `Images.jl` release) should trigger a documentation rebuild, and should tag the documentation when necessary. Currently, documentation is always built on the `latest`, which isn't so good.

Of course, we could manually tag a release in juliaimages.github.io side, but that's not a programmatic way.

timholy:
Can we do this through CompatHelper now? Whenever a new release gets made in any dependency that requires a bump in bounds, wouldn't there be a notification?

Oh, wait, all the "0" and "1" in the `[compat]` section of https://github.com/JuliaImages/juliaimages.github.io/blob/source/Project.toml will block that, right? What if we went back to the idea of declaring specific (minor) versions?

timholy:
:+1: to the idea of versioned docs. But what should that be? AFAICT everyone wants Images.jl to just become a meta-package, and once that's done it is unlikely to release new versions on a regular basis.

johnnychen94:
> What if we went back to the idea of declaring specific (minor) versions?

That works. But actually I'm thinking one step further on how to manage all the documentations organically. For example, if we introduce `ImageBinarization` into `Images.jl`, should we add a page to juliaimages.github.io explaining how to use `ImageBinarization`? If yes, what's the relation between Images's documentation and ImageBinarization's documentation(https://github.com/JuliaImages/Images.jl/issues/542#issuecomment-254438376)?

The map in my mind is:

1. trigger a doc rebuild on each release of sub-packages under JuliaImages
2. host docs of sub-packages as submodules of Images's docs, so that every update on sub-packages' docs syncs to Images's docs. (An alternative might be a simple URL link from Images.jl's docs)
3. The role of Image's docs is to
  1. plot the general map of the ecosystem, design philosophy and dev manual
  2. host demos
  3. leave other details to sub-packages

> everyone wants Images.jl to just become a meta-package, and once that's done it is unlikely to release new versions on a regular basis.

As long as we limit [compat] of Images.jl to patch level, CompatHelper helps to do so. For example, each patch release of sub-packages triggers a patch release of Images.jl. Some discussion about this is on https://github.com/JuliaImages/Images.jl/issues/825

johnnychen94:
> Whenever a new release gets made in any dependency that requires a bump in bounds, wouldn't there be a notification?

Actually I want things to go as it should without any notification...

johnnychen94:
> Whenever a new release gets made in any dependency that requires a bump in bounds, wouldn't there be a notification?

Actually I want things to go as it should without any notification...

timholy:
> if we introduce ImageBinarization into Images.jl, should we add a page to juliaimages.github.io explaining how to use ImageBinarization? If yes, what's the relation between Images's documentation and ImageBinarization's documentation(#542 (comment))?

I'm torn on this, but in general I don't think it's useful for users to go to ImageCore for documentation on functionality defined in ImageCore, to Colors for its docs, etc. Some packages have their own docs but that's a bit legacy right now. Even ImageFeatures & ImageSegmentation, which are *not* loaded into Images, have pages on JuliaImages.

When a package has documentation that's also hosted locally, it would be ideal if both sets of docs are in sync. Perhaps we should set up a system where the repo's docs get symlinked prior to building the docs at JuliaImages?

> As long as we limit [compat] of Images.jl to patch level, CompatHelper helps to do so. For example, each patch release of sub-packages triggers a patch release of Images.jl, this also triggers a doc rebuild if we take this PR's strategy. Some discussion about this is on #825

That's reasonable. But once we hit 1.0 on packages, then wouldn't we declare bounds as `ImageCore = "1"`? The idea on semver is that you shouldn't have any breaking changes without bumping the major version. New features get added in minor releases, but since those should break any dependent package, why would we need to register a new release of Images?

timholy:
We could also use GitHub Actions to write custom logic for rebuilding the docs. Perhaps dating the docs, rather than giving them a version number, makes most sense?

johnnychen94:
Hmmm, not quite so IMO. I'll draft the issue later this week on this more carefully. Now I need to catch up with my own work.

PR #860, opened 2020-01-10
Notation/naming tweaks to computeft!

timholy:
This is purely cosmetic, but it increases the similarity to the pseudocode in the paper.

(I'm using this as an example of Julia's advantages in a grant application. See also https://discourse.julialang.org/t/in-as-few-lines-as-possible-describe-why-you-love-julia/33179/16)


johnnychen94:
Oops apparently I messed things up in #858

johnnychen94:
The broken test is fixed by #861

johnnychen94:
Looks like this PR needs more time to fix the Julia 1.0 compatibility issue, so I've just tagged a release for #861

PR #861, opened 2020-01-10
support ImageMetadata v0.9.0

johnnychen94:
ImageMetadata v0.9.0 deprecates the symbol `data` in favor of `arraydata`,
and this would cause a name conflict with ImageAxes when we import both
packages. This commit rebinds the deprecation of `data` to `arraydata` and thus
fixes this issue.

Besides this, this package fixes the depwarn brought by ImageMetadata v0.9.0,
and provides backward compatibility to ImageMetadata < v0.9.0

This commit also removes ImageAxes v0.5 from the compat section, it
doesn't affect anything because ImageAxes v0.5 isn't compatible to
ImageCore v0.8.1 (it's ImageCore = "0.2-0.7" according to the General
registry)

cf. this commit fixes the broken test in #858


johnnychen94:
I must add a lot of unnecessary tests to the CI with my frequently force-push... just feel free to closes them...

PR #865, opened 2020-01-23
Adds deprecation warnings and exports ImageContrastAdjustment

zygmuntszpak:
Also removes code in exposure.jl which is now found in the ImageContrastAdjustment.jl package.


zygmuntszpak:
Addresses part of https://github.com/JuliaImages/Images.jl/issues/841

zygmuntszpak:
Are we OK to bump the minimum requirement for Julia for the `Images` package to version 1.1?  The minimum requirement for ImageContrastAdjustment is Julia 1.1 and I now recall that this is because I used `Base.@kwdef` in various place. That function only got introduced in Julia version 1.1 https://pkg.julialang.org/docs/julia/THl1k/1.1.1/NEWS.html

Tokazama:
I personally think that if JuliaPro has reached a version number for a while it's probably safe to set it as the minimum version number. They seem to be pretty proactive about making sure packages work with each new JuliaPro release.

johnnychen94:
Unless there's another LTS version, dropping 1.0 compatibility is really risking

Tokazama:
I guess we should probably consider which ["Risk Tolerance Persona"](https://julialang.org/blog/2019/08/release-process/) we should cater to.

timholy:
Yeah, I'm a bit uncomfortable about not continuing to support Julia 1.0. How bad would it be to just write out the default-filling constructors manually?

Otherwise, this is really nice to see!

zygmuntszpak:
I've tagged a new version of ImageContrastAdjustment which now supports Julia 1.0. Hence, I am going to close and re-open this pull-request to trigger a new build.

zygmuntszpak:
@timholy I'll wait for you to approve this before I merge.

src/exposure.jl
```diff
@@ -1,3 +1,18 @@
+### Deprecation warnings in view of functionality in ImageContrastAdjustment.jl
+# supplanting all the functionality in exposure.jl
+
+Base.depwarn("`imhist` will be removed in a future release, please use `build_histogram` instead.", :imhist)
+Base.depwarn("`imadjustintensity` will be removed in a future release, please use `adjust_histogram(img, LinearStretching())` instead.", :imadjustintensity)
```
timholy: When deprecating something, we need to make sure that users' old code continues to work properly, but they get the warning. Unless I'm misunderstanding, the way this has been done just breaks the old code---there is no method called `imhist` anymore. It's generally good practice to run the old tests with the new code and make sure all the tests pass even though they issue warnings. (Failure is OK when it's a test like `@allocated(foo(x, y)) == 0`, because the depwarn allocates memory.)

To do this more gracefully, there are two strategies:
- `@deprecate foo(x, y) bar(x, y+1)`: use this form when `bar(x, y+1)` is a drop-in replacement for `foo(x, y)`, meaning it returns functionally the same thing. (Differences in rounding, etc, would also be fine as far as I am concerned.)
- Leave the old code around but insert the `Base.depwarn` call in the body of the old method: use this form when the replacement returns something that is functionally different from the old code. That way, users' code still works exactly the way it always did, and it's when they have time to sit down and fix the warning that they'll discover the differences.

In this case, I think the binning frequently differs so most likely it will be easiest to leave the old code around but perhaps move it to a file called `deprecations.jl`.

You can also use `Base.depwarn` directly if you have to write a "wrapper" function, i.e., if it's a bit more complicated than `bar(x, y+1)`. The `@deprecate` syntax is just darned convenient when it works.

zygmuntszpak: The code for all the methods that have been listed as deprecated in `exposure.jl` is still there and works. For example, `imhist` works just the way it did before. I also didn't remove the tests for any of the functions that are listed as deprecated. 

However, there is a subtlety that you drew my attention to that I didn't consider carefully enough. I actually broke the API of the functions that were the _precursor_ of the new exposure API and I didn't actually give a deprecation message for those. For example, `exposure.jl` had an implementation such as 

`adjust_histogram(Equalization(),img, nbins, minval, maxval)` 

 which evolved to become 

`adjust_histogram(img, Equalization(nbins = X,  minval = Y, maxval = Z))` 

for the `ImageContrastAdjustment` package. The final form of the API was a culmination of  various API discussions with Johnny. 

Note that the nature of the struct `Equalization` has changed between the precursor and final version. It now takes keyword parameters that were previously passed as parameters to the function `adjust_histogram`. I haven't deprecated the old `adjust_histogram` and corresponding structs. I just removed them and their tests, and reexported the new versions which are now in `ImageContrastAdjustment`. Hence, someone who wasn't using the established `imhist` API, but started using the precursor API `adjust_histogram` that appeared in `Images` will have their code broken.

I can think of a way to redirect the precursor API so that it effectively calls the new API under the hood for all cases except for `Matching()`. The issue is that under the new API, you have to specify the target image whose histogram you want to match when you instantiate `Matching`, i.e. you have to do `Matching(targetimg = X)`. You will get an error  `Field 'targetimg' has no default, supply it with keyword` if you do not. The precursor API, however, expects that one can instantiate `Matching()`.  Do you have any suggestion for how to proceed in this instance?

timholy: If it can at least throw an error message that indicates how to fix it, I think you'll have done due diligence.

Thanks for explaining the rest of the strategy. I am then confused about the purpose of these `depwarn` calls. They run when the package builds, not when the method gets called, right?

zygmuntszpak: Yes, they run when the package gets build (i.e. when `using Images` is evaluated), but not when the method gets called. I will revise things so that they also get called each time you run the method as well.

timholy: No need to have them run when it gets built. For people who aren't using those functions, that's just spam.

src/Images.jl
```diff
@@ -53,6 +53,7 @@ const is_little_endian = ENDIAN_BOM == 0x04030201
 @reexport using ImageFiltering
 @reexport using ImageMorphology
 @reexport using ImageDistances
+@reexport using ImageContrastAdjustment
```
timholy: Just checking, this doesn't cause any name conflicts from the fact that we have both `Images.build_histogram` and `ImageContrastAdjustment.build_histogram`? If it does, better to `import ImageContrastAdjustment` and then manually export what you need.

src/deprecations.jl
```diff
@@ -0,0 +1,73 @@
...
+
+
+function adjust_histogram!(operation::Equalization, img::AbstractArray, nbins::Integer, minval::Union{Real,AbstractGray} = 0, maxval::Union{Real,AbstractGray} = 1)
+    Base.depwarn("`adjust_histogram!(Equalization(),img)` will be deprecated, use adjust_histogram!(img,img, Equalization(; nbins, minval, maxval)) instead.", :adjust_histogram!)
+    ImageContrastAdjustment.adjust_histogram!(img, img, Equalization(nbins = nbins, minval = minval, maxval = maxval))
+end
+
+function ImageContrastAdjustment.Matching()
+     Base.depwarn("`Matching()` will be deprecated, use Matching(targetimg = targetimg) instead.", :adjust_histogram)
+    return ImageContrastAdjustment.Matching(targetimg = zeros(3,3))
```
timholy: Will this actually return the same result they got before? If not, might be best just to make this an error.

src/deprecations.jl
```diff
@@ -0,0 +1,73 @@
...
+end
+
+
+function adjust_histogram!(operation::Equalization, img::AbstractArray, nbins::Integer, minval::Union{Real,AbstractGray} = 0, maxval::Union{Real,AbstractGray} = 1)
+    Base.depwarn("`adjust_histogram!(Equalization(),img)` will be deprecated, use adjust_histogram!(img,img, Equalization(; nbins, minval, maxval)) instead.", :adjust_histogram!)
+    ImageContrastAdjustment.adjust_histogram!(img, img, Equalization(nbins = nbins, minval = minval, maxval = maxval))
+end
+
+function ImageContrastAdjustment.Matching()
+     Base.depwarn("`Matching()` will be deprecated, use Matching(targetimg = targetimg) instead.", :adjust_histogram)
```
timholy: There's an "overwritten method" warning in the Travis logs here. Seems like Parameters generates a zero-arg constructor, though I don't understand why. (The `targetimg` argument is required, right?)

src/Images.jl
timholy: Seems like it's fine, from the Travis logs.

src/deprecations.jl
zygmuntszpak: Originally the `Matching()` type was used solely to determine what operation to dispatch on. The API required a user to type something like `adjust_histogram(Matching(),img, targetimg, nbins)`.  There was no other way `Matching()` could be used.  

The new API requires `adjust_histogram(img, Matching(targetimg = targetimg, nbins = nbins))` and `targetimg` is a required keyword since no sensible default value exists (we cannot predict  what image you want to match to). 

In order not to break anyones code in the transition to the new API, I defined an empty constructor `Matching()` in `Images.jl` which sets `targetimg` to a placeholder value. This means that their code 

`adjust_histogram(Matching(),img, targetimg, nbins)`

will not error. I ensure that their previous code gives the correct and equivalent result by instantiating the required expression, e.g.

```julia
function adjust_histogram(operation::Matching, img::AbstractArray, targetimg::AbstractArray, nbins::Integer = 256)
    ImageContrastAdjustment.adjust_histogram(img, Matching(targetimg = targetimg, nbins = nbins))
end
```

If we don't do this and let the code error then won't we be breaking semvar? Or perhaps you were intending to ultimately tag the next release as a breaking change?

src/deprecations.jl
zygmuntszpak: I'm not sure why Parameters is doing this. The value `targetimg` is required, and when one calls `Matching()` without it one receives an appropriate error. I added the zero-arg constructor for the reasons given above.

src/deprecations.jl
timholy: Got it! Sorry, I should have figured out your intent here on my own.

I am torn about what to do with version numbers wrt deprecations. They generally succeed in making me modifiy my code immediately, which in effect makes them breaking. But I know many other people disagree.

src/deprecations.jl
timholy: Looks like `Base.@kwdef` does the same thing. Seems to be a consequence of creating the keyword-only constructor.

There seem to be two options: put this definition in ImageContrastAdjustment (it doesn't seem to complain if it's in the same module), or something like
```julia
struct Matching{T₁ <: AbstractArray,
                            T₂ <: Union{Integer, Nothing},
                            T₃ <: Union{AbstractRange, Nothing}} <: AbstractHistogramAdjustmentAlgorithm
    targetimg::T₁
    nbins::T₂
    edges::T₃
end

Matching(targetimg; nbins=256, edges=nothing) = Matching(targetimg, nbins, edges)
```
That way you always have to call it with one positional argument.

zygmuntszpak: It turns out that while the first option you suggested did not throw a warning when I tried it on my local dev branch, it does actually throw a warning when you pull the package from the General registry and try to use it. In other words, now we get the warning in ImageContrastAdjustment@0.3.2 instead of in Images. 

The second option is not suitable because it constitutes a breaking change in `ImageContrastAdjustment` itself. I have several tests where I use the keyword `targetimg` when constructing `Matching()`. 

In light of this, and since we don't want the warning message to appear, I will tag a new version of ImageContrastAdjustment which undos 0.3.2 and then just make the creation of `Matching()` an error in `Images.jl`.

zygmuntszpak:
- Moved all deprecated code from `exposure.jl` into `deprecations.jl`
- Deprecation messages say "is deprecated" instead of "will be deprecated".
- Removed mention of `adjust_histogram`, `adjust_histogram!`, `Matching`, `Equalization` and `build_histogram` from export statement  (it is now automatically `reexported`)
- Rerouted old variants of  `adjust_histogram`, `adjust_histogram!`, and `build_histogram` so that they call the new API in ImageContrastAdjustment. User receives a deprecation message, but their code still works (with the exception of code that used `Matching()`).
- Any code that matched histograms using `adjust_histogram(Matching(), ...)` is broken. The user receives an error message stating that the keyword `targetimg` in the construction of `Matching` needs to be supplied.

johnnychen94:
It's the Spring festival here in China so I didn't follow the whole discussion here with my limited time. Looks like there're still some breaking changes, so a minor version bump is required, i.e., `0.21.0`

timholy:
Thanks for merging, I've been swamped! We should get this out as 0.21, will try to get to it over the weekend.

timholy:
And it's out! Thanks again @zygmuntszpak, great contribution!

tlnagy:
The suggested fix in the dep warning for `imadjustintensity` doesn't work:

```julia
julia> using Images

julia> img = Gray{N0f8}.(rand(10, 10));

julia> imadjustintensity(img);
┌ Warning: `imadjustintensity` will be removed in a future release, please use `adjust_histogram(img, LinearStretching())` instead.
│   caller = top-level scope at REPL[10]:1
└ @ Core REPL[10]:1

julia> adjust_histogram(img, LinearStretching())
ERROR: MethodError: no method matching adjust_histogram(::Array{Gray{Normed{UInt8,8}},2}, ::LinearStretching{Float64,Float64})
Stacktrace:
 [1] top-level scope at REPL[11]:1
```

zygmuntszpak:
It looks like the reason for this is that I forgot to add
```julia
import ImageContrastAdjustment: adjust_histogram, adjust_histogram!
```
to ensure that the new definitions of `adjust_histogram` and `adjust_histogram!`  in `deprecations.jl` don't make the definitions in `ImageContrastAdjustment` "disappear". 

I'm putting together a quick fix, and will add some test to make sure that the advertised API is actually callable. Sorry for the oversight, and thanks for reporting the issue.

timholy:
Another thing we should have done is fill in the parameters for the actual usage; anyone who copy/pastes the suggested deprecation would be throwing away whatever custom parameters they're supplying for the call. A prototype of a better deprecation might have been
```julia
@deprecate clahe(img, nbins=100) adjust_histogram(img, AdaptiveEqualization(; nbins))
```
and then the user will see a copy/pastable suggestion.

Issue #867, opened 2020-02-04
Docs clarification for functions downstream of imgradients: g1, g2 vs gy, gx


stillyslalom:
Many of the [edge-detection functions](https://juliaimages.org/latest/function_reference/#Edge-detection-1) require the gradients of an image, _e.g._ `orientation(grad_x, grad_y)`. However, `imgradients` now delivers gradients along each axis of an image (`g1`, `g2`, ...), but it seems that the idea hasn't propagated through the entirety of the codebase:

```
julia> grad_y, grad_x, mag, orient = imedge(img, KernelFactors.bickley)

julia> orientation(grad_y, grad_x) == orient # true

help?> orientation

  orientation(grad_x, grad_y) -> orient

  Calculate the orientation angle of the strongest edge from gradient images 
given by grad_x and grad_y. Equivalent to atan(grad_x, grad_y). When both 
grad_x and grad_y are effectively zero, the corresponding angle is set to zero.
```

timholy:
Oh dear! Thanks for noticing this. It will take some thought about how to fix this; switching argument order is just about impossible to deprecate.

zygmuntszpak:
As part of our move to a consistent API I was contemplating the creation of ImageEdgeDetection.jl which functions such as `detect_edges(img, Canny(args...)`. Perhaps we can introduce functions with slight different names such as `determine_orientation` which allows one to specify the co-ordinate system convention.  Users can then migrate their code to the new API and the old functions can call the new one with their old co-ordinate system conventions.

stillyslalom:
Another option: returning a NamedTuple `grads=(gy, gx)` from `imgradients`, and adding new single-argument methods for all the downstream edge-detection functions:
```
gy, gx = grads = imgradients(img, KernelFactors.bickley)
magnitude(grads)
phase(grads)
```
These methods could also take information from image traits, e.g. [`spacedirections`](https://juliaimages.org/latest/function_reference/#ImageCore.spacedirections). One of my longstanding gripes with image processing is the translation from physical coordinates (x, y, z) to image coordinates (-y, x, z), which makes the trigonometry of these sorts of functions harder to reason about.

stillyslalom:
[...]

PR #868, opened 2020-02-05
Fix g1/g2 vs g_x/g_y scrambles

stillyslalom:
ref. https://github.com/JuliaImages/Images.jl/issues/867


johnnychen94:
I think issue #867 is fixed by https://github.com/JuliaImages/ImageEdgeDetection.jl/pull/10 so I'm closing this one.

PR #870, opened 2020-02-07
Bumps version number in preparation for patch release

zygmuntszpak:


timholy:
Thanks so much for handling this issue! Just FYI I want you to feel empowered to make releases. You're doing awesome work, and making JuliaImages much better as a consequence.

PR #873, opened 2020-02-11
reexport ImageQualityIndexes

### LINKED ISSUE #811: import ImageQualityIndexes
I'd like to reexport `ImageQualityIndexes.jl` in `Images.jl`, drop julia 0.7 compatibility, and bump a minor release. But I'm not sure if this is the appropriate now:

* `ImageQualityIndexes.jl` requires `ImageCore = ">=0.8.3"`. There're still some non-JuliaImages packages not switched to the new registrator and incompatible to `ImageCore v0.8` in `General`.
* Both `StatsBase` and `ImageQualityIndexes.jl` exports `psnr`. (Looks like `StatsBase` isn't a direct dependency anymore, at least I'm told so from tests by removing it.)

johnnychen94:
closes #811

This is not the perfect solution because it uses the same name with `StatsBase.psnr`, but it's the simplest one :P Bumped a minor version for safety and hope nobody's using `StatsBase.psnr` together with `Images`...

Another choice is to extend StatsBase's implementation, then I'll need to add it as a dependency of ImageQualityIndexes. However, in `Images.jl` there's a comment to remove it as a dependency...

https://github.com/JuliaImages/Images.jl/blob/d8ea4d7b2fd9505d5a010549ab34e3269eca1279/src/Images.jl#L41


johnnychen94:
rebase and merge

PR #879, opened 2020-03-07
Bug/move std to cvs

### LINKED ISSUE #698: Shouldn't complement live in Colors.jl?
Small function, small issue, but seems like `complement` which acts on a color should live in `Colors.jl`.

### LINKED ISSUE #872: move `var`, `std` to ColorVectorSpace
The right place for `var` and `std` should be ColorVectorSpace

https://github.com/JuliaImages/Images.jl/blob/2176612df7e863e534af66d146227523d0685715/src/algorithms.jl#L79-L97

Steps to achieve it:

1. copy codes to `ColorVectorSpaces`, merge and ask for a new release
2. remove codes here: since ColorVectorSpace is loaded in this package, we don't need to deprecate them. But we need some compatibility-checking codes so that if an older version of `ColorVectorSpace` is installed, this package still functions normally.

wizofe:
Remove std, mean and complement and migrate to `ColorVectorSpace.jl` including tests.

closes #872
closes #698


johnnychen94:
Thanks! Looks like exposure.jl needs to be removed as well (don't forget to remove the line `include("exposure.jl")` in Images.jl)

The test here fails because https://github.com/JuliaGraphics/ColorVectorSpace.jl/pull/125 isn't merged yet. Retrigger the test after a new release in ColorVectorSpaces should give it a green check

wizofe:
Thank you @johnnychen94 for your review. I did both changes and hopefully the tests will go through after merging as you predicted :) 

Indeed I removed `exposures.jl` earlier but I forgot to push. Leaving this open now and moving my efforts to the other PR.

johnnychen94:
> Leaving this open now and moving my efforts to the other PR.

Sure, as you might already notice, [the step 1](https://github.com/JuliaGraphics/ColorVectorSpace.jl/pull/125) doesn't go very smoothly and more discussion is needed there. But for this PR I think it's ready.

Doing things parallelly is definite one most common thing in open-source would because most people don't do it as a full-time job and might not respond very quickly.

Looking forward to your next PR!

kimikage:
JuliaGraphics/ColorVectorSpace.jl#125 was closed, and JuliaGraphics/ColorVectorSpace.jl#144 was merged.

Therefore, we need to limit the lower bound of `[compat]` (of [ImageCore?](https://github.com/JuliaImages/Images.jl/pull/937#issuecomment-765887105)) or add a `hasmethod` check to prevent redefinition of `complement`.

cc: @johnnychen94, @timholy

PR #895, opened 2020-06-17
Port bwdist code to ImageMorphology

tlnagy:
Fixes https://github.com/JuliaImages/Images.jl/issues/883. This should not be merged until the following are complete:

BLOCKING

- [x] https://github.com/JuliaImages/ImageMorphology.jl/pull/28 needs to be merged first
- [x] Tag new release of ImageMorphology


johnnychen94:
CI fails because pkg server, which is enabled since Julia 1.5, doesn't get synced yet, I'll retrigger the CI later today.

johnnychen94:
I should have said that this workflow isn't perfect before merging this, e.g., Images v0.22.2 isn't compatible to ImageMorphology v0.2.6.

So I'll need to bump a patch version for Images very quickly so that users don't hit this incompatibility.

timholy:
Yes, I've begun to think that semver + Pkg's version resolver really *can't* solve the problem fully. Adding a new method to ImageMorphology is not a breaking change in and of itself, only in the context of some other environment. It also runs into problems like https://github.com/JuliaLang/Pkg.jl/issues/1251#issuecomment-647131387.

PR #902, opened 2020-06-29
port `clearborder` from Images.jl to ImageMorphology.j

mgautam98:
Corrosponding PR https://github.com/JuliaImages/ImageMorphology.jl/pull/29 for moving `clearborder` from Images.jl to ImageMorphology.j


Project.toml
```diff
@@ -40,7 +40,7 @@ ImageCore = "0.8.3"
 ImageDistances = "0.2.5"
 ImageFiltering = "0.6.3"
 ImageMetadata = "0.6, 0.7, 0.8, 0.9"
-ImageMorphology = "0.2.6"
+ImageMorphology = "0.2.7"
```
johnnychen94, line 43: We need to bump the version here so that users don't accidentally run into ImageMorphology<= 0.2.6, in which case `clearborder` is missing in both packages.

I've directly modified this in my side and push to your commit.

mgautam98: Oh! I missed it. Thanks

johnnychen94: Said this, the best way to avoid such version conflict is to check if the methods exists, an example can be found at

https://github.com/JuliaImages/Images.jl/blob/27f880eab79fe348f9203ba38226912ad3e22c58/src/Images.jl#L68-L76

This should be the safest way.

Since `clearborder` isn't such a fundamental operation, and we only bump patch versions on both Images and ImageMorphology, I *think* it's relatively safe to do so. We could just fool ourselves, save us some efforts from maintaining dependencies, and fix it when there's a bug report coming in.

johnnychen94:
It's weird that CI doesn't run, but I've tested it locally it works as expected so I'll merge it and tag a release.

PR #903, opened 2020-07-11
Update docstring for imcorner

merckxiaan:
The docstring of imcorner seems to be outdated as of #617 , here's a quick fix.
English is not my first language so I don't doubt that sentence could be made a bit clearer.


johnnychen94:
Thanks! I've made some changes to it, hope you don't mind 😄

PR #904, opened 2020-07-21
[deprecation] shepp_logan and one(::CartesionIndex)

johnnychen94:
This PR actually consists of four independent changes (If needed, I could rebase it into four commits...)

* deprecate `Images.shepp_logan` in favor of `TestImages.shepp_logan`
* `one(::CartesianIndex)` -> `oneunit`
* suppress the noisy deprecation output; we only need to make sure it runs and test passes
* `QuartzImageIO` is removed as test dependency; it's more reliable to use only one IO backend

It's worth noting that there is a name conflict between `TestImages.shepp_logan` and `Images.shepp_logan`. The only solution to this that I can come up with is to reexport `TestImages.shepp_logan`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -50,6 +50,7 @@ OffsetArrays = "0.8, 0.9, 0.10, 0.11, 1.0.1"
 Reexport = "0.2"
 StaticArrays = "0.8, 0.9, 0.10, 0.11, 0.12"
 StatsBase = "0.24, 0.25, 0.26, 0.27, 0.28, 0.29, 0.30, 0.31, 0.32, 0.33"
+Suppressor = "0.2"
 TiledIteration = "0.2"
 julia = "1"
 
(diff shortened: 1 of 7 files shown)


timholy:
It seems worth considering re-exporting to avoid the name conflict. I guess that would transiently make TestImages a dependency of Images? Are there big downsides to that? I wouldn't want people relying on it...but I guess as long as we don't re-export *all* of TestImages but just that one function, seems like it should be OK.

johnnychen94:
> It seems worth considering re-exporting to avoid the name conflict. I guess that would transiently make TestImages a dependency of Images?

Yes, it seems that there isn't a good way to un-export `shepp_logan` from Images.jl. I see three options:

Option 1:

1. un-export `TestImages.shepp_logan`; this is quite safe because we just released it yesterday.
2. deprecate `Images.shepp_logan`
3. remove `Images.shepp_logan` in a future breaking release
4. export `TestImages.shepp_logan`

Option 2:

1. remove `Images.shepp_logan` right now and tag a breaking release

Option 3:

1. reexport `TestImages.shepp_logan` in Images.jl
2. remove this symbol in Images v1.0.0 release; there's no way to deprecate this.


I'm going to take option 1 here.

PR #905, opened 2020-07-22
[WIP] reexport ImageBinarization and HistogramThresholding

johnnychen94:
We need to review and update both packages before we reexport them here

TODO: (there will be more as we proceeding this)

- [ ] HistogramThresholding
- [ ] ImageBinarization
  - [ ] [dispatch binarize on HistogramThresholding.ThresholdAlgorithm](https://github.com/zygmuntszpak/ImageBinarization.jl/issues/67)

The compilation time contributed by them doesn't look good right now

```
# Version 1.6.0-DEV.497 (2020-07-20)

# before
julia> @time using Images
  3.687936 seconds (9.14 M allocations: 584.557 MiB, 4.90% gc time)

# after
julia> @time using Images
  4.420445 seconds (11.35 M allocations: 714.150 MiB, 4.69% gc time)
```

cc: @zygmuntszpak


timholy:
Closed by #1030

PR #919, opened 2020-10-08
Reexport the deprecated permuteddimsview

timholy:
Currently tests are failing; it turns out that deprecations aren't re-exported automatically. This does it manually.

The absence of reports about this does suggest that most people seem to have figured this out on their own, but it is a little troubling.

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -28,6 +28,9 @@ using SparseArrays: findnz
 
 using Reexport
 @reexport using ImageCore
+if isdefined(ImageCore, :permuteddimsview)
+    export permuteddimsview
+end
 if isdefined(ColorTypes, :XRGB) && isdefined(ColorTypes, :RGB1)
     Base.@deprecate_binding RGB1 XRGB
     Base.@deprecate_binding RGB4 RGBX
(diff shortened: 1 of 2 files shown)


johnnychen94:
Actually, this is a test-only issue so I think it's okay to not export it.

```julia
$julia --startup=no --depwarn=yes
  
(@v1.5) pkg> st Images ImageCore
Status `~/.julia/environments/v1.5/Project.toml`
  [a09fc81d] ImageCore v0.8.17
  [916415d5] Images v0.22.4

julia> using Images

julia>
```

timholy:
Prior to this PR,
```julia
julia> using Images
[ Info: Precompiling Images [916415d5-f1e6-5110-898d-aaa5f9f070e0]

julia> A = [1 2; 3 4]
2×2 Matrix{Int64}:
 1  2
 3  4

julia> permuteddimsview(A, (2, 1))
ERROR: UndefVarError: permuteddimsview not defined
Stacktrace:
 [1] top-level scope
   @ REPL[3]:1

julia> using ImageCore

julia> permuteddimsview(A, (2, 1))
2×2 PermutedDimsArray(::Matrix{Int64}, (2, 1)) with eltype Int64:
 1  3
 2  4
```

PR #920, opened 2020-10-08
Get MappedArrays via ImageCore

timholy:
Avoids the need for a [compat] bound


johnnychen94:
I've wanted this for so long but not sure if this is really desired!

johnnychen94:
To be clear, I didn't realize this is doable; I thought MapppedArrays needed to be reexported in ImageCore somehow. Given the fact that not all downstream packages need the Symbol, I didn't propose to reexport that.

This is a cleaner solution to me. 👍

timholy:
I'm not sure when I realized this was a possibility either. We do have to make sure that ImageCore's bounds on its dependencies are strict enough to satisfy needs; I think there was at least one package that I passed by because it seemed more restrictive.

PR #927, opened 2020-11-22
deprecate ColorizedArray in favor of MappedArray

johnnychen94:
`MappedArray` provides a more generic version of `ColorizedArray` so I believe it is good to not maintain `ColorizedArray` anymore.

`LazyArray` also provides similar operation; `LazyArray(@~ @. I * indexed_img)`.

closes https://github.com/JuliaImages/juliaimages.github.io/issues/132

### DIFF
--- a/src/labeledarrays.jl
+++ b/src/labeledarrays.jl
@@ -18,6 +18,8 @@ associated with that point's label.
 This computation is performed lazily, as to be suitable even for large arrays.
 """
 function ColorizedArray(intensity, label::IndirectArray{C,N}) where {C<:Colorant,N}
+    depwarn("`ColorizedArray(intensity, label)` is deprecated, use `mappedarray(*, intensity, label)` from `MappedArrays` instead", :ColorizedArray)
+
     axes(intensity) == axes(label) || throw(DimensionMismatch("intensity and label must have the same axes, got $(axes(intensity)) and $(axes(label))"))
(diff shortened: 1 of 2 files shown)


timholy:
The one thing I worry about is ImageView. If I remember correctly, hover-over-pixel for a ColorizedArray displays the underlying value, not the mapped value. But we can do this if you think it's better.

johnnychen94:
> If I remember correctly, hover-over-pixel for a ColorizedArray displays the underlying value, not the mapped value.

I see. It does give some visual aids, but this doesn't sound like what `ColorizedArray` is designed for. Probably making hover-over-pixel behavior customizable is a better choice. I don't use ImageView but I guess it's similar to https://github.com/JuliaImages/ImageView.jl/issues/228.

timholy:
> this doesn't sound like what ColorizedArray is designed for

Not sure I really remember, but I think that is its main purpose. But you're right there are other ways to solve this. If this just seems too marginal we can do it, I was just thinking that there are many cases where array content and array labeling are two things that are useful to be able to link without losing track of either one.

johnnychen94:
I get your point here. Given that you are probably the only user of this, I'll leave this to you to decide whether we should do it or not.

BTW, I search the whole GitHub and couldn't find any place other than Images that contains `ColorizedArray` 😨

timholy (approved):

PR #935, opened 2021-01-14
added some more tests to arrays.jl

krishna-vasudev:
Have added tests for different dimensions of arrays, the only type present was for 2 by 2 matrix,so i added tests for a 2 by 3 matrix.

### DIFF
--- a/test/arrays.jl
+++ b/test/arrays.jl
@@ -24,7 +24,7 @@ using Images, IndirectArrays, Test
    target1 = intensity1 .* labels1
    @test eltype(A1) == RGB{Float64}
    @test size(A1) == (2,3)
-   @test axes(A1) == (Base.OneTo(2), Base.OneTo(3))
+   @test axes(A1) == (1:2, 1:3)
    for i = 1:6
        @test A1[i] === target1[i]
    end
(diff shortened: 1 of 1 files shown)


test/arrays.jl
```diff
@@ -17,6 +17,23 @@ using Images, IndirectArrays, Test
...
         @test a === t
     end
+
+    intensity1 = [0.1 0.3 0.6; 0.2 0.4 0.1]
+   labels1 = IndirectArray([1 2 3; 2 1 3], [RGB(1,0,0), RGB(0,0,1),RGB(0,1,0)])
+   A1 = ColorizedArray(intensity1, labels1)
+   target1 = intensity1 .* labels1
+   @test eltype(A1) == RGB{Float64}
+   @test size(A1) == (2,3)
+   @test axes(A1) == (Base.OneTo(2), Base.OneTo(3))
```
timholy, line 27:
```suggestion
   @test axes(A1) == (1:2, 1:3)
```
unless you think it's important to test type-equality, in which case you need to use `===` rather than `==` (and I don't see why that would be important).

timholy (commented):
LGTM

timholy:
Thanks!

PR #948, opened 2021-03-28
add io backends ImageIO and ImageMagick as dependencies

### LINKED ISSUE #24: Move away from ImageMagick
Relying on ImageMagick seems to be presenting problems for Mac users (see #15, #23), and also leads to slow performance. For at least PNG, JPG, and TIF it would be good to find an alternative, presumably by directly calling the corresponding libraries.

### LINKED ISSUE #824: ImageMagick should be in the [deps] section in Project.toml
I use Images.jl **only** to save a jpg image, using this statement:

using Images
...
save("filename.jpg", someUInt8Matrix)

But my project/enviroment needs ImageMagick to be added before this works. I'm not using it myself directly (I don't have a "using ImageMagick" in my code).

Somewhere Images.jl does use it. I should not have to add "ImageMagick" to my project, it should automatically be added to my Manifest when I Pkg.add Images.

You can replicate by temporarily removing your enviroment, add just Images, and try and use it...

johnnychen94:
QuartzImageIO is not added for two reasons: 1) it only works for macOS, and 2) it has less functionality than ImageMagick.

Note that the julia compat "1" now becomes a fake one as the declared ImageIO and ImageMagick compatibility versions require `Julia >= v"1.3"`

@timholy Another thing that catches my interest: since `TestImages >= v"1.2"` are also backed by artifacts, could we also reexport this package in Images.jl? (One more step to make Images an umbrella package)

closes #24
closes #824


johnnychen94:
I think we now need to test four Julia versions in CI:

- `1.3`: the lowest version that we claim compatible
- `1`: the latest stable version, which happens to be `1.6` right now.
- `1.6`: the latest LTS version
- `nightly`: to early find out upstream breaking changes.

johnnychen94:
Since we've already discussed this in https://github.com/JuliaImages/Images.jl/issues/824#issuecomment-683412325 I'll just merge this version. It's not a breaking version, but since it restricts the Julia compatibility to 1.3, I bumped the minor version here. I don't plan to tag a release for this immediately, so I add the `-dev` suffix.

PR #956, opened 2021-04-27
Images v0.24.1

johnnychen94:
Images v0.24.0 is somehow buggy for ColorVectorSpace 0.9 so we need to introduce patch #955 to fix it. This release resolves the compatibility issue https://github.com/JuliaImages/Images.jl/pull/954#issuecomment-827361075 found by our Sharlock @kimikage.

This will be the last release that compatible to ImageCore 0.8.


kimikage:
Ah...`Images.complement !== ColorVectorSpace.complement`
Nevermind. It is not a practical problem.

(I was more preoccupied with [my own mistake](https://github.com/JuliaGraphics/ColorVectorSpace.jl/commit/06d70b8a28f5c263f52e414745c2066ccb72b518).:sob:)

johnnychen94:
I'll live with that because Images doesn't reexport ColorVectorSpace.

ImageCore v0.9 will reexport ColorVectorSpace so I need to work out the deprecation before we can release the new version that is compatible with ImageCore 0.9.

kimikage:
Yes, I mean...
```julia
julia> using Images

julia> complement(Gray(0.2)) # imported into `Main`
Gray{Float64}(0.8)

julia> using ColorVectorSpace
WARNING: using ColorVectorSpace.complement in module Main conflicts with an existing identifier.

julia> complement(Gray(0.2)) # still imported
Gray{Float64}(0.8)
```

```julia
julia> using Images

julia> using ColorVectorSpace

julia> complement(Gray(0.2))
WARNING: both ColorVectorSpace and Images export "complement"; uses of it in module Main must be qualified
ERROR: UndefVarError: complement not defined

PR #968, opened 2021-07-28
bring back Julia 1.0 compatibility

johnnychen94:
This partially reverts #948 by bringing back Julia 1.0 compatibility.

The only reason that Images requires 1.3 was due to ImageIO but now we can get a dummy ImageIO v0.0.1 for the old Julia version. A drawback of this change is that now users of old Julia versions (i.e., <1.3) might again meet some binary dependency installation issues, but given that the recommended solution is to "use the latest Julia release", I don't think this drawback will be a pain.

FWIW, I also documented the guarantee and non-guarantee of Images compatibility for reference purposes. This exact statement is not discussed in previous history but there're some similar thoughts. (e.g., https://github.com/JuliaImages/Images.jl/issues/824#issuecomment-683412325)

cc: @IanButterworth


.github/workflows/UnitTest.yml
```diff
@@ -7,24 +7,32 @@ on:
     branches:
       - master
   pull_request:
-  schedule:
-    - cron: '20 00 1 * *'
```
johnnychen94, line 11: The schedule job is removed because otherwise Github Actions will disable the CI automatically after 30/60 days of inactivity.

PR #971, opened 2021-08-03
🧹 code organization and dead code elimination

johnnychen94:
By setting compatibility to ImageCore 0.9 only, there are a lot of legacy codes and compat patches being unnecessary now. This PR aims to clean the codebase.

TODO:
- [x] decide which deprecations need to be `Base.depwarn(...; force=true)`

closes #974
closes #976


timholy:
It's quite satisfying to rip out large hunks of legacy code...thanks for starting this effort!

johnnychen94:
At first, I only meant to clean some useless codes and upgrade to ImageCore 0.9, but it turns out that we're cleaning almost the entire `src/algorithms.jl` here:

- [ ] `var`/`std` -> deprecated since ColorVectorSpace v0.9 (This is the only blocking issue for ImageCore v0.9 compatibility)
- [ ] `entropy` -> maybe ImageBase (I have no idea what's the direct usage of this)
- [ ] `div` -> ImageBase and rename to `fdiv` finite divergence (corresponding to our new `fdiff`)
- [ ] `imgaussiannoise` -> ImageNoise
- [ ] `imROF` -> ImageNoise
- [ ] `dilate`/`erode` -> ImageMorphology (with requires.jl)
- [ ] `imaverage`/`imlaplacian` -> deprecated with ImageFilterings.jl`
- [ ] `boxdiff`/`integral_image` -> ? (is https://github.com/JuliaImages/IntegralArrays.jl a direct replacement?)
- [ ] `gaussian_pyramid` -> maybe ImageFiltering with new name `build_pyramid` and supports generic kernels.
- [ ] `test_approx_eq_sigma_eps` let's still keep it here for the moment until we have our test utils package ready.
- [ ] `blob_LoG`/`findlocalextrema`/`findlocalmaxima`/`findlocalminima` -> ImageFiltering https://github.com/JuliaImages/ImageFiltering.jl/pull/223
- [ ] `otsu_threshold`/`yen_threshold` https://github.com/JuliaImages/Images.jl/pull/897

timholy:
The `var`/`std` fix relies on https://github.com/JuliaGraphics/ColorVectorSpace.jl/pull/175

johnnychen94:
I won't be available this week. I'll work on the `imROF` and `*_threshold` for the next two weeks. For `entropy`, if we decide to move to ImageQualityIndexes, then we can do it like [`colorfulness`](https://github.com/JuliaImages/ImageQualityIndexes.jl/blob/master/src/colorfulness.jl) so that we can allow a few more entropy implementations with shannon the default.

timholy:
No rush. I think it will be better to wait for you and do the whole thing before making the next release. But if you disagree, I can get this stuff finished and make a 0.25 release, and then we can do the rest in 0.26.

timholy:
OK, I think I've done all the ones you haven't carved out for yourself. I'm leaving `div` for you because it really seems to be part of `imROF`.

johnnychen94:
For `test_approx_eq_sigma_eps`, there are two possible places to host it: 1) https://github.com/JuliaPlots/VisualRegressionTests.jl already has a duplication so I believe people there would like to help maintain it, 2) separate https://github.com/JuliaTesting/ReferenceTests.jl into core package and image-specific packages so that we can add ImageFiltering dependency.

timholy:
I'd be fine with either. We could also put it in ImageFiltering I guess?

timholy:
OK if I rebase this? It turns out #927 had a breakage. But if you have local changes I don't want to make it harder for you.

johnnychen94:
Sure feel free to rebase

johnnychen94:
This turns out to be a big PR. Squash and merge would almost just breaks the git history. I'll do a complete rebase and hopefully make it clearer to git-blame. For items in the unchecked todo list, let's continue in new PRs.

src/deprecations.jl
```diff
@@ -1068,3 +1068,83 @@ function shepp_logan(M,N; highContrast=true)
...
+    Base.depwarn("`var(A::AbstractArray{<:Colorant})` is deprecated (and not recommended, use a MathTypes colorspace instead), please use `colorview($Cbase, var(channelview(A); dims=$newdims))[]` instead.", :var)
+    return colorview(Cbase, var(channelview(A); dims=newdims, kwargs...))[]
+end
+function std(A::AbstractArray{C}; dims=nothing, kwargs...) where C<:Colorant
+    dims === nothing || error("dims was never supported, but is now for MathTypes")
+    newdims = 2:ndims(A)+1
+    Cbase = base_colorant_type(C)
+    Base.depwarn("`std(A::AbstractArray{<:Colorant})` is deprecated (and not recommended, use a MathTypes colorspace instead), please use `colorview($Cbase, std(channelview(A); dims=$newdims))[]` instead.", :var)
+    return colorview(Cbase, std(channelview(A); dims=newdims, kwargs...))[]
+end
```
johnnychen94, line 1140: > decide which deprecations need to be Base.depwarn(...; force=true)

I feel `std` and `var` on non-mathtypes seems to be the only ones that need force deprecation.

src/Images.jl
```diff
@@ -1,113 +1,55 @@
...
+using ImageTransformations.Interpolations
 @reexport using ImageAxes
 @reexport using ImageMetadata
 @reexport using ImageFiltering
 @reexport using ImageMorphology
 @reexport using ImageDistances
 @reexport using ImageContrastAdjustment
 @reexport using ImageQualityIndexes
+@reexport using IntegralArrays
+@reexport using IntegralArrays.IntervalSets
```
johnnychen94, line 30: Just a confirmation: do we want to reexport these two modules here?

johnnychen94:
The following is a list of symbol changes by checking `names(Images)`: most of them comes from the reexport of ImageBase and IntervalSets

```diff
- imlineardiffusion
+ fdiff!
+ ImageBase
+ var
+ std
+ imresize!
+ length
+ ⊇
+ ssim
+ ±
+ isleftopen
+ duration
+ ClosedInterval
+ supremum
+ isleftclosed
+ isrightclosed
+ minimum_finite
+ OpenInterval
+ endpoints
+ leftendpoint
+ isclosedset
+ closedendpoints
+ varfinite
+ IntegralArray
+ isclosed
+ AbstractInterval
+ Interval
+ IntegralArrays
+ convert
+ ordered
+ data
+ rightendpoint
+ ..
+ isopenset
+ isrightopen
+ psnr
+ fdiff
+ maximum_finite
+ infimum
+ IntervalSets
+ sumfinite
```

I plan to remove the deprecations in ImageQualityIndexes https://github.com/JuliaImages/ImageQualityIndexes.jl/blob/master/src/deprecations.jl so that we remove `psnr` and `ssim` from the list.

johnnychen94:
The commits are rebased in a way that `Pkg.update(); Pkg.test()` passes for all of them (locally in macOS). Once https://github.com/JuliaImages/ImageMorphology.jl/pull/53 is merged we can drop d0273ff and merge this PR.

src/Images.jl
timholy: I *think* so. Though `IntervalSets` maybe not? It does seem to contribute a lot of extra names.

johnnychen94: In b4be381 I changed to

```julia
@reexport using IntegralArrays.IntervalSets.EllipsisNotation
```

johnnychen94:
This is a long journey, finally, we're so close. Every commit is carefully checked to be clean so that if we get into trouble we can quickly identify the issues.

There are still some unchecked items in the todo-list but that doesn't block the new release.

@timholy After this PR is merged we then need to update NEWS.md and the module docstring #982 .

timholy:
Thanks for all your efforts here. It is a ton of work, but the outcome feels like a major leap forward. 1.0 is starting to seem within sight...

johnnychen94:
If there's no plan to review this PR then I think we can merge it and prepare for the v0.25 release.

PR #980, opened 2021-09-14
Added stdfinite, a standard deviation equivilant to meanfinite

jagoosw:
I can't seem to find anything that has similar functionality to meanfinite but for standard deviations and so have written up what I've been using here. It calculated the standard deviation of values in an array along given dimensions and ignores any NaN values.

I have included some basic tests and it is passing the others.

Thanks!

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -183,6 +183,7 @@ export
     magnitude,
     magnitude_phase,
     meanfinite,
+    stdfinite,
     entropy,
     orientation,
     padarray,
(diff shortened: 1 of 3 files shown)


timholy:
Make sense!

These functions are moving to ImageBase, see https://github.com/JuliaImages/ImageBase.jl/pull/13. Could I ask you to reopen this there? Also, can we consider:
- making this `varfinite` (users can take the sqrt themselves)
- imitating the implementation of the other functions? I think the new implementations are both more flexible and more efficient (although checking that would be useful...)

jagoosw:
Absolutely, sorry I didn't check that issue when I was looking for where to put it.

I'll change over to that imitate the new implementations then reopen over there!

PR #985, opened 2021-11-04
rewrite the module docstring

### LINKED ISSUE #982: rewrite the `Images` docstring
The current docstring is useless so we need a rewrite of it.

https://github.com/JuliaImages/Images.jl/blob/9796acc8bf1d56d9574c555656fdba362f84a988/src/Images.jl#L206-L236

My idea is to remove the verbose list and only give a general introduction. For instance, modify from https://juliaimages.org/stable/pkgs/.

johnnychen94:
I deliberately separate this into a new PR so we don't get lost in the lengthy discussions in #971 

closes #982

### DIFF
--- a/src/Images.jl
+++ b/src/Images.jl
@@ -105,14 +105,14 @@ Images.jl is an "umbrella package" that exports a set of packages which are usef
 common image processing tasks. Most of these packages are hosted at JuliaImages,
 JuliaArrays, JuliaIO, JuliaGraphics, and JuliaMath.
 
-The purpose of this package is to have an out-of-box experiences for most of the stable
-functionalities. This means when you do `using Images`, you load a lot of packages that
-would otherwise be imported by lengthy using command, e.g., `using ImageCore, ImageShow,
-ImageTransformations, FileIO`. If you care about package loading time, you should probably
-use those small packages and try to composite your own toolbox.
+Images provides an out-of-box toolkit for image processing. This means when you do `using
+Images`, you load a lot of packages that might require multiple smaller packages, e.g.,
+`using ImageCore, ImageShow, ImageTransformations, FileIO`. However, if you want to reduce
+package, you should probably use those small packages for a narrower toolbox customized to
+your specific needs.
 
-The documentation for the JuliaImages ecosystem can be found in https://juliaimages.org, yet
-there still exist some package-specific documentation. For instance, the documentation for
+The documentation for the JuliaImages ecosystem can be found at https://juliaimages.org.
+Some dependencies have their own documentation. For instance, the documentation for
 Colors.jl is hosted in https://juliagraphics.github.io/Colors.jl even though it is included
 and exported by Images.jl.
 """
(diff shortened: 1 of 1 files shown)


timholy:
If we're removing the API summary, perhaps we should make sure that all listed `@reexport`ed modules have their own module docstrings. Otherwise there is some loss to this. But I think we could merge this and then fix those.

johnnychen94:
> If we're removing the API summary, perhaps we should make sure that all listed @reexported modules have their own module docstrings. Otherwise there is some loss to this. But I think we could merge this and then fix those.

This seems to be a 1.0 milestone. I plan to have an issue to track all the exported symbols and add necessary docstrings and demo page before 1.0 release.

src/Images.jl
```diff
@@ -101,33 +101,20 @@ export
...
-    - `shepp_logan`
+Images.jl is an "umbrella package" that exports a set of packages which are useful for
+common image processing tasks. Most of these packages are hosted at JuliaImages,
+JuliaArrays, JuliaIO, JuliaGraphics, and JuliaMath.
+
+The purpose of this package is to have an out-of-box experiences for most of the stable
+functionalities. This means when you do `using Images`, you load a lot of packages that
+would otherwise be imported by lengthy using command, e.g., `using ImageCore, ImageShow,
+ImageTransformations, FileIO`. If you care about package loading time, you should probably
+use those small packages and try to composite your own toolbox.
```
timholy, line 112:
```suggestion
use those small packages for a narrower toolbox customized to your specific needs.
```

src/Images.jl
```diff
@@ -101,33 +101,20 @@ export
...
+
+The documentation for the JuliaImages ecosystem can be found in https://juliaimages.org, yet
```
timholy, line 114:
```suggestion
The documentation for the JuliaImages ecosystem can be found at https://juliaimages.org. Some
```

src/Images.jl
```diff
@@ -101,33 +101,20 @@ export
...
+there still exist some package-specific documentation. For instance, the documentation for
```
timholy, line 115:
```suggestion
dependencies have their own documentation. For instance, the documentation for
```

src/Images.jl
```diff
@@ -101,33 +101,20 @@ export
...
-
-Test images and phantoms (see also TestImages.jl):
-
-    - `shepp_logan`
+Images.jl is an "umbrella package" that exports a set of packages which are useful for
+common image processing tasks. Most of these packages are hosted at JuliaImages,
+JuliaArrays, JuliaIO, JuliaGraphics, and JuliaMath.
+
+The purpose of this package is to have an out-of-box experiences for most of the stable
+functionalities. This means when you do `using Images`, you load a lot of packages that
```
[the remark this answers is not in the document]
johnnychen94: > for most of the stable functionalities.

I do have some concerns with this. I'll write up a new issue to explain it.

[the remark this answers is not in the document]
johnnychen94: See https://github.com/simonster/Reexport.jl/pull/37. This is why I think we can't expect Images to provide up-to-date functionalities if we want to follow the semver strictly. We have abused it a lot but I think we need to be more cautious on this after v1.0

PR #994, opened 2022-01-04
Images v0.25.1

IanButterworth:
Given https://github.com/JuliaImages/Images.jl/pull/993 has been merged it would be good to get a new patch release to free up the ecosystem for the new ImageIO version.

As an aside, I think we've tagged a couple of non-breaking releases as breaking releases for ImageIO.
I think the rationale we've had has been that it makes sense when adding new backends, but I think it's not strictly a breaking change and causes downstream noise (like this)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Images"
 uuid = "916415d5-f1e6-5110-898d-aaa5f9f070e0"
-version = "0.25.0"
+version = "0.25.1"
 
 [deps]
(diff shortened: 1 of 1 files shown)


johnnychen94 (approved):

johnnychen94:
> As an aside, I think we've tagged a couple of non-breaking releases as breaking releases for ImageIO.
I think the rationale we've had has been that it makes sense when adding new backends, but I think it's not strictly a breaking change and causes downstream noise (like this)

Because the new version drops the old LTS Julia, I feel it's a good reason to bump a minor release.

PR #1030, opened 2023-07-14
Adds ImageBinarization to Images.jl

ashwani-rathee:
- [x] Adds ImageBinarization to Images.jl


.github/workflows/UnitTest.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        julia-version: ['1.3', '1', 'nightly']
+        julia-version: ['1.6', '1', 'nightly']
```
johnnychen94, line 17: If we drop the CI for Julia < 1.6, then it's better to change `julia = "1.3"` in the `[compat]` section in Project.toml as well

johnnychen94 (approved):

johnnychen94:
I'm okay to have ImageBinarization in Images, but since I'm not working actively here, it's better for others to merge.

.github/workflows/UnitTest.yml
ashwani-rathee: Okay 👍🏻 ,Is it alright that I am changing version to 0.26.0, assuming it's kind of breaking?

johnnychen94: If so, it might be a good idea to merge #1011 and perhaps #1027 for 0.26

ashwani-rathee: So I merge this first and then see what could be done about #1011 and #1027? After all that is done, then release?

ashwani-rathee: Also I again don't have merge rights here and for good reasons 🤣

PR #1031, opened 2023-07-20
Bump compat on ImageMorphology and TiledIteration

timholy:
Closes #1011
Closes #1027

Since this will be a breaking release, it should be OK to remove
or update tests. But this part deserves scrutiny.


timholy:
The removed tests are all due to ImageMorphology, which I'm less familiar with. I locked the [compat] at 0.4 because it doesn't seem sensible to allow Images to give variable behavior depending on which version of ImageMorphology satisfies [compat] bounds.

PR #1035, opened 2023-07-22
[compat]: ImageCore/Transformations/Binarization

timholy:
Closes #1032
Closes #1033
Closes #1034


timholy:
We seem to be getting close to last call for the next release, any reasons to hold off? A motivation for doing it soon is simply to have consistent versions across the ecosystem (less precompiling...) for the JuliaCon workshop, but we can hold off if there are good reasons.

There's less breaking/deprecated stuff by far than for 0.25; in a sense the only thing breaking is ImageCore (https://github.com/JuliaImages/ImageCore.jl/blob/master/NEWS.md). Since Images.jl is sort of a "front" for ImageCore, that might be enough reason for a breaking release. But I recently learned that just dropping old Julia versions does *not* require a minor version bump for 0.x releases, so in principle this could be 0.25.4. Thoughts?

Issue #1026, opened 2023-05-03
Deprecate corner.jl in Images.jl


ashwani-rathee:
- To eventually use ImageCorners.jl as part of our restructuring effort

timholy:
Since ImageCorners is not currently a dependency of Images, this will be an easy one. I assume it's even non-breaking, right? If so we can switch now and make a release.

ashwani-rathee:
related pr : 
https://github.com/JuliaImages/Images.jl/pull/1028

timholy:
Closed by #1028

Issue #1013, opened 2022-07-18
Images.jl incompatible with ImageMorphology@0.4.1 


DominikZians:
Can't install Images along ImageMorphology version 0.4.1

Minimum example:
```
(@v1.7) pkg> generate test

(@v1.7) pkg> activate test

(test) pkg> add ImageMorphology@0.4.1
  
(test) pkg> pin ImageMorphology

(test) pkg> add Images
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package ImageMorphology [787d08f9]:
 ImageMorphology [787d08f9] log:
 ├─possible versions are: 0.1.0-0.4.1 or uninstalled
 ├─restricted to versions * by test [ff4e7b30], leaving only versions 0.1.0-0.4.1
 │ └─test [ff4e7b30] log:
 │   ├─possible versions are: 0.1.0 or uninstalled
 │   └─test [ff4e7b30] is fixed to version 0.1.0
 ├─restricted to versions 0.4.1 by an explicit requirement, leaving only versions 0.4.1
 └─restricted by compatibility requirements with Images [916415d5] to versions: 0.1.1-0.3.2 — no versions left
   └─Images [916415d5] log:
     ├─possible versions are: 0.17.3-0.25.2 or uninstalled
     └─restricted to versions * by an explicit requirement, leaving only versions 0.17.3-0.25.2
```

DominikZians:
N.B. ImageSegmantations.jl is also incompatible with ImageMorphology@0.4.1

johnnychen94:
Yeap, this is deliberately made so because otherwise, we'll need to bump Images v0.26. https://github.com/JuliaImages/Images.jl/pull/1011#issuecomment-1158671434

The current ImageMorphology@0.4 version is more or less a "preview" version for people who care about morphology processing. For common Images users that don't care much about morphology, it's better to wait ImageMorphology@0.5 (a version that will match MATLAB's functionality); by then I'll bump Images v0.26 for it.

timholy:
Closed by #1031

Issue #841, opened 2019-11-28
introduce ImageBinarization.jl and ImageContrastAdjustment.jl into Images.jl


johnnychen94:
A large part of legacy codes in Images.jl (`exposure.jl` and `algorithm.jl`)are for histogram and binarization. Since `ImageBinarization.jl` becomes stable now, we could `@reexport ImageBinarization` and do some pruning work.

It's quite difficult to determine which part of codes are legacy and can be safely deleted, this issue is used to track the list and make discussions.

Some functions I've identified (and there're more):

* `otsu_threshold` in favor of `ImageBinarization.Otsu`
* `yen_threshold` in favor of `ImageBinarization.Yen`
* `imhist`, `build_histogram` in favor of `ImageContrastAdjustment.build_histogram`
* `adjust_histogram`, `adjust_histogram!` in favor of `ImageContrastAdjustment.adjust_histogram` 

@zygmuntszpak it would be great if you can give some help 😄

zygmuntszpak:
I would like to implement a few more algorithms in both packages and then perhaps submit these packages to `https://joss.theoj.org/` with at least, @johnnychen94 and @timholy as coauthors. Would that pose a problem?

zygmuntszpak:
If we re-export the contrast adjustment functionality from `ImageContrastAdjustment`, do we replicate the tests that are in the `ImageContrastAdjustment` package in the `Images` package? Or do we want some simpler tests that just verify that the API is callable?

johnnychen94:
> do we replicate the tests that are in the ImageContrastAdjustment package in the Images package?

Regardless of CI pressure, the only problem is those tests (if hosted in Images.jl) are very likely to be outdated soon. Not only tests, but documentation also has the same problem. And the current status of Images.jl as far as I've seen is to do whatever choice here (or even without tests here); many test codes in Images.jl is becoming legacy.

> Or do we want some simpler tests that just verify that the API is callable?

This can be a good choice and I think this is what we can promise so far without too many efforts. (P.S. The current API we use in ImageBinarization doesn't take executing context (e.g., multi-threads, GPU/CPU) into consideration yet.)

As complementary to this option, I'm trying to set up a "syncing" functionality using the power of git submodules in #843 to tackle this issue for documentation, and if that works well I'm planning to do the same thing for tests.

timholy:
Closed by #1030

PR #1037, opened 2023-07-23
Add NEWS for 0.26

timholy:


NEWS.md
```diff
@@ -2,6 +2,14 @@
 
 For the full changes, please check the git history and the [release page](https://github.com/JuliaImages/Images.jl/releases)
 
+# v0.26
```
ashwani-rathee, line 5: This is looking good, let's release Images.jl 0.26: hopefully close #1026 , #1013, #841 , #682. But that, we need to deprecate these: https://github.com/JuliaImages/Images.jl/blob/4854014dd408969a640f03e20fd8d74f350607fb/src/algorithms.jl#L137. These two weren't on my mind just noticed. Also I am not sure if we need these still: `test_approx_eq_sigma_eps`

ashwani-rathee: Equivalent functions are in HistogramThresholding.jl though and is rexported by ImageBinarization.jl which we are adding in this update.

timholy: Can you make this fix? I think you know that package better than I do.

ashwani-rathee: Sure thing, let me give it a shot!

timholy: NVM, I'll give it a whirl unless you're already working on it; I'm sure you're busy with the presentation...

timholy: Hah, one thing we hadn't done was remove the old deprecations!

ashwani-rathee: Yeah, I am thinking of wrapping mostly today only

timholy: I'm already working on the fix now. It turns out to be more compllicated: we have to delete the entire contents of `src/deprecations.jl` (they're all older than 2 years, and depwarns should last only until the next breaking release) and put the new deprecations there.

timholy: xref https://github.com/JuliaImages/ImageContrastAdjustment.jl/pull/59

PR #1053, opened 2024-02-27
Update Project.toml - drop very old versions in deps

ViralBShah:
Set ver to 0.26.1


ViralBShah:
@johnnychen94 I'm merging this for now, but should be easy enough to fine tune in a new PR. I'll wait for a few days to register the release.

ViralBShah:
@johnnychen94 @timholy Would it be ok to register 0.26.1 that essentially removes some of these very old deps?

johnnychen94:
Yeah! I just triggered the registration!

PR #1063, opened 2024-11-09
Bump Project.toml version and fix a test

ViralBShah:
@johnnychen94 @timholy There's a couple of tests failing here, which would be nice to resolve and then merge.

PR #1064, opened 2024-11-09
Use ImageMagick 1.4

ViralBShah:


ViralBShah:
Same test errors as #1063

johnnychen94 (approved):

ViralBShah:
@johnnychen94 A test error here need to be fixed, which is probably introduced in one of the dependent package bumps.

```
Histogram Equalisation: Error During Test at /Users/runner/work/Images.jl/Images.jl/test/exposure.jl:78
  Got exception outside of a @test
  ArgumentError: N0f8 is an 8-bit type representing 256 values from 0.0 to 1.0; cannot represent 2
  Stacktrace:
    [1] throw_converterror(::Type{N0f8}, x::Int64)
      @ FixedPointNumbers ~/.julia/packages/FixedPointNumbers/Dn4hv/src/FixedPointNumbers.jl:326
```

ViralBShah:
The ImageMagick bump is pulling in more things. Hence can't merge it by itself and need to fix the underlying issue.

